PR #21, opened 2020-04-23
Coordinate system manager

vhirtham:


weldx/utility.py
```diff
@@ -98,3 +106,351 @@ def vector_is_close(vec_a, vec_b, abs_tol=1e-9):
...
+    """
+    # TODO: Reevaluate this branch. Does not return an "empty" object as stated in the
+    #  docstring
+    # if isinstance(input, (xr.DataArray, xr.Dataset)):
+    #    return input
+    if isinstance(input, (pd.DatetimeIndex, pd.TimedeltaIndex)):
+        return xr.DataArray(data=None, dims=["time"], coords={"time": input})
+    elif isinstance(input, dict):
+        return xr.DataArray(data=None, dims=list(input.keys()), coords=input)
+    return None  # TODO: Wouldn't an exception be more appropriate?
```
CagtayFabry, line 152: check if still needed with new xr_interp_like syntax

weldx/transformations.py
```diff
@@ -379,45 +518,449 @@ def _calculate_orthogonal_axis(a_0, a_1):
...
+
+    def interp_time(self, time: pd.DatetimeIndex) -> "LocalCoordinateSystem":
+        """
+        Interpolates the data in time.
+
+        :param time: Series of times.
+        :return: Coordinate system with interpolated data
+        """
+        if not isinstance(time, pd.DatetimeIndex):
+            raise Exception("Invalid parameter type.")
```
CagtayFabry, line 593: Instead of raising an exception, maybe we should allow for more flexibility by trying to cast to `pd.DatetimeIndex` here.
I think `List[pd.Timestamp]` (allready works) or a single `pd.Timestamp` (works if converted to List) should also be allowed and listed in type hints.

weldx/transformations.py
```diff
@@ -379,45 +518,449 @@ def _calculate_orthogonal_axis(a_0, a_1):
...
+        :param refernce: Coordinate system that provides the reference time.
+        :return: Coordinate system with interpolated data
+        """
+        if not isinstance(refernce, LocalCoordinateSystem):
+            raise Exception("Invalid reference type")
+        if refernce.time is not None:
+            times = refernce.time
+            return self.interp_time(times)
+        else:
+            raise Exception("Reference coordinate system has no time component")
```
CagtayFabry, line 614: Do we need this for error handling? Maybe we can just use `lcs.interp_time(reference.time)` instead ?

weldx/transformations.py
```diff
@@ -379,45 +518,449 @@ def _calculate_orthogonal_axis(a_0, a_1):
...
+        Raise an exception if the specified coordinate system does not exist.
+
+        :param coordinate_system_name: Name of the coordinate system, that should be
+        checked.
+        :return: ---
+        """
+        if not self.has_coordinate_system(coordinate_system_name):
+            raise Exception(
+                "There is no coordinate system with name " + str(coordinate_system_name)
+            )
```
CagtayFabry, line 681: I think raising [`ValueError`](https://docs.python.org/3.7/library/exceptions.html#ValueError) (or `KeyError`?) is more fitting in these cases

weldx/transformations.py
```diff
@@ -196,23 +232,71 @@ def __init__(
...
-        self._location = ut.to_float_array(origin)
+        if construction_checks:
+            if basis is None:
+                basis = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
+
+            if origin is None:
+                origin = np.array([0, 0, 0])
+
+            if time is not None and not isinstance(time, pd.DatetimeIndex):
+                raise Exception("time must be an instance of pandas.DateTimeIndex")
```
CagtayFabry, line 243: try casting as `pd.DatetimeIndex` instead for more flexibility

weldx/transformations.py
```diff
@@ -196,23 +232,71 @@ def __init__(
...
+                input_core_dims=[["c", "v"]],
+                output_core_dims=[["c", "v"]],
+            )
+
+            # unify time axis
+            if ("time" in basis.coords) and ("time" in origin.coords):
+                if not np.all(basis.time.data == origin.time.data):
+                    time_basis = pd.DatetimeIndex(basis.time.data)
+                    time_origin = pd.DatetimeIndex(origin.time.data)
+                    time_union = time_basis.union(time_origin)
```
CagtayFabry, line 280: we can use `utility.get_time_union` here

weldx/transformations.py
```diff
@@ -379,45 +518,449 @@ def _calculate_orthogonal_axis(a_0, a_1):
...
+        :param local_coordinate_system: An instance of
+        weldx.transformations.LocalCoordinateSystem that describes how the new
+        coordinate system is oriented in its parent system.
+        :return: ---
+        """
+        if not isinstance(local_coordinate_system, LocalCoordinateSystem):
+            raise Exception(
+                "'local_coordinate_system' must be an instance of "
+                + "weldx.transformations.LocalCoordinateSystem"
+            )
```
CagtayFabry, line 721: This (and similar cases) should raise [`TypeError`](https://docs.python.org/3.7/library/exceptions.html#TypeError).

weldx/transformations.py
```diff
@@ -379,45 +518,449 @@ def _calculate_orthogonal_axis(a_0, a_1):
...
+            source_coordinate_system_name, target_coordinate_system_name
+        )
+        if isinstance(data, xr.DataArray):
+            mul = ut.xr_matmul(
+                lcs.orientation,
+                data,
+                dims_a=["c", "v"],
+                dims_b=["c", "n"],
+                dims_out=["c", "n"],
+            )
```
CagtayFabry, line 849: Since this is matrix * vector it should look like this (without specifying __n__) for the data dimensions
```python
mul = ut.xr_matmul(
    lcs.orientation,
    data,
    dims_a=["c", "v"],
    dims_b=["c"],
    dims_out=["c"],
)
```
However, output will be of shape (n,3) instead of (3,n). I think we should agree on (n,3)-like shapes by default for broadcasting.

Issue #29, opened 2020-05-04
Tutorial groove Types


mark141:
* [x] Tutorial how to work with groove types
* [x] Plots of all groove types for a better overview
* [x] quick asdf example of groove types with notebook fileprinter

CagtayFabry:
closed by #33

PR #33, opened 2020-05-11
Tutorial grooves

mark141:
#29 Tutorial Groove Types


mark141:
Content:
* Tutorial on how to use the grooves
* profiles have an attrribute dictionary with a `unit` entry which is `None` by default
* plots will automatically name the x axis `y direction in *unit` and the y axis `z direction in *unit` because the welding direction is the x axis and the heigth is usually the z axis.
* the `*unit` will be defined by either a special option in the plot function or if that is not given by the unit in the profile attributes dictionary

CagtayFabry:
> * profiles have an attribute dictionary with a `unit` entry which is `None` by default
> * plots will automatically name the x axis `y direction in *unit` and the y axis `z direction in *unit` because the welding direction is the x axis and the height is usually the z axis.
> * the `*unit` will be defined by either a special option in the plot function or if that is not given by the unit in the profile attributes dictionary

thanks, I changed "unit" to "units" for consistency

Issue #42, opened 2020-06-10
asdf schema shape validation syntax


CagtayFabry:
We should define (and implement) a syntax that allows validation of "shape" properties (such as found in ndarray) against some templates. (updated to use `~` instead of `:`)

```
# syntax
"3"		single dimension of length 3
"~"		single dimension of arbitrary length
"..."	arbitrary number of dimensions of arbitrary length
"(3)"	optional dimension of length 3
"(~)"	optional single dimension of arbitrary length

# true
[3] -> [3]
[2,4,5] -> [2,4,5]
[1,2,3] -> [...]
[1,2,3] -> [1,2,...]
[1,2] -> [1,2,...]
[1,2,3] -> [1,2,~]
[1] -> [1,(~)]
[1] -> [1,...]
[1,2] -> [1,2,(3)]
[1,2,3] -> [1,2,(3)]

# false
[1] -> [1,2]
[1,2] -> [1]
[1,2] -> [3,2]
[1] -> [1,~]
[1,2,3] -> [1,2,(4)]
```

CagtayFabry:
might be relevant: https://medium.com/@nearlydaniel/assertion-of-arbitrary-array-shapes-in-python-3c96f6b7ccb4

vhirtham:
Additionally, the `:` operator can also be used to provide a dimension range -> `1:4`

mark141:
Added a custom validator which uses strings for `shape` and `expected_shape`. I believe all criteria are met. It would be great if someone else would check it out.

The branch:
https://github.com/BAMWelDX/weldx/tree/42_shape_validation

Question: is the `shape` a String or is it something else?

CagtayFabry:
> Added a custom validator which uses strings for `shape` and `expected_shape`. I believe all criteria are met. It would be great if someone else would check it out.
> 
> The branch:
> https://github.com/BAMWelDX/weldx/tree/42_shape_validation
> 
> Questions:
> 
> * is the `shape` a String or is it something else?
> * are there "[" in the strings - since we are operating with lists?

* Input of `_custom_shape_validator(shape, expected_shape):` will be List of Ints for `shape` and a List of Int/Strings for `expected_shape`
* the original yaml looks like this for shape `[4,5,2,1]` and the same for `expected_shape` (example: `[4,5,:,1,...]` so we cannot use addition `[ ]` as this would probably break the yaml parser

Some important notes regarding allowed syntax that passes the yaml parser:
[...]

PR #55, opened 2020-06-30
42 shape validation

CagtayFabry:
Implements the extended shape validation syntax as described in https://github.com/BAMWelDX/weldx/issues/42
example schema implementation in debug/test_shape_validator-1.0.0.yaml

A few notes:
- we should rework the behaviour to correctly yield `ValidationError` as described here https://asdf.readthedocs.io/en/2.6.0/asdf/extensions.html#adding-custom-validators
- cleanup and refactor shape validation functions
- add more tests
- how to handle the case of `( )` before `...` ?
I think this should be allowed: `[1, (n), ...]`
- remove `:` use in code ?


weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+        if ranges[0] == "" and ranges[1] == "":
+            return True
+        else:
+            if ranges[0] != "" and ranges[1] != "":
+                if int(ranges[0]) <= int(ranges[1]):
+                    pass
+                else:
+                    raise ValueError(
+                        f"The range should not be descending in {exp_string}"
+                    )
```
vhirtham, line 129: using a `>` instead of `<=` should get rid of the unused branch (`pass`). Then you can also combine it with the outer branch. So this should work:

~~~ python
if ranges[0] != "" and ranges[1] != "" and int(ranges[0]) > int(ranges[1]):
    raise ValueError(f"The range should not be descending in {exp_string}")
~~~

weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+        elif validator == 2:
+            raise ValueError('After "..." should not be another dimension.')
+        elif "..." in str(exp):
+            if "..." != exp:
+                raise ValueError(
+                    f'"..." should not have additional properties:' f" {exp} was found."
+                )
+            validator = 2
+        elif "(" in str(exp):
+            validator = 1
```
vhirtham, line 202: consider merging the first 2 `if` statements. Then you can replace the `elif`s with `if`s because a `raise` will always prevent further code execution. Then also merge these 2 statements 

~~~ python
elif "..." in str(exp):
            if "..." != exp:
~~~

into a single `if` statement.

weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+        # Compare List with expected List
+        _dict_values = _compare_lists(list_test, list_expected)
+        if _dict_values is False:
+            return False
+        for key in _dict_values:
+            if key in dict_values:
+                if dict_values[key] != _dict_values[key]:
+                    return False
+            else:
+                dict_values[key] = _dict_values[key]
```
vhirtham, line 330: Same as above

weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+        return dict_values
+    elif isinstance(dict_expected, dict):
```
vhirtham, line 332: Unnecessary `elif` ---> `if` ;)

weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+            # go one level deeper in the dictionary
+            _dict_values = _custom_shape_validator(dict_test[item], dict_expected[item])
+            if _dict_values is False:
+                return False
+            for key in _dict_values:
+                if key in dict_values:
+                    if dict_values[key] != _dict_values[key]:
+                        return False
+                else:
+                    dict_values[key] = _dict_values[key]
```
vhirtham, line 343: as above

weldx/asdf/validators.py
```diff
@@ -184,13 +439,22 @@ def wx_shape_validator(
...
-        allow_missing_keys=False,
-    )
+    dim_dict = _custom_shape_validator(instance, wx_shape)
+
+    if isinstance(dim_dict, dict):
+        return None
+    else:
+        yield ValidationError(
+            f"Error validating shape {wx_shape}.\nOn instance {instance}"
+        )
```
vhirtham, line 449: unnecessary `else`

weldx/asdf/validators.py
```diff
@@ -95,6 +95,261 @@ def _unit_validator(
...
+                # compare
+                if _list[i] != dict_values[exp]:
+                    # error found
+                    return False
+            else:
+                # add to dict_values
+                dict_values[exp] = _list[i]
+        else:
+            if i >= len(_list) or not _compare(_list[i], str(exp)):
+                return False
```
vhirtham, line 259: Again a lot of unnecessary `elif`/`else`. Also, try to combine multiple nested ifs into one statement. Every `if` introduces a new indent, making the code hard to read and hides unnecessary `elif`/`else`. So I`d suggest simplifying the whole block.

weldx/asdf/schemas/weldx.bam.de/weldx/debug/test_shape_validator-1.0.0.yaml
```diff
@@ -0,0 +1,51 @@
...
+flowStyle: block
+additionalProperties: true
+wx_shape:
+  prop1: [(~),2,n]
+  prop2: [n,2,1]
+  prop3: [2,4,5~7,...]
+  prop4: [1,3,5,k,m]
+  nested_prop:
+    p1: [10,1~10,6,4,2]
+    p2: [(m),7,5,3,1]
```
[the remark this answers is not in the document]
CagtayFabry: Yes, since every instance of `wx_shape` in the yaml tree triggers it's own validation at the specific tree location.

So if you want to validate multiple shapes with matching dimensions (e.g. `n`) like
```yaml
wx_shape:
  prop1: [(~),2,n]
  prop2: [n,2,1]
```
you have to define this in a part of the tree that has all required properties in it's scope.

For example:
When defining an "inline" `wx_shape` validator with a single "variable" like
```yaml
  prop2:
    wx_shape: [3,n,1]
```
this should be equivalent to
```yaml
  prop2:
    wx_shape: [3,~,1]
```
since there is nothing else to compare the `n` to in the first example.

tests/test_asdf.py
```diff
@@ -366,3 +368,76 @@ def test_validators():
...
+    assert val([1], [1, "..."])
+    assert val([1, 2, 3, 4, 5], [1, "..."])
+    assert val([1, 2, 3, 4, 5], ["...", 4, 5])
+    assert val([1, 2], [1, 2, "(3)"])
+    assert val([2, 3], ["(1)", 2, 3])
+    assert val([1, 2, 3], [1, "1~3", 3])
+    assert val([1, 2, 3], [1, "1~", 3])
+    assert val([1, 2, 3], [1, "~3", 3])
+    # assert val([1, 2, 3], [1, "(n)", "..."])  # should this be allowed?
+
```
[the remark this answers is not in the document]
CagtayFabry: Just add any test case you can think of (I don't think redundant tests are bad in this case either since we can use them as examples)

tests/test_asdf.py
```diff
@@ -366,3 +368,76 @@ def test_validators():
...
+    assert not val([2, 2, 3], [1, "..."])
+    assert not val([2, 2, 3], ["...", 1])
+    assert not val([1], [1, 2])
+    assert not val([1, 2], [1])
+    assert not val([1, 2], [3, 2])
+    assert not val([1], [1, "~"])
+    assert not val([1], ["~", 1])
+    assert not val([1, 2, 3], [1, 2, "(4)"])
+    assert not val([1, 2, 3], ["(2)", 2, 3])
+    assert not val([1, 2], [1, "4~8"])
```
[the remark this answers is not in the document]
CagtayFabry: please add, see above

tests/test_asdf.py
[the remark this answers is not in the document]
vhirtham: I don't even think they are redundant ;) - We don't know if we will change anything in future implementations and the mentioned test cases are different cases from the already tested ones :) --- I will add them

tests/test_asdf.py
```diff
@@ -366,3 +364,109 @@ def test_validators():
...
+        val([1, 2], [1, 2, "(-3)"])
+    with pytest.raises(ValueError):
+        val([1, 2], [1, 2, "(-3~-1)"])
+    with pytest.raises(ValueError):
+        val([1, 2], [1, 2, "(-3~1)"])
+    with pytest.raises(ValueError):
+        val([1, 2, 1], ["(-3~1)", 2, 1])
+
+
+test_shape_validator_syntax()
```
vhirtham, line 472: Remove

weldx/asdf/tags/weldx/debug/test_shape_validator.py
```diff
@@ -0,0 +1,47 @@
...
+from weldx.asdf.types import WeldxType
+from weldx.asdf.utils import drop_none_attr
+from weldx.asdf.validators import wx_shape_validator
+
+__all__ = ["ShapeValidatorTestClass", "ShapeValidatorTestClassType"]
+
+
+@dataclass
+class ShapeValidatorTestClass:
+    """<TODO CLASS DOCSTRING>"""
```
vhirtham, line 15: Just add a short description

weldx/asdf/tags/weldx/debug/test_shape_validator.py
```diff
@@ -0,0 +1,47 @@
...
+
+    prop1: np.ndarray
+    prop2: np.ndarray
+    prop3: np.ndarray
+    prop4: np.ndarray
+    nested_prop: dict
+
+
+class ShapeValidatorTestClassType(WeldxType):
+    """<TODO ASDF TYPE DOCSTRING>"""
```
vhirtham, line 25: Just add a short description

PR #56, opened 2020-07-01
Enhance CSM get_local_coordinate_system

vhirtham:
This PR fixes the problem that rotations might be affected by serious interpolation errors if a time-dependent coordinate system was transformed to another reference system.

### DIFF
--- a/weldx/transformations.py
+++ b/weldx/transformations.py
@@ -1044,15 +1044,15 @@ class CoordinateSystemManager:
 
         time_union = self.time_union(path_edges)
 
-        lcs = self.graph.edges[path[0], path[1]]["lcs"]
+        lcs = self.graph.edges[path_edges[0]]["lcs"]
+
         if time_union is not None:
             lcs = lcs.interp_time(time_union)
-
-        for edge in path_edges[1:]:
-            lcs_rhs = self.graph.edges[edge[0], edge[1]]["lcs"]
-            if time_union is not None:
-                lcs_rhs = lcs_rhs.interp_time(time_union)
-            lcs = lcs + lcs_rhs
+            for edge in path_edges[1:]:
+                lcs = lcs + self.graph.edges[edge]["lcs"].interp_time(time_union)
+        else:
+            for edge in path_edges[1:]:
+                lcs = lcs + self.graph.edges[edge]["lcs"]
 
         return lcs
 
--- a/weldx/transformations.py
+++ b/weldx/transformations.py
@@ -1021,9 +1021,7 @@ class CoordinateSystemManager:
         path = nx.shortest_path(
             self.graph, coordinate_system_name, reference_system_name
         )
-        path_edges = []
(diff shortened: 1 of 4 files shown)


weldx/transformations.py
```diff
@@ -1021,11 +1021,23 @@ def get_local_coordinate_system(
...
+            lcs = lcs.interp_time(time_union)
+
         length_path = len(path) - 1
         if length_path > 1:
             for i in np.arange(1, length_path):
-                lcs = lcs + self.graph.edges[path[i], path[i + 1]]["lcs"]
+                lcs_rhs = self.graph.edges[path[i], path[i + 1]]["lcs"]
+                if time_union is not None:
+                    lcs_rhs = lcs_rhs.interp_time(time_union)
+                lcs = lcs + lcs_rhs
```
CagtayFabry, line 1040: We can do some list operations here 😃 
```python
        path_edges = [edge for edge in zip(path[:-1], path[1:])]

        time_union = self.time_union(path_edges)

        lcs = self.graph.edges[path[0], path[1]]["lcs"]
        if time_union is not None:
            lcs = lcs.interp_time(time_union)

        for edge in path_edges[1:]:
            lcs_rhs = self.graph.edges[edge[0], edge[1]]["lcs"]
            if time_union is not None:
                lcs_rhs = lcs_rhs.interp_time(time_union)
            lcs = lcs + lcs_rhs
```

vhirtham: Added some further improvements. Now it looks really nice 😄

weldx/transformations.py
```diff
@@ -1021,11 +1040,20 @@ def get_local_coordinate_system(
...
+
+        lcs = self.graph.edges[path_edges[0]]["lcs"]
+
+        if time_union is not None:
+            lcs = lcs.interp_time(time_union)
+            for edge in path_edges[1:]:
+                lcs = lcs + self.graph.edges[edge]["lcs"].interp_time(time_union)
+        else:
+            for edge in path_edges[1:]:
+                lcs = lcs + self.graph.edges[edge]["lcs"]
```
CagtayFabry, line 1055: What do you think about `LocalCoordinateSystem.interp_time` simply returning `self` without any interpolation when `None` is passed for `time`? @vhirtham 

Than this could simply be
```python
lcs = lcs.interp_time(time_union)
for edge in path_edges[1:]:
    lcs = lcs + self.graph.edges[edge]["lcs"].interp_time(time_union)
```

I think this could simplify calls to `interp_time` in some other places as well.

vhirtham: Nice idea. We should do that.

CagtayFabry: Done @vhirtham 

Ready to merge on my part if CI passes :)

vhirtham: Like the other resulting simplifications too :) --- Merge at will!

PR #79, opened 2020-07-22
add WXRotation class and ASDF schema

CagtayFabry:
-  add `core/transoformations/rotation-1.0.0` schema that implements `scipy.spatial.transform.Rotation`
- `Rotation` objects will be stored as quaternions by default (matching the underlying base storage implementation)
- added drop-in class `WXRotation` that implements classmethods `from_quat`, `from_euler`, `from_matrix`, `from_rotvec`
With this we can simply import `WXRotation` instead of `scipy.spatial.transform.Rotation` in places where we want to keep serialization info


weldx/asdf/schemas/weldx.bam.de/weldx/core/transformations/rotation-1.0.0.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+title: |
+  Schema to describe a 3D rotation.
+description: |
+  ASDF implementation for Scipy.Rotation class.
+
+anyOf:
+  - type: object
+    description: |
+      Quaternion rotation representation.
```
vhirtham, line 15: `oneOf`?

weldx/asdf/schemas/weldx.bam.de/weldx/core/transformations/rotation-1.0.0.yaml
```diff
@@ -0,0 +1,58 @@
...
+        description: |
+          Euler angles rotation sequence.
+        type: string # TODO: add enum or regex
+      angles:
+        description: |
+          Rotation angles data.
+        tag: "tag:stsci.edu:asdf/unit/quantity-1.1.0"
+    required: [sequence, angles]
+    flowStyle: block
+    additionalProperties: false
```
vhirtham, line 57: - Should we use the unit validator here ---> validate for degrees/rad? 
- Should we validate that the length of the angles array is equal to the sequence length?

PR #82, opened 2020-07-23
raise ValidationError when validating against non existing key

### LINKED ISSUE #80: Shape validation in anyOf constructs causes problems
In the current version of the TimeSeries schema, we have 3 options which types are supported. Only for one type (discrete values) it makes sense to validate the shape. The relevant implementation looks like this:

https://github.com/BAMWelDX/weldx/blob/ba94e7ed647ed7ad0e34ee4e4ed4b8de99dd7f45/weldx/asdf/schemas/weldx.bam.de/weldx/core/time_series-1.0.0.yaml#L11-L79

The problem here is, that the validator   

https://github.com/BAMWelDX/weldx/blob/ba94e7ed647ed7ad0e34ee4e4ed4b8de99dd7f45/weldx/asdf/schemas/weldx.bam.de/weldx/core/time_series-1.0.0.yaml#L72-L79

tries to access the `"time"` key of the current object, but neither constant value nor expressions have one. Therefore an `InvalidKeyError` is raised which causes a crash. Moving the discrete implementation to the end circumvents this since the execution of `anyOf` ends as soon as a match is found. However, this is not a fix but just a "cheat" ;) - It can probably be fixed by catching the error and raising a `ValidationError` instead.

CagtayFabry:
fixes #80

### DIFF
--- a/weldx/asdf/validators.py
+++ b/weldx/asdf/validators.py
@@ -337,7 +337,7 @@ def _custom_shape_validator(dict_test, dict_expected):
         elif "shape" in dict_test:
             list_test, list_expected = _prepare_list(dict_test["shape"], dict_expected)
         else:
-            return ValidationError(f"Could not find shape key in instance {dict_test}.")
+            raise ValidationError(f"Could not find shape key in instance {dict_test}.")
 
         _validate_expected_list(list_expected)
         _dict_values = _compare_lists(list_test, list_expected)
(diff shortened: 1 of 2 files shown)


weldx/asdf/validators.py
```diff
@@ -337,7 +337,7 @@ def _custom_shape_validator(dict_test, dict_expected):
         elif "shape" in dict_test:
             list_test, list_expected = _prepare_list(dict_test["shape"], dict_expected)
         else:
-            return ValidationError(f"Could not find shape key in instance {dict_test}.")
```
vhirtham, line 340: 🤣

PR #83, opened 2020-07-24
update environment & dependency versions to asdf 2.7

CagtayFabry:


weldx/asdf/schemas/weldx.bam.de/weldx/aws/design/weld_details-1.0.0.yaml
```diff
@@ -13,12 +13,12 @@ properties:
   joint_design:
     description: |
       AWS 3.0, p. 19 : "Detailed Description of mating configuration of two components, sufficient information for shop preparation of the components." TODO CLASS
-    $ref: ../../core/iso_groove-1.0.0
+    tag: "tag:weldx.bam.de:weldx/core/iso_groove-1.0.0"
```
CagtayFabry, line 16: for some reason using `$ref` here will fail `oneOf` validation in `iso_groove-1.0.0` but no idea why (maybe cause of nested relative `#definition` refs in `iso_groove-1.0.0` ?)

So try to use `tag` validation over `$ref` as much as possible I guess

PR #94, opened 2020-07-28
Enhance tests

vhirtham:


tests/test_core.py
```diff
@@ -3,443 +3,356 @@
...
+    # Helper functions -----------------------------------------------------------------
+
+    @staticmethod
+    def _check_params_and_vars(expression, exp_params, exp_vars):
+        """Check parameters and variables of an MathematicalExpression."""
+        assert expression.num_parameters == len(exp_params)
+        assert len(expression.parameters) == len(exp_params)
+        for parameter, value in exp_params.items():
+            assert parameter in expression.parameters
+            assert expression.parameters[parameter] == value
```
CagtayFabry, line 54: I think `assert exp_params == expression.parameters` works here (dict comparison)

tests/test_core.py
```diff
@@ -3,443 +3,356 @@
...
+        expr = MathematicalExpression(expression=expression, parameters=parameters)
+
+        assert expr.num_variables == len(exp_vars)
+        for variable in exp_vars:
+            assert variable in expr.get_variable_names()
+
+        assert expr.num_parameters == len(parameters)
+        for parameter, value in parameters.items():
+            assert parameter in expr.parameters
+            assert expr.parameters[parameter] == value
```
CagtayFabry, line 81: `_check_params_and_vars(expression, exp_params, exp_vars)` ?

CagtayFabry:
How about
```python
return [
    str(var)
    for var in self._expression.free_symbols
    if not str(var) in self._parameters
]
```
here
https://github.com/BAMWelDX/weldx/blob/3cd1125182688d0375d45cc54cffda5ab93e0d6b/weldx/core.py#L205-L209

tests/test_core.py
```diff
@@ -3,443 +3,356 @@
...
 from weldx.core import MathematicalExpression, TimeSeries
 
-# MathematicalExpression ---------------------------------------------------------------
 
+# Todo: Move this to conftest.py?
+def get_test_name(param):
+    """Get the test name from the parameter list of a parametrized test."""
+    if isinstance(param, str) and param[0] == "#":
+        return param[1:]
+    return ""
```
CagtayFabry, line 20: Sounds good 👍

vhirtham: Those fancy generators 😄 - Added it

weldx/transformations.py
```diff
@@ -985,8 +985,6 @@ def __eq__(self: "CoordinateSystemManager", other: "CoordinateSystemManager"):
 
         if len(graph_0.nodes) != len(graph_1.nodes):
             return False
-        if len(graph_0.edges) != len(graph_1.edges):
-            return False
```
vhirtham, line 989: Removed it, since the number of edges is currently bound to the number of nodes (factor 2) - no chance to get this branch triggered

PR #126, opened 2020-09-08
61 lcs timedelta support

vhirtham:


tests/test_transformations.py
```diff
@@ -9,9 +9,13 @@
 import pandas as pd
 import pytest
 import xarray as xr
+from pandas import TimedeltaIndex as TDI
+from pandas import Timestamp as TS
+from pandas import date_range
 
 import weldx.transformations as tf
 import weldx.utility as ut
+from weldx.constants import WELDX_QUANTITY as Q_
```
CagtayFabry, line 18:
```suggestion
from weldx import Q_
```
let's make this the default :)

weldx/transformations.py
```diff
@@ -1004,6 +1061,13 @@ def as_euler(
         """
         return self.as_rotation().as_euler(seq=seq, degrees=degrees)
 
+    def reset_reference_time(self, time_ref_new):
+        if isinstance(time_ref_new, str):
+            time_ref_new = pd.Timestamp(time_ref_new)
```
CagtayFabry, line 1066:
```suggestion
        if not isinstance(time_ref_new, pd.Timestamp):
            time_ref_new = pd.Timestamp(time_ref_new)
```
allow everything

CagtayFabry (approved):

weldx/transformations.py
```diff
@@ -423,6 +438,7 @@ def __init__(
         orientation.name = "orientation"
 
         self._dataset = xr.merge([coordinates, orientation], join="exact")
+        self._dataset.attrs["time_ref"] = time_ref
```
CagtayFabry, line 441: We could also move this directly to the `time` coordinate for a more general way of handling time-like coordiantes in xarray (datetime-like or timedelta-like with reference time)
```python
self._dataset.time.attrs["time_ref"] = time_ref
```
What do you think @vhirtham ?

Maybe instead of setting `None` just skip and omit key?

CagtayFabry (approved):

weldx/transformations.py
vhirtham: Good idea --- done --- should be mentioned in the developer's documentation if standardize this

weldx/transformations.py
vhirtham: The issue here is, that with no extra instructions, the current exception tests for invalid inputs would fail. `None` as well as integers can be converted to `pandas.Timestamp`. Maybe we want to allow that, but we can discuss and change that another time.

PR #137, opened 2020-09-25
ReadTheDocs Documentation

vhirtham:


doc/conf.py
```diff
@@ -53,12 +56,29 @@
...
+    "numpydoc",
 ]
 
+# autosummary --------------------------------------------------------------------------
+autosummary_generate = True
+numpydoc_show_class_members = False
+
+# version = .__version__
+# The full version, including alpha/beta/rc tags.
+release = numpydoc.__version__
```
CagtayFabry, line 73: This should give the `numpydoc` instead of `weldx` version? (see line 43)

vhirtham: I have to review the whole block (5 lines) myself. I am not even sure if it is needed.  I will try removing it and see what happens :)

vhirtham: Just checked the numpydoc documentation and updated the block (removed the line you mentioned)

CagtayFabry (approved):

PR #143, opened 2020-09-30
add base classes to init file

### LINKED ISSUE #134: Collect all frequently used classes in __init__.py
Currently, we need to import several of our own and external modules if we want to write a script or notebook. Maybe we should gather all frequently used classes in the __init__.py to increase the usage compfort. I think that this would also be good for new users if they don't need to remember which module contains which class. So writing `import weldx as wx` would be sufficient in most cases.

CagtayFabry:
closes #134

### DIFF
--- a/weldx/__init__.py
+++ b/weldx/__init__.py
@@ -9,7 +9,17 @@ import weldx.geometry
 import weldx.transformations
 import weldx.utility
 import weldx.welding
+
+# class imports to weldx namespace
+from weldx.asdf.tags.weldx.core.iso_groove import get_groove
 from weldx.constants import WELDX_QUANTITY as Q_
+from weldx.core import MathematicalExpression, TimeSeries
+from weldx.geometry import ArcSegment, Geometry, LineSegment, Trace
+from weldx.transformations import (
+    CoordinateSystemManager,
+    LocalCoordinateSystem,
+    WXRotation,
+)
 
 # versioneer
 from ._version import get_versions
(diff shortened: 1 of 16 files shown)


CagtayFabry:
you can just add more classes and review once done @vhirtham

weldx/__init__.py
```diff
@@ -9,7 +9,17 @@
 import weldx.transformations
 import weldx.utility
 import weldx.welding
+
+# class imports to weldx namespace
+from weldx.asdf.tags.weldx.core.iso_groove import get_groove
 from weldx.constants import WELDX_QUANTITY as Q_
+from weldx.core import MathematicalExpression, TimeSeries
+from weldx.geometry import ArcSegment, Geometry, LineSegment, Profile, Shape, Trace
```
vhirtham, line 17: We should add the other classes of the package too (trace segments, variable profile)

vhirtham (changes requested):

weldx/__init__.py
CagtayFabry: > We should add the other classes of the package too (trace segments, variable profile)

I skipped them since they are not actively used anywhere currently I think and I would prefer not cluttering the weldx namespace too much.
But feel free to add them

vhirtham: > I skipped them since they are not actively used anywhere currently I think
At least the trace segments should appear in every file that has a geometry since you need them to create a trace. However, we can always add more later, so just merge it.

vhirtham (approved):

PR #168, opened 2020-10-09
add sine TimeSeries function

CagtayFabry:
## Changes
Adds a very simple function to easily create sine-like `TimeSeries` for tutorials etc.

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/utility.py
```diff
@@ -13,6 +13,41 @@
 
 import weldx.transformations as tf
 from weldx.constants import WELDX_QUANTITY as Q_
+from weldx.core import MathematicalExpression, TimeSeries
+
+
+def _sine(
```
vhirtham, line 19: why "private" (`_`)?

vhirtham (approved):

weldx/utility.py
CagtayFabry: I think we should "publish" functions like this later in a more concise way (like a "math" module or similar)

So for now nice to have but not really kind of the public API

PR #174, opened 2020-10-19
132 csm timedelta update cf

CagtayFabry:
## Changes
Code refactor regarding handling of time-like types and objects

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/utility.py
```diff
@@ -146,21 +158,48 @@ def to_pandas_time_index(time) -> Union[pd.TimedeltaIndex, pd.DatetimeIndex]:
...
+        return time
+
+    # try manual casting for object dtypes (i.e. strings), should avoid integers
+    # warning: this allows something like ["1","2","3"] which will be ns !!
+    if is_object_dtype(time):
+        for func in (pd.DatetimeIndex, pd.TimedeltaIndex):
+            try:
+                return func(time)
+            except Exception:
+                continue
```
CagtayFabry, line 198: only str -> DatetimeIndex

weldx/transformations.py
```diff
@@ -490,25 +477,10 @@ def __init__(
...
-                    "time delta type."
-                )
-            if "time" in coordinates.coords and not isinstance(
-                ut.to_pandas_time_index(coordinates.time.data), pd.TimedeltaIndex
-            ):
-                raise ValueError(
-                    "The 'coordinate' data set's time coordinate is not a valid time"
-                    " delta type."
-                )
-
```
CagtayFabry, line 507: now covered with `xr_check_coords`

weldx/transformations.py
```diff
@@ -2423,11 +2401,14 @@ def get_cs(
...
-                    lcs = deepcopy(lcs)
+            if lcs.is_time_dependent:
+                if not lcs.has_reference_time and self.has_reference_time:
+                    time_lcs = time_interp + (time_ref_interp - self.reference_time)
+                    lcs = lcs.interp_time(time_lcs)
                     lcs.reset_reference_time(self.reference_time)
-                lcs = lcs.interp_time(time_interp, time_ref_interp)
+                    lcs.reset_reference_time(time_ref_interp)
+                else:
+                    lcs = lcs.interp_time(time_interp, time_ref_interp)
```
CagtayFabry, line 2411: I added this as an alternative to creating a deepcopy
Basically create a single `TimedeltaIndex` for LCS interpolation that is shifted to match the CSM reference time. Need to add time_ref data afterwards and shift to final `time_ref_interp`

we can also keep the old code @vhirtham

CagtayFabry:
should be ready now @vhirtham

weldx/transformations.py
```diff
@@ -1393,12 +1352,25 @@ def uses_absolute_times(self) -> bool:
 
         Returns
         -------
-        bool :
+        bool :[
```
vhirtham, line 1355: Is the `[` wanted? If yes, what's the purpose?

weldx/transformations.py
vhirtham: As you like, but isn't `lcs.reset_reference_time(self.reference_time)` obsolete? Why do we need to reset the reference time twice directly after each other? The second call should be independent of the first one or am I missing something?

weldx/utility.py
```diff
@@ -142,25 +154,52 @@ def to_pandas_time_index(time) -> Union[pd.TimedeltaIndex, pd.DatetimeIndex]:
 
     Returns
     -------
-    Union[pandas.TimedeltaIndex, pandas.DatetimeIndex]:
-        Variable as pandas time index
+    pandas.TimedeltaIndex or pandas.DatetimeIndex
+        Time union of all input objects
```
vhirtham, line 158: We should definitely mention the capabilities of this function in the developer's documentation. Otherwise, it might get overlooked in the future since the naming doesn't give much of a hint here.

weldx/utility.py
```diff
@@ -883,3 +939,58 @@ def interp_like(self, da, *args, **kwargs):
...
+                return da.time.attrs["time_ref"]
+
+        return None
+
+    @time_ref.setter
+    def time_ref(self, value: pd.Timestamp):
+        """Convert INPLACE to new reference time.
+
+        If no reference time exists, the new value will be assigned
+        TODO: should None be allowed and pass through or raise TypeError ?
```
vhirtham, line 983: Would tend to raise a `ValueError` for now until we think we need to change that

vhirtham (approved):

weldx/transformations.py
CagtayFabry: Certainly not 🙃

weldx/utility.py
CagtayFabry: Maybe add to time-related dev doc?

weldx/utility.py
CagtayFabry: Ok, was thinking about if we might need `da1.weldx.time_ref = da2.weldx.time_ref` somewhere

PR #177, opened 2020-10-23
example implementation with removing default coordinates/orientation

### LINKED ISSUE #157: remove LCS default values from serialization
Don't add default translation/rotation when serializing to ASDF files

CagtayFabry:
## Changes
If the coordinates or orientation values of an LCS instance equate to their default values (`[0,0,0]` or `np.eye(3)`) the property is removed form the ASDF serialization to keep the resulting file more readable.

## Related Issues
Closes #157 

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


CagtayFabry:
When working at this I noticed that we create a "broken" LCS instance with for some tests in `tests/test_asdf_core.test_local_coordinate_system_shape_violation`

I don't really think is should be allowed, even with `construction_checks==False` we should keep the dimension and dtype checks in place. Maybe we should think about a different way to test the asdf schema.

What do you think @vhirtham

vhirtham (approved):

PR #199, opened 2020-11-19
Add geometry distance example

CagtayFabry:
## Changes
Add a simple example tutorial that describes (in theory) how to design and evaluate the TCP distance to the workpiece with complex weaving motions and a linear change i groove profiles.

Also adds a little helper function `lcs_coords_from_ts`

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/utility.py
```diff
@@ -105,6 +107,32 @@ def _sine(
...
+    -------
+    A xarray.DataArray with correctly labeled dimensions to be used for LCS creation.
+
+    """
+    ts_data = ts.interp_time(time=time)
+    # assign vector coordinates and convert to mm
+    ts_data = ts_data.rename({"dim_1": "c"}).assign_coords({"c": ["x", "y", "z"]})
+    ts_data.data = ts_data.data.to("mm").magnitude
+    ts_data["time"] = pd.TimedeltaIndex(ts_data["time"].data)
+    return ts_data
```
vhirtham, line 133: Maybe I just missed it but shouldn't we keep the unit (`mm`)?

vhirtham (approved):

weldx/utility.py
CagtayFabry: I don't think we support units anywhere with regards to LCS (yet)

It's no problem creating a "quantity LCS" here but combining with non unit aware LCS during CSM interpolation throws the expected `DimensionalityError`

Should try to fix this soon, currently not even a default unit for LCS coords is mentioned or agreed upon anywhere

weldx/utility.py
```diff
@@ -122,7 +120,8 @@ def lcs_coords_from_ts(
 
     Returns
     -------
-    A xarray.DataArray with correctly labeled dimensions to be used for LCS creation.
+    xarray.DataArray :
+        A DataArray with correctly labeled dimensions to be used for LCS creation.
```
CagtayFabry, line 126: @vhirtham 👍 🚀

PR #209, opened 2020-12-04
add default metadata attributes

### LINKED ISSUE #205: how to handle user metadata support for all objects
Now that quality standards are on their way 🚀 we should discuss how to support/implement additional user requirements to schemas.
Restricting or changing existing schema requirements should be straight forward by replacing the schema mappings, but what about adding new fields etc. ? The problem I see here is that by default, the weldx implemented `to_tree` methods would not know about additional fields to serialize even if these were added as attributes of a class.

I think we could do something similar to what `xarray` does with `.attrs`. Define a `WeldxBase` class with a designated attribute (suppose `metadata` oder `userdata`) that can hold a dict. This would be the place for users to store any kind of object they want to serialize with this specific instance.

Now if a user wants to define their own standard with completely different metadata fields (think standard definitions for internal workflows) they could implement this on the `metadata` property without affecting or having to change anything on the python side.

```yaml
# example for a standard requiring a metadata field for "arrangement"
properties:
  metadata:
    type: object
    properties:
      arrangement:
        description: camera arrangement layout
        type: string
        enum: [linear, circular]
    requiredProperties: [arrangement]
```

### how could we make this work?
I think it would be cumbersome and error prone to implement the serialization of the `metadata` attribute in every single `to_tree` method by hand. Instead we should define a base (meta)class that dynamically applies the necessary changes to the `to_tree` method after class creation. We could also use a class decorator similar to here for this but I think defining a more general metaclass might have other uses down the line.
Or maybe there is a simpler way?

### schema considerations
While the Python side should be quite flexible, the ASDF schemas in this regard are rather not. Two possibilities:
1. add a `metadata` property to every single schema.
1. remove `additionalProperties: true` from schemas where it is set
1. the mixer: only add the `metadata` property in schemas where we absolutely need `additionalProperties: true`

CagtayFabry:
## Changes
basically the simplest way I could think of to always serialize `metadata` attributes of any class (if it exists)

short example tutorial [here](https://weldx--209.org.readthedocs.build/en/209/tutorials/custom_metadata.html) @AndreasPittner 

## ToDos
- [x] wrap `from_tree` to read metadata field
- [x] remove `additionalProperties` from all schemas by default
- [ ] I have a feeling it would be safer to wrap `to_tree_tagged` instead of `to_tree` since this is the final function that produces the `TaggedObject`
I think we can skip this for now but may have to be careful for classes that implement their own `to/from_tree_tagged`

## Related Issues
Closes #205 

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/tutorials/custom_metadata.ipynb
+++ b/tutorials/custom_metadata.ipynb
@@ -4,14 +4,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Adding custom meta-data\n",
-    "It is possible to attach and store arbitrary meta-data to WelDX objects as python attributes.\n",
+    "# Adding custom metadata\n",
+    "It is possible to attach and store arbitrary metadata to WelDX objects as python attributes.\n",
     "The WelDX API reserves the attribute names `wx_metadata` and `wx_user` for these use cases.\n",
     "\n",
     "The `wx_metadata` field is intended for additional information that might be required as part of a custom Quality Standard. The `wx_user` attribute is intended to add any user data and should only be used for additional information that will not get accessed by the API. \n",
     "\n",
     "## A small example\n",
-    "Let's say we want to use meta-data to add a custom information to a current-sensor used during a welding experiment. As an example we want to add a timestamp to indicate when the sensor was last calibrated."
+    "Let's say we want to use metadata to add a custom information to a current-sensor used during a welding experiment. As an example we want to add a timestamp to indicate when the sensor was last calibrated."
    ]
   },
   {
@@ -48,11 +48,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now we want to add the timestamp of the last time the sensor was calibrated as additional meta-data. To do this we simple assign a new `.wx_metadata` attribute to the sensor. We will also add a small personal note to `wx_user`.\n",
+    "Now we want to add the timestamp of the last time the sensor was calibrated as additional metadata. To do this we simple assign a new `.wx_metadata` attribute to the sensor. We will also add a small personal note to `wx_user`.\n",
     "\n",
-    "<div class=\"alert alert-block alert-info\"><b>Note:\n",
-    "    By convention the meta-data fields should be python dictionaries.\n",
-    "</b>"
+    "*Note: By convention the metadata fields should be python dictionaries.*"
    ]
   },
   {
@@ -71,7 +69,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now let's look at the ASDF file contents with our meta-data:"
+    "Now let's look at the ASDF file contents with our metadata:"
    ]
   },
   {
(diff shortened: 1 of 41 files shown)


CagtayFabry:
A note regarding the TimeSeries schema:

It's a bit tricky to remove `additionalProperties` the way the schema properties are defined right now because they all use the same key `values` and `unit` right now
We could use different keys for `values` depending on the `TimeSeries` type, e.g. `const`, `expression` and `values`
```yaml
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
id: "http://weldx.bam.de/weldx/core/time_series-1.0.0"
tag: "tag:weldx.bam.de:weldx/core/time_series-1.0.0"

title: |
  Schema that describes a time series.
description: |
  Describes a time dependent quantity.
oneOf:
  - type: object
    description: |
      Implementation for constant values.
    properties:
      values:
        description: |
          Number or n-dimensional array that is constant in time.
        oneOf:
          - type: number
          - tag: "tag:stsci.edu:asdf/core/ndarray-1.0.0"
      unit:
        description: |
          Unit of the data.
        type: string
    required: [values, unit]
    additionalProperties: false

  - type: object
    description: |
      Implementation for expressions.
    properties:
      values:
        description: |
          A mathematical expression that describes the time dependent behaviour.
        tag: "tag:weldx.bam.de:weldx/core/mathematical_expression-1.0.0"
      unit:
        description: |
          Resulting unit of the data when the expression is evaluated.
        type: string
      shape:
        description: |
          (optional) Resulting shape of the data when the expression is evaluated.
        type: array
    required: [values, unit]
    additionalProperties: false

  - type: object
    description: |
      Implementation for discrete data.
    properties:
    ....
      unit:
        description: |
          Unit of the data.
        type: string
     ....
      values:
        description: |
          Set of discrete n-dimensional data.
        tag: "tag:stsci.edu:asdf/core/ndarray-1.0.0"
    required: [time, unit, shape, interpolation, values]
    additionalProperties: false

propertyOrder: [time, unit, shape, interpolation, values]
flowStyle: block
...
```

tutorials/custom_metadata.ipynb
```diff
@@ -0,0 +1,138 @@
...
+   "metadata": {},
+   "source": [
+    "# Adding custom meta-data\n",
+    "It is possible to attach and store arbitrary meta-data to WelDX objects as python attributes.\n",
+    "The WelDX API reserves the attribute names `wx_metadata` and `wx_user` for these use cases.\n",
+    "\n",
+    "The `wx_metadata` field is intended for additional information that might be required as part of a custom Quality Standard. The `wx_user` attribute is intended to add any user data and should only be used for additional information that will not get accessed by the API. \n",
+    "\n",
+    "## A small example\n",
+    "Let's say we want to use meta-data to add a custom information to a current-sensor used during a welding experiment. As an example we want to add a timestamp to indicate when the sensor was last calibrated."
```
vhirtham, line 14:
```suggestion
    "Let's say we want to use meta-data to add a custom information to a current-sensor used during a welding experiment. As an example, we want to add a timestamp to indicate when the sensor was last calibrated."
```

vhirtham (approved):
👍

PR #212, opened 2020-12-10
remove None values form asdf tree by default

CagtayFabry:
This is to reproduce the old ASDF behavior pre https://github.com/asdf-format/asdf/pull/863
Some tutorials seem to run now 👍 

Also, some utilities provided by `boltons` could be useful for further tree manipulation
https://sedimental.org/remap.html#drop-empty-values
https://boltons.readthedocs.io/en/latest/iterutils.html#nested

## Notes
- We should discuss if we want this as default behavior
- we can remove the `drop_none_attr` wrapper from `weldx.asdf.utils` with this (_done_)

## Changes
- default wrapper for `to_tree` now also removes `None` values

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


CagtayFabry:
this should work with any ASDF version so I think we should include this in master and rebase here once ready @vhirtham 

this branch -> master -> asdf28

vhirtham (approved):

Issue #91, opened 2020-07-28
store external file contents


CagtayFabry:
support including external files into weldx files
- filetype and other metadata
- provide explicit hash for fast checks?

CagtayFabry:
we should discuss which metadata to store in the ASDF tree, probably starting with https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat 

for the schema I suggest the following 
- file_uri (absolute)
- filename
- suffix (may be redundant)

vhirtham:
To keep the discussion in one place, here your suggestion from the pull request:

~~~ yaml
properties:
  file_uri:
    type: string
    pattern: "^file://"
    description: The absolute file URI.
  filename:
    type: string
    pattern: # some pattern to validate filenames
    description: original filename
  suffix:
    type: string
    description: The filetype/suffix.
    # this would be a good place to put restrictions in other schemas or quality standards
  created:
    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.0.0"
  modified:
    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.0.0"
  # some other file metadata
  content:
    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
  content_hash:
    type: object
    properties:
      algo:
        type: string
        enum: [MD5, SHA-256, SHA-512] # probably look up some codes
      value:
        type: string
~~~

I am not sure about the `file_uri`.  Maybe I am getting confused again by URI and URL but what is its purpose? If we want to define directory structures of multiple related files I'd suggest that we create an extra `file_structure` or `file_system` schema for that like we did with the `CoordinateSystemManager` and `LocalCoordinateSystem`s. 

I am also not sure about the `suffix`. It is either redundant or requires us to apply another operation on the name that doesn't provide any additional information.

I would consider `created` and `modified` as metadata. I am not even sure if we can restore the file with those timestamps applied to it. Well technically it should be possible since archiving tools can do that too, but I don't know if that might get ugly.

The `hash` is something we should store as extra validation. Just need to perform some research on our options.

PR #215, opened 2020-12-10
91 asdf external file

vhirtham:
## Changes
Add classes to document and store external files in an ASDF file

## Related Issues
Closes [#91](https://github.com/BAMWelDX/weldx/issues/91)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


CagtayFabry:
Some suggestions for extending the file schema properties:

```yaml
properties:
  file_uri:
    type: string
    pattern: "^file://"
    description: The absolute file URI.
  filename:
    type: string
    pattern: # some pattern to validate filenames
    description: original filename
  suffix:
    type: string
    description: The filetype/suffix.
    # this would be a good place to put restrictions in other schemas or quality standards
  created:
    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.0.0"
  modified:
    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.0.0"
  # some other file metadata
  content:
  content_hash:
    type: object
    properties:
      algo:
        type: string
        enum: [MD5, SHA-256, SHA-512] # probably look up some codes
      value:
        type: string
```

weldx/core.py
```diff
@@ -540,3 +543,41 @@ def units(self) -> str:
...
+        if isinstance(path, str):
+            path = Path(path)
+
+        self._filename = path.name
+
+        if buffer is None:
+            if file_system is None:
+                with OSFS(path.parent.absolute().as_posix()) as file_system:
+                    self._buffer = np.frombuffer(
+                        file_system.readbytes(self._filename), dtype=np.int8
```
CagtayFabry, line 564: not sure if this changes anything but maybe `np.uint8` is safer here

vhirtham: You're probably right. Missed that when I copied the line from Stackoverflow.

vhirtham:
Some things that (might) need to be changed:
- The hash is currently stored as a numpy array in the block since I couldn't convert it to a `utf-8` encoded string
- The hash should probably be checked using a validator

CagtayFabry:
some notes:

- store hash value as string from `hasher.hexdigest()`
- add filesize info to ASDF schema
- I think the buffer (and hash) should only be created during `to_tree`, storing the file earlier will just fill up memory
- I would prefer not using `fs` here and keep it as simple as possible (sticking to `pathlib` it should have all we need)
- by default, don't store file contents in the asdf tree, just file metadata (we can make it a `store_contents` property to toggle before `to_tree`)
- hash should be created form chunks to keep maximum memory consumption low (see https://stackoverflow.com/a/22058673)
  - if the file gets serialized it's probably best to do this over the buffer in memory?

weldx/core.py
```diff
@@ -540,3 +542,62 @@ def units(self) -> str:
...
+        ----------
+        path : :Union[str, Path]
+            Path where the file should be written.
+
+        """
+        if isinstance(path, str):
+            path = Path(path)
+
+        with OSFS(path.absolute().as_posix()) as file_system:
+            file_system.writebytes(self._filename, self._buffer)
```
CagtayFabry, line 603:
```python
path.write_bytes(buffer)
# or maybe
with path.open("wb") as f:
    f.write(self._buffer)
```

weldx/core.py
```diff
@@ -540,3 +542,62 @@ def units(self) -> str:
...
+# ExternalFile -------------------------------------------------------------------------
+
+
+class ExternalFileBuffer:
+    """Handles external files."""
+
+    def __init__(self, path, file_system=None, buffer: np.ndarray = None):
+        if isinstance(path, str):
+            path = Path(path)
+
```
CagtayFabry, line 556: we should check `path.is_file()` here

vhirtham:
Unfortunately I didn't push yesterday and some comments refer to outdated implementations :(

>- store hash value as string from `hasher.hexdigest()`

will try that

>- add filesize info to ASDF schema

but only for the unserialized files, right? 

>- I think the buffer (and hash) should only be created during `to_tree`, storing the file earlier will just fill up memory

Was already changed, yesterday

>- I would prefer not using `fs` here and keep it as simple as possible (sticking to `pathlib` it should have all we need)

not sure why you want to avoid `fs` so hard. I get that the user shouldn't have to bother with it, if he doesn't know and need it, but that is not the case here. It is just an option that provides additional flexibility at basically no cost regarding the implementation. 

>- by default, don't store file contents in the asdf tree, just file metadata (we can make it a `store_contents` property to toggle >before `to_tree`)

Already changed

>- hash should be created form chunks to keep maximum memory consumption low (see >https://stackoverflow.com/a/22058673)

will have a look into that

>  - if the file gets serialized it's probably best to do this over the buffer in memory?

Not sure what you mean here

CagtayFabry:
> 
> > * add filesize info to ASDF schema
> 
> but only for the unserialized files, right?
> 

I would suggest always storing the filesize in the same place, even if it's redundant with array length


> 
> > * I would prefer not using `fs` here and keep it as simple as possible (sticking to `pathlib` it should have all we need)
> 
> not sure why you want to avoid `fs` so hard. I get that the user shouldn't have to bother with it, if he doesn't know and need it, but that is not the case here. It is just an option that provides additional flexibility at basically no cost regarding the implementation.

I just don't see the need for `fs` currently

 
> > * if the file gets serialized it's probably best to do this over the buffer in memory?
> 
> Not sure what you mean here

If the buffer is (or will get) loaded for serialization, calculate hash from the buffer (in memory) to avoid another disk access :)

vhirtham:
Regarding this point:

> hash should be created form chunks to keep maximum memory consumption low (see https://stackoverflow.com/a/22058673)
if the file gets serialized it's probably best to do this over the buffer in memory?

Currently, we have to load the whole file into a buffer anyway to store it in the tree.  Until we find out how we can write/stream directly into the binary block, there is no point in creating the hash in chunks.

weldx/asdf/schemas/weldx.bam.de/weldx/core/file-1.0.0.yaml
```diff
@@ -0,0 +1,67 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://weldx.bam.de/weldx/core/file-1.0.0"
+tag: "tag:weldx.bam.de:weldx/core/file-1.0.0"
+
```
CagtayFabry, line 6: It would be nice add an example here that will be rendered in the documentation

see:
https://github.com/BAMWelDX/weldx/blob/8e14c0f0fa7290d99a760a2c7d7e4efe2996c900/weldx/asdf/schemas/weldx.bam.de/weldx/groove/iso_9692_1_2013_12/VGroove-1.0.0.yaml#L10-L19

vhirtham: Unfortunately, that isn't so easy. During the asdf tests, the code of the example gets actually evaluated and the `to_tree` and `from_tree` are executed. Due to the missing file, that always results in a failure. Any idea how to solve that?

vhirtham:
Still thinking of moving the code from the `core` module into the asdf directory since it has no use without asdf. What do you think?

weldx/core.py
```diff
@@ -540,3 +544,176 @@ def units(self) -> str:
...
+    asdf_save_content: bool = False
+    buffer: bytes = None
+    file_system = None
+
+    def __post_init__(self):
+        """Initialize the internal values."""
+        if self.file_system is not None:  # pragma: no cover
+            raise Exception(
+                "Using file systems to load a file is currently not supported."
+            )
```
CagtayFabry, line 578: I think we can remove the `file_system` attribute from the class?
I like the way it gets passed to `write_to` when needed 👍

weldx/asdf/schemas/weldx.bam.de/weldx/core/file-1.0.0.yaml
```diff
@@ -0,0 +1,84 @@
...
+title: |
+  Schema for a file.
+description: |
+  This schema describes a file by compiling all its meta data. Optionally, the whole file can also be stored in the
+  binary block of an asdf file.
+
+examples:
+  -
+    - A simple DHU-Groove
```
CagtayFabry, line 15: Doesn't look like DHU-Groove to me 😉

weldx/asdf/schemas/weldx.bam.de/weldx/core/file-1.0.0.yaml
```diff
@@ -0,0 +1,84 @@
...
+      The timestamp when the file was modified last.
+    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.0.0"
+  hostname:
+    description: |
+      The name of the files host machine.
+    type: string
+  directory:
+    description: |
+      The directory of the file on the specified host system.
+    type: string
```
CagtayFabry, line 56:
```suggestion
  hostname:
    description: |
      The name of the host machine accessing the file.
    type: string
  directory:
    description: |
      The directory of the file as seen from the host machine.
    type: string
```
Is this how it works?

CagtayFabry:
> Still thinking of moving the code from the `core` module into the asdf directory since it has no use without asdf. What do you think?

yes this can live inside the asdf code files
But we can import it at a higher level namespace

CagtayFabry (changes requested):
What do you think about the code changes in #230 ? @vhirtham

vhirtham:
> What do you think about the code changes in #230 ? @vhirtham

They are nice, just left a comment in the PR about what I think we should change.

weldx/asdf/schemas/weldx.bam.de/weldx/core/file-1.0.0.yaml
vhirtham: Then keep looking closer :D --- After several asdf test failures and copy-pasting it started looking like one to me ;) --- I'll fix it

weldx/core.py
vhirtham: We can do it but I thought we should keep it as a reminder since I think it is more useful for actually reading a file (archives, servers, etc) than for writing (Jupyter examples).  --- However you like

CagtayFabry: My vote goes to removing for now and pick this up again if we need it :)

tests/asdf_tests/test_asdf_core.py
```diff
@@ -1,18 +1,32 @@
...
 from weldx.core import MathematicalExpression as ME  # nopep8
 from weldx.core import TimeSeries
 from weldx.transformations import WXRotation
 
+
+def get_test_name(param):
+    """Get the test name from the parameter list of a parametrized test."""
+    if isinstance(param, str) and param[0] == "#":
+        return param[1:]
+    return ""
```
vhirtham, line 27: This is a duplication of a function in the test file of the core module. It`s needed for the exception. It produces a clearer output of the test results. Will move it to `_helpers.py`

CagtayFabry (approved):

CagtayFabry:
LGTM 🚀 

please add PR link to `CHANGELOG.MD` and fill out the PR template here before merge @vhirtham

PR #219, opened 2020-12-16
add CSM.relabel function

CagtayFabry:
## Changes
add `CSM.relabel` function to rename nodes in `CSM`

This should be safe right.. @vhirtham ? ;)
More of a utility so no test for now

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/weldx/transformations.py
+++ b/weldx/transformations.py
@@ -1795,6 +1795,20 @@ class CoordinateSystemManager:
                     lcs,
                 )
 
+    def relabel(self, mapping: Dict[str, str]):
+        """Rename one or more nodes of the graph.
+
+        See `networkx.relabel.relabel_nodes` for details.
+        CSM will always be changed inplace.
+
+        Parameters
+        ----------
+        mapping
+             A dictionary mapping with the old node names as keys and new node names
+             labels as values.
+        """
+        nx.relabel_nodes(self.graph, mapping, copy=False)
+
     def assign_data(
         self, data: xr.DataArray, data_name: str, coordinate_system_name: str
     ):
--- a/weldx/transformations.py
+++ b/weldx/transformations.py
@@ -1808,6 +1808,9 @@ class CoordinateSystemManager:
              labels as values.
 
         """
+        if len(self.subsystems) > 0:
+            raise NotImplementedError("Cannot relabel nodes on merged systems.")
+
         nx.relabel_nodes(self.graph, mapping, copy=False)
 
     def assign_data(
--- a/weldx/transformations.py
+++ b/weldx/transformations.py
@@ -1806,6 +1806,7 @@ class CoordinateSystemManager:
         mapping
              A dictionary mapping with the old node names as keys and new node names
              labels as values.
+
         """
         nx.relabel_nodes(self.graph, mapping, copy=False)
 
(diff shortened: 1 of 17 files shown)


vhirtham:
> This should be safe right.. @vhirtham ? ;)

Have to review the corresponding code sections, but I am pretty sure that it isn't safe in its current form when combined with subsystems.  If you rename the common node of the parent and child system, you need to update the subsystem data accordingly. Additionally, you might need to store the original name. That might get a little bit ugly. That's why we skipped this functionality back then.

CagtayFabry:
> > This should be safe right.. @vhirtham ? ;)
> 
> Have to review the corresponding code sections, but I am pretty sure that it isn't safe in its current form when combined with subsystems. If you rename the common node of the parent and child system, you need to update the subsystem data accordingly. Additionally, you might need to store the original name. That might get a little bit ugly. That's why we skipped this functionality back then.

Ah yes, I see this could be dangerous on already merged subsystems ... maybe we can catch this and disable in this case?
However on a "clean" CSM before merging this would be safe? I need to rename a node to make merging possible

> 
> > More of a utility so no test for now
> 
> You are living dangerous...

😈

vhirtham:
> Ah yes, I see this could be dangerous on already merged subsystems ... maybe we can catch this and disable in this case?
However on a "clean" CSM before merging this would be safe? I need to rename a node to make merging possible

Yeah, we can disable it for now as a quick fix
~~~ python
if self._subsystems:
    raise("CFabry did this :p")
~~~

weldx/transformations.py
```diff
@@ -1795,6 +1795,24 @@ def add_cs(
...
+
+        Parameters
+        ----------
+        mapping
+             A dictionary mapping with the old node names as keys and new node names
+             labels as values.
+
+        """
+        if len(self.subsystems) > 0:
+            raise NotImplementedError("Cannot relabel nodes on merged systems.")
```
CagtayFabry, line 1812: tested locally and seems to raise correctly

we can also make this a ""private" function for now until we are sure this doesn't break anything? @vhirtham

weldx/transformations.py
```diff
@@ -1795,6 +1795,24 @@ def add_cs(
...
+        CSM will always be changed inplace.
+
+        Parameters
+        ----------
+        mapping
+             A dictionary mapping with the old node names as keys and new node names
+             labels as values.
+
+        """
+        if len(self.subsystems) > 0:
```
vhirtham, line 1811:
```suggestion
        if self.subsystems:
```
Should be enough

weldx/transformations.py
vhirtham: > tested locally and seems to raise correctly

copy the local test and make it a real one ;) We don't need to think of a good one for now but a quick and dirty test is better than no test :p

can be "public". The exception catches the only problem

vhirtham (approved):

weldx/transformations.py
CagtayFabry: test added :)

CagtayFabry:
Test don't really seem to work from forks 😞

Issue #197, opened 2020-11-18
3D geometry/mesh representation


CagtayFabry:
Define a container class that can hold 3D data in different formats (depending on the source/use case) and transform to one or multiple default formats (pointcloud) for plotting, analysis etc.

Simplest use case would be to support Point cloud data, examples to look at:
- ply format
- vtk format

I imagine a simple xarray Dataset with x,y,z coordinates would be a good starting point. With vertices and other (optional) data like color as additional variables.

The ASDF schema `Geometry3D` (or similar) should hold the different formats.

Things to consider:
- unit support?
- *add more here :wink:*

One special format should also be a combination of 3D trace +2D Profiles (like our simple welding examples **or** how the 3D scan data is represented). We should discuss the interfaces we need for that.

CagtayFabry:
VTK seems to offer the most support in terms of extensions (k3d, ipygany, vtk tools)

CagtayFabry:
closing for now

PR #223, opened 2020-12-18
update plot and rasterize functions

CagtayFabry:
## Changes
- `geometry.Profile.rasterize` can return list of rasterized shapes
- `geometry.Profile.plot` plots individual line objects for each shape (instead of a single line object)

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


vhirtham (approved):

CagtayFabry:
thanks for looking over this @vhirtham 🚀 
I realized later I could probably just loop over the shapes in my code just as well.. 😄 But it simplifies things a bit

We can take a more in-depth look at this in #197 again

PR #230, opened 2021-01-18
91 asdf external file cf

CagtayFabry:
## Changes
some code changes

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -605,13 +605,15 @@ class ExternalFile:
 
     @staticmethod
     def calculate_hash(
-        path: Union[str, Path, bytes], algorithm: str, buffer_size: int = 65536
+        path_or_buffer: Union[str, Path, bytes],
+        algorithm: str,
+        buffer_size: int = 65536,
     ) -> str:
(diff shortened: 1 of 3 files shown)


CagtayFabry:
you can merge into the main PR branch if you agree with the changes @vhirtham :)

weldx/core.py
```diff
@@ -595,94 +597,44 @@ def __post_init__(self):
...
+        path: Union[str, Path, bytes], algorithm: str, buffer_size: int = 65536
     ) -> str:
         """Calculate the hash of a file.
 
         Parameters
         ----------
-        path : Union[str, Path]
-            Path of the file
+        path : Union[str, Path, bytes]
+            Path of the file or buffer as bytes
```
vhirtham, line 615: If we combine both functions I think it would be cleaner if we separate it into `path` and `buffer` or give it a more generalized name like `file`.

CagtayFabry: yes, `file` or `path_or_buffer` is fine I think

vhirtham: `path_or_buffer` is probably the clearest

PR #231, opened 2021-01-25
129 lcs plot

### LINKED ISSUE #129: add LCS plot functions
add different plot functions to visualize `LocalCoordinateSystems` in the `transformations` package

### static
- default static 3D plot
- nice axis labels (support pint quantities)
- if time-dependent plot trajectory as line (with direction)

### dynamic
- based on static plot
- use widget output/backend
- add widget slider to move through timeindex

If possible define `_ipython_display_` to dynamically detect matplotlib backend (dynamic / `widget`) and choose accordingly
```python
import matplotlib
matplotlib.get_backend()
```
https://stackoverflow.com/a/15346737/11242411

see `weldx.visualization.plot_coordinate_system`

vhirtham:
## Changes
_Describe changes in this PR_

## Related Issues
Closes #129

## Checks
- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


vhirtham:
Here is a quick test script for the CSM plot function:

~~~ python
import matplotlib.pyplot as plt
import pandas as pd

csm.create_cs("lcs 1", "root", coordinates=[2, 5, 6])
csm.create_cs_from_euler(
    "lcs 2",
    "root",
    sequence="z",
    angles=[0, 90, 270],
    degrees=True,
    coordinates=[[4, -2, 1], [-5, 6, 2], [3, 4, 2]],
    time=["2010-02-01", "2010-02-05", "2010-02-08"],
)

new_times = pd.date_range(start="2010-02-01", end="2010-02-8", freq="6H")

csm.plot_coordinate_systems(
    reference_system="root", time=new_times, show_axes=True, show_trace=True
)
plt.show()
~~~

It's really interesting to see the trajectories if you change the reference coordinate system in the plot function to "lcs 1" :D

weldx/transformations.py
```diff
@@ -1154,39 +1188,53 @@ def invert(self) -> "LocalCoordinateSystem":
...
+        axes=None,
+        color=None,
+        label=None,
+        time=None,
+        time_ref=None,
+        show_trace=True,
+        show_axes=True,
+    ):
+        if axes is None:
+            _, axes = plt.subplots(subplot_kw={"projection": "3d"})
```
CagtayFabry, line 1202:
```suggestion
            _, axes = plt.subplots(subplot_kw={"projection": "3d", "proj_type": "ortho"})
```

weldx/transformations.py
```diff
@@ -2766,6 +2814,46 @@ def plot(self, ax=None):
...
+        if time is not None:
+            self.interp_time(time=time, time_ref=time_ref).plot_coordinate_systems(
+                axes=axes,
+                reference_system=reference_system,
+                show_trace=show_trace,
+                show_axes=show_axes,
+            )
+        else:
+            if axes is None:
+                _, axes = plt.subplots(subplot_kw={"projection": "3d"})
```
CagtayFabry, line 2835:
```suggestion
                _, axes = plt.subplots(subplot_kw={"projection": "3d", "proj_type": "ortho"})
```

weldx/transformations.py
```diff
@@ -2772,6 +2877,89 @@ def plot(self, ax=None):
...
+                limits=limits,
+                time=time,
+                time_ref=time_ref,
+                show_labels=show_labels,
+                show_origins=show_origins,
+                show_traces=show_traces,
+                show_vectors=show_vectors,
+            )
+        elif backend == "mpl":
+            set_ax_eq = axes is None
```
CagtayFabry, line 2943: I would avoid setting axes=equal by default as much as possible, it's good for small examples but messes with visibility for the welding applications

weldx/visualization.py
```diff
@@ -1,10 +1,104 @@
...
+    0x00AA00,
+    0x0000FF,
+    0xAAAA00,
+    0xFF00FF,
+    0x00FFFF,
+    # todo: add colors manually or find better solution since random colors are not
+    #       guaranteed to have good visibility and to differ significantly from already
+    #       existing colors
+    *[color_rgb_to_int(random_color_rgb()) for _ in range(100)],
+]
```
CagtayFabry, line 43: I suggest using values from one of matplotlibs built in colormaps, e.g. [`tab20`](https://matplotlib.org/3.1.0/tutorials/colors/colormaps.html#palettable) looks good `plt.cm.tab20(np.arange(20))`

weldx/visualization.py
```diff
@@ -1,10 +1,104 @@
...
+
+
+def color_generator_function():
+    for color in _color_list:
+        yield color
+    raise Exception("No more colors available.")
```
CagtayFabry, line 49: I like this, if possible just start from beginning again?

weldx/visualization.py
```diff
@@ -83,3 +181,575 @@ def set_axes_equal(axes):
...
+            show_traces,
+            show_vectors,
+        )
+
+        # create plot using dict comprehension
+        self._color_generator = color_generator_function()
+        plot = k3d.plot(
+            grid_auto_fit=False,
+            camera_auto_fit=False,
+            grid=(limits[0], limits[0], limits[0], limits[1], limits[1], limits[1]),
```
CagtayFabry, line 556: can `k3d` do non equal grid sizes?

weldx/visualization.py
```diff
@@ -83,3 +181,575 @@ def set_axes_equal(axes):
...
+            show_origins,
+            show_traces,
+            show_vectors,
+        )
+
+        # create plot using dict comprehension
+        self._color_generator = color_generator_function()
+        plot = k3d.plot(
+            grid_auto_fit=False,
+            camera_auto_fit=False,
```
CagtayFabry, line 555: Why not `True`? It difficult to position the camera for welding examples this way

weldx/visualization.py
```diff
@@ -83,3 +181,575 @@ def set_axes_equal(axes):
...
+            lcs_name: CoordinateSystemVisualizerK3D(
+                self._csm.get_cs(lcs_name, reference_system),
+                plot,
+                lcs_name,
+                color=self._get_color(lcs_name, colors),
+                show_origin=show_origins,
+                show_trace=show_traces,
+                show_vectors=show_vectors,
+            )
+            for i, lcs_name in enumerate(coordinate_systems)
```
CagtayFabry, line 568:
```suggestion
            for lcs_name in coordinate_systems
```

weldx/transformations.py
vhirtham: Well, this is a complicated topic. I see your point, but a simple fix is just setting new limits to the axes (forgot to return it but one can use `gca`). If we don't use equal axes, the coordinate systems get skewed with the current implementation, which looks strange.  So at the moment, we have the choice between: 

- default equal axes and adjust them to larger objects later
- default adjusted size with skewed coordinate systems and call `axis_equal` later

However, I have some ideas on how we can tackle this a bit better, but I think the most convenient fix, for now, would be to add another function parameter to enable/disable `axis_equal`. 

I am curious if k3d provides the possibility to distort a specific axis... currently, all plots have equal axes.

weldx/visualization.py
vhirtham: I will take a look at it

weldx/visualization.py
vhirtham: if you mean if it runs out of colors, yep:
~~~ python
while True:
    for color in _color_list:
        yield color
~~~

If you want to reset it manually, you have to reset your variable that holds the generator:

~~~ python
my_generator = color_generator_function()
~~~

weldx/visualization.py
vhirtham: Yes, but that only changes the appearance of the rendered grid and doesn't apply any distortion to the rendered objects. I have to see if there are any possibilities to distort the axes. Guess, they have it somewhere ;)

weldx/visualization.py
vhirtham: Because I never deleted those lines 😆  I will remove them. Additionally, I will provide the option to access the plot object directly so that one can adjust it as needed.

weldx/asdf/tags/weldx/core/geometry/point_cloud.py
```diff
@@ -29,7 +31,9 @@ def to_tree(cls, node: PointCloud, ctx):
             'weldx.core.ExternalFile' type to be serialized.
 
         """
-        return node.__dict__
+        tree = deepcopy(node.__dict__)
```
marscher, line 34: just out of curiosity, does asdf modify the passed dict?

vhirtham: Yes. It replaces all python objects in the dict with a dictionary representation of this object. For some reason (I am not totally aware of its purpose) the to tree method gets called twice when you serialize an object. However, in this case, it's really helpful since it usually creates an error during the second execution when the dictionary of a dataclass gets overwritten.

marscher: calling it twice for the same instance sounds like a bug to me.

vhirtham: Well, as far as I remember, it is not. For some reason, they had to do it twice. I just can't remember the specifics anymore. I think it had something to do with some things that couldn't be evaluated during the first pass.

.github/workflows/pytest.yml
```diff
@@ -1,5 +1,5 @@
 name: pytest
-on: [push, pull_request]
+on: [push]
```
marscher, line 2: it seemed mandatory to execute the tests. At least in #241 and #238 no tests and code analysis steps were executed without the pull_request trigger.

vhirtham: I know but for some reason, it's different on this branch. I guess it was some weird behavior on the other PR.

tests/test_geometry.py
```diff
@@ -2773,3 +2775,63 @@ def test_geometry_rasterization_profile_interpolation():
...
             )
             assert ut.vector_is_close(data[:, idx_0 + j], point_exp)
+
+
+# --------------------------------------------------------------------------------------
+# SpatialData
+# --------------------------------------------------------------------------------------
+
+
+class TestPointCloud:
```
CagtayFabry, line 2785:
```suggestion
class TestSpatialData:
```

weldx/visualization.py
```diff
@@ -83,3 +324,1128 @@ def set_axes_equal(axes):
...
+
+        """
+        coordinates, orientation = self._get_coordinates_and_orientation(lcs)
+        self._lcs = lcs
+        self._color = color
+
+        self._vectors = k3d.vectors(
+            origins=[coordinates for _ in range(3)],
+            vectors=orientation.transpose(),
+            colors=[[0xFF0000, 0xFF0000], [0x00FF00, 0x00FF00], [0x0000FF, 0x0000FF]],
```
CagtayFabry, line 675: what do you think about defining and using constants like `RGB_RED=0xFF0000` in `visualization.py` ?

weldx/visualization.py
```diff
@@ -83,3 +324,1128 @@ def set_axes_equal(axes):
...
+            The orientation
+
+        """
+        coordinates = np.array(lcs.coordinates.values, dtype="float32")
+        orientation = np.array(lcs.orientation.values, dtype="float32")
+        if coordinates.ndim > 1:
+            coordinates = coordinates[index]
+
+        if orientation.ndim > 2:
+            orientation = orientation[index]
```
CagtayFabry, line 761: How about some xarray indexing? 🧙 (I think this should work 😉) 
```suggestion
        coordinates = lcs.coordinates.isel(time=index, missing_dims="ignore").values.astype("float32")
        orientation = lcs.orientation.isel(time=index, missing_dims="ignore").values.astype("float32")
```

.github/workflows/pytest.yml
CagtayFabry: do we add this back in before merging?

CagtayFabry (changes requested):

vhirtham:
> do we add this back in before merging?

@CagtayFabry Not sure here. In this PR, adding `pull_request` double triggered the actions. So I removed it.

weldx/visualization.py
vhirtham: Seems reasonable. Will do that.

weldx/visualization.py
vhirtham: Seems to work :)

CagtayFabry (approved):

marscher:
:+1:

PR #234, opened 2021-02-03
add PointCloud and simple geometry triangulation

CagtayFabry:
## Changes
add the `stack` option to rasterization functions in `geometry` (old behavior via `stack=True` as per default)

with `stack=False`, rasterization will return a list of arrays, one list item for each profile p. The returned arrays will also be shaped `(s,3,n)` with `s` being the number of profiles (trace rasterization steps) and `n` the number of points for each profile.
With `stack=True` the the old behavior is preserved returning a single pointcloud array of shape `(3,p*s*n)`

This only works for static profile interpolation ! I only implemented it for this case.

The utility function `_triangulate_geometry` creates a simple mesh triangulation shaped `(m,3)` for this case and flattens the data into shape `(s*n,3)` for each profile.
Small example notebook included

## Related Issues
Closes # (add issue numbers)

## Checks
- [x] updated CHANGELOG.md
- [ ] updated tests
- [x] updated doc/


tests/test_core.py
```diff
@@ -435,3 +435,8 @@ def test_interp_time_exceptions(ts, time, exception_type, test_name):
         """Test the exceptions of the 'set_parameter' method."""
         with pytest.raises(exception_type):
             ts.interp_time(time)
+
+
+# --------------------------------------------------------------------------------------
+# PointCloud
+# --------------------------------------------------------------------------------------
```
vhirtham, line 442: I will add the missing tests on the other branch :)

weldx/asdf/schemas/weldx.bam.de/weldx/core/geometry/point_cloud-1.0.0.yaml
```diff
@@ -0,0 +1,61 @@
...
+    type: object
+  triangles:
+    description: |
+      An array of index triplets that specify which points form a triangle
+    tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
+    properties:
+      datatype:
+        type: string
+        enum: [uint32, uint64]
+    wx_shape: [(i), 3]
```
vhirtham, line 56: Think that `[(:), 3]` would be enough because we probably don't need data that is stored per triangle. But we can leave it as is since it doesn't hurt.

vhirtham (approved):

weldx/asdf/schemas/weldx.bam.de/weldx/core/geometry/point_cloud-1.0.0.yaml
CagtayFabry: ah yes, no need to use `i` (or any letter for that matter) here you`re right

PR #241, opened 2021-02-18
[ci/gha] use pytest-xdist to use multiple cores

marscher:
## Changes

Use xdist to utilize multiple cpu cores on Github actions containers. Tested locally, that coverage reports are the same. Without slow tests the execution time went down from 128s to 70s.


vhirtham (approved):
LGTM.

PR #243, opened 2021-02-23
only squeeze last axis in ut.mat_vec_mul 

### LINKED ISSUE #239: CSM interp_time doesn't work with scalar values
The following script crashes

~~~ python

csm.create_cs("lcs 1", "root", coordinates=[1, 2, -1])
csm.create_cs("lcs 2", "root", coordinates=[[1, 2, 0], [1, 1, 0]], time=Q_([0, 7], "s"))

csm.interp_time(Q_(1, "s"))
~~~

The error message is:

~~~
ValueError: applied function returned data with unexpected number of dimensions. Received 1 dimension(s) but expected 2 dimensions with names: ('time', 'c')
~~~

Creating a quantity with a list that contains a single item (`Q_([1], "s")`) still fails. Only if a second member is added, the script runs as expected.

CagtayFabry:
## Changes
fixes a bug that would remove all existing (wanted) singleton dimensions

### quick explanation for reference
With a singleton dimension (time in this case), the broadcasting in preparation for multiplication produced the following:

```python
np.matmul(a.data,b.data[...,np.newaxis]).shape
>> (1, 3, 1)
```
The first 1 being the singleton time dimension (we want to keep), the last one the temporary singleton vector dimension

The old `squeeze` removed all singleton dimensions, now the time is missing
```python
np.matmul(a.data,b.data[...,np.newaxis]).squeeze().shape
>> (3,)
```
Since we know which dimension to remove, this is what we want:
```python
np.matmul(a.data,b.data[...,np.newaxis]).squeeze(-1).shape
>> (1, 3)
```

## Related Issues
Closes #239

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


tests/test_transformations.py
```diff
@@ -3713,6 +3713,34 @@ def test_coordinate_system_manager_interp_time():
...
+    # specific coordinate system (single, scalar timestamp) -----------------
+    time_interp_lcs0 = TDI([3], "D")
+    csm_interp_single = csm.interp_time(time_interp_lcs0, None, "lcs_0")
+
+    coords_0_exp = np.array(coords_0_exp)[[0], :]  # modified in prev test !
+    orient_0_exp = np.array(orient_0_exp)[[0], :]  # modified in prev test !
+    lcs_0_in_root_exp = tf.LocalCoordinateSystem(
+        orient_0_exp, coords_0_exp, time_interp_lcs0
+    )
+
```
CagtayFabry, line 3743: I quickly hacked this together from the previous test

not pretty but it does catch the old error without the fix @vhirtham

vhirtham: As long as it does the trick :p

vhirtham (approved):
Deepsource issues are already fixed on the plot branch. So merge afterwards.

.github/workflows/pytest.yml
```diff
@@ -1,5 +1,5 @@
 name: pytest
-on: [push]
+on: [push, pull_request]
```
marscher, line 2: @vhirtham so it is mandatory by now to specify the pr action ;)

CagtayFabry: what did I miss? 😉 
just wanted to add this back in as quickly as possible 😃

vhirtham: still curious why it differs between PRs ❓

PR #247, opened 2021-02-25
split up visualization and transformation modules into smaller pieces

marscher:
## Changes
- unified utility modules to the new name "util"
- split up transformations module into local_cs, cs_manager, rotation and util.
- split up visualization module into impl specific parts for matplotlib and k3d.
- only import heavy-weight pkgs like matplotlib, sympy etc. upon first
  usage. Now import time of weldx is ~1s, before it was almost 4s


weldx/utility.py
```diff
@@ -72,6 +72,16 @@ def inner_decorator(
     return inner_decorator
 
 
+def ureg_check_method(*args):
```
marscher, line 75: accidentally commited to wrong branch

weldx/constants.py
```diff
@@ -12,3 +12,4 @@
 WELDX_UNIT_REGISTRY.define("percent = 0.01*count = %")
 # swap plank constant for hour definition
 WELDX_UNIT_REGISTRY.define("hour = 60*minute = h = hr")
+#WELDX_UNIT_REGISTRY.autoconvert_offset_to_baseunit = True
```
marscher, line 15: remove again

tests/asdf_tests/test_asdf_groove.py
```diff
@@ -83,9 +83,6 @@ def test_asdf_groove_exceptions():
             groove_angle=Q_(50, "deg"),
         )
 
-    with pytest.raises(NotImplementedError):
```
marscher, line 86: wrong branch

weldx/asdf/utils.py
```diff
@@ -95,86 +97,84 @@ def _write_read_buffer(
...
-        code = _get_yaml_header(file)
-
-        formatter = HtmlFormatter()
-        return IPython.display.HTML(
-            '<style type="text/css">{}</style>{}'.format(
-                formatter.get_style_defs(".highlight"),
-                highlight(code, lexer, formatter),
-            )
+    from pygments.formatters.html import HtmlFormatter
+    from IPython.display import HTML
```
CagtayFabry, line 139: The "good" thing about the old implementation wrapped in 

```python
try:
...
except ImportError:
pass
```
was that the function would simply not get defined in the wrong context (without IPython). But running all imports is a bit heavy I agree.
Maybe we can use a construct like
```python
if importlib.util.find_spec("IPython") and importlib.util.find_spec("pygments"):
def notebook_fileprinter(file, lexer="YAML"):
...
```

marscher: So find_spec does not do the heavy import, but only checks if the package is available?

CagtayFabry: yes it should simply return `None` if the package is not installed (and some package metadata if it is) 
https://docs.python.org/3/library/importlib.html#importlib.util.find_spec

marscher: Having functions available only if a certain package is available/importable seems kind of confusing. You would just get an ImportError/AttributeError further downstream, if IPython is not in place. I suggest, we define these tools as stubs if Ipython is not there, and define/implement them for the other case.

CagtayFabry: or just keep your current proposed implementation, calling these functions really doesn't make any sense outside of an IPython of Juypter context

marscher: Now we have stubs, if pygments/ipython is not available and proper implementation for the usual case.

marscher:
mhm, for some PRs CI runs, for some not (shouldn't have merged master again). Somebody keeps flipping the "run ci" bit once in a while :)

CagtayFabry:
> mhm, for some PRs CI runs, for some not (shouldn't have merged master again). Somebody keeps flipping the "run ci" bit once in a while :)

The actions do appear to run ([passing even](https://github.com/BAMWelDX/weldx/actions/runs/618443831)) but somehow the status does not get reported to the "PR check"

CagtayFabry:
we should probably agree on a single format for `utility` submodules, so far I can spot `weldx.utility`, `weldx.asdf.utils`, `weldx.transformations.util`

My votes goes to `util` for brevity 🚀

marscher:
sphinx_type_annotations does not set typing.TYPE_CHECKING to true while inspecting the annotations. Using string annotations like ` from __future__ import annotations ` does not solve this. Further setting it forcefully to true in conf.py makes importing pandas fail. So I currently do not know how to treat this.

marscher:
these 52 deepsource doc issues were introdocued a long time ago. How should we proceed with them? I guess we don't need docstrings in tests, right?

CagtayFabry:
> these 52 deepsource doc issues were introduced a long time ago. How should we proceed with them? I guess we don't need docstrings in tests, right?

Generally we stick to pydocstyle for docstring lintering (ignoring private methods) so these deepsource issues can be ignored. I also think the deepsource run should pass if only docstring/TODO issues arise

weldx/visualization/matplotlib_impl.py
```diff
@@ -173,7 +173,7 @@ def plot_local_coordinate_system_matplotlib(
     label : str
         Name of the coordinate system
     time : pandas.DatetimeIndex, pandas.TimedeltaIndex, List[pandas.Timestamp], or \
-           LocalCoordinateSystem
+           ~weldx.transformations.LocalCoordinateSystem
```
CagtayFabry, line 176: @vhirtham @marscher 
for classes that are ambiguous (because defined in different init namespaces or other reasons) this is the explicit way to link in docstrings
For type hints this is not necessary (since they get resolved from the file imports I guess)

using the `~` will shorten the visible link in the documentation to the final classifyer, see https://stackoverflow.com/a/21297186/11242411

marscher: That looks nice. Thanks a lot for taking a shot on this. Sphinx is sometimes hard to convince to do the simplest things.

marscher: Ain't the convention to omit the types in the docstring, but let it be derived from the type annotation?

marscher:
Thanks @vhirtham for pointing out the testing documentation policy. Documenting complicated tests is a very good practice.

CagtayFabry:
If have touched up the links and docstring formatting here https://github.com/BAMWelDX/weldx/pull/247/commits/768cb899c452eeb44a738b54aa047af4719afc4e for future reference @vhirtham @marscher

weldx/visualization/matplotlib_impl.py
CagtayFabry: yes, if possible types can and should be inferred from the type hints but can be set manually like here. I have left this example in the docstrings since I think in this case it looks nicer than `Union[pandas.DatetimeIndex, pandas.TimedeltaIndex, List[pandas.Timestamp], LocalCoordinateSystem]`

marscher:
Thanks for all the contributions to this. Finally the checks are passing and this is ready for review @vhirtham, @CagtayFabry

weldx/__init__.py
```diff
@@ -14,7 +14,7 @@
     )
 
 # main modules
-import weldx.utility  # import this first to avoid circular dependencies
+import weldx.util  # import this first to avoid circular dependencies
```
vhirtham, line 17: Just a general comment: Should test if this is still the case.

vhirtham (approved):

weldx/visualization/matplotlib_impl.py
marscher: This union format issue is solved by the software. I've seen an option to format it without the Union thing, but can not recall it right now. I think we should stick to the type annotation wherever possible to avoid duplication.

weldx/__init__.py
marscher: I've changed it along the way to revert it back. There is still a circular dep within the util module

doc/conf.py
```diff
@@ -22,11 +22,20 @@
 
 import traitlets
 
+# find weldx from parent path.
+sys.path.insert(0, os.path.abspath("../"))
+
+import typing
+
+import ipywidgets
+import pandas as _
```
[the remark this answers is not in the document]
marscher: because it is not needed downstream (just be loaded prior type checking gets turned on). Sorry for the inconsistency.

PR #248, opened 2021-02-25
Compute groove cross sectional area and welding speed

marscher:
## Changes

Groove types get a property to compute the cross sectional area. The implementation uses sympy.geometry types to wrap up three polygons, two for the left and right work-piece and one for the outer spanning or bounding box. These are
then used to compute the area.

Added compute_welding_speed in welding.util sub-package.

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests


weldx/welding/groove/iso_9692_1.py
```diff
@@ -163,6 +187,22 @@ def to_profile(self, width_default: pint.Quantity = Q_(5, "mm")) -> geo.Profile:
...
         return geo.Profile([shape, shape_r], units=_DEFAULT_LEN_UNIT)
 
+    @WELDX_UNIT_REGISTRY.check(None, '[length]', '[length]/[time]', '[length]', '[length]')
+    #TODO: extend the ckeck decorator to check on output! # ret='[length]/[time]')
+    #TODO: thinkness of wire??
+    def compute_welding_speed(self, seam_length: pint.Quantity, wire_feed: pint.Quantity,
+                              width_default: pint.Quantity = Q_(5, 'mm'),
+                              wire_thinkness=Q_(1, 'mm')):
+        # TODO: we need all lengths in same units, otherwise extending area to volume etc does not work...
+        WELDX_UNIT_REGISTRY.auto_reduce_dimensions = True
```
CagtayFabry, line 197: I think a final call to `speed.to("mm/s")` or `speed.to_reduced_units()` before return should work as well here

weldx/welding/groove/iso_9692_1.py
```diff
@@ -99,16 +100,39 @@ def plot(
...
         ----------
-        width_default :
+        width_default : pint.Quantity['length']
              optional width to extend each side of the profile (Default value = None)
 
         """
-        raise NotImplementedError("to_profile() must be defined in subclass.")
+        pass
+
+    def compute_welding_speed(self, seam_length: pint.Quantity,
```
CagtayFabry, line 115: I would suggest only defining an (abstract) `area` property/method on the groove types and moving the welding speed calculation to another module

```python
wire_area = np.pi / 4 * wire_diameter
weld_speed = wire_area * wire_feed / groove_area
```

weldx/welding/utility.py
```diff
@@ -0,0 +1,40 @@
...
+    wire_diameter: pint.Quantity["length"]
+        diameter of welding wire
+
+    Returns
+    -------
+    speed: pint.Quantity["length/time"]
+    """
+    groove_area = groove.cross_sect_area
+    wire_area = np.pi / 4 * wire_diameter
+    weld_speed = wire_area * wire_feed / groove_area * seam_length
```
CagtayFabry, line 37: I think `seam_length` is not needed to calculate the welding speed (it is in fact no necessary for this function, I should clarify that in the use-case)

weldx/welding/utility.py
```diff
@@ -0,0 +1,40 @@
...
+        feed of the wire
+    wire_diameter: pint.Quantity["length"]
+        diameter of welding wire
+
+    Returns
+    -------
+    speed: pint.Quantity["length/time"]
+    """
+    groove_area = groove.cross_sect_area
+    wire_area = np.pi / 4 * wire_diameter
```
CagtayFabry, line 36:
```suggestion
    wire_area = np.pi / 4 * wire_diameter**2
```
my bad

marscher: mine too, as I didn't noticed the mistake :) BTW how do you directly propose such changes to be committed. This is really handy.

weldx/welding/utility.py
marscher: I thought one would like to compute the speed for the whole process, which would then have included the whole volume to fill up. Thanks for clarifying that.

weldx/welding/utility.py
vhirtham: @marscher if you mark a comment during review, there is a small button with +-. The tooltip says that you can also press ctrl+g. This generates the following markdown:
~~~
```suggestion
    $THE_NEW_CODE
```
~~~

marscher:
Thanks for sharing! This is really helpful :)

weldx/welding/groove/iso_9692_1.py
```diff
@@ -163,6 +211,16 @@ def to_profile(self, width_default: pint.Quantity = Q_(5, "mm")) -> geo.Profile:
 
         return geo.Profile([shape, shape_r], units=_DEFAULT_LEN_UNIT)
 
+    @property
+    def cross_sect_area(self):
+        from sympy.geometry import Polygon
+
+        b = float(self.b.magnitude)
+        t = float(self.t.magnitude)
```
CagtayFabry, line 219: I guess there is no way to create a `Polygon` from `pint` quantities directly? That would be super convenient.
`pint` doesn't seem to work nicely with `Point2D`

marscher: No, unfortunately the geometry classes of sympy dont play along with pint.

marscher:
pydocstyle is a static linter, so it cannot handle the (dynamic) inheritance of docstrings. They provide an option to ignore attributes with given decorators, but I've chosen to implement the inherit_docstring decorator to fix the whole class instead of every individual method.

The only workaround would be to decorate every individual method/property which seems tedious to me.

xref: https://github.com/PyCQA/pydocstyle/issues/274

weldx/welding/utility.py
```diff
@@ -0,0 +1,41 @@
...
+    groove: IsoBaseGroove,
+    seam_length: pint.Quantity,
+    wire_feed: pint.Quantity,
+    wire_diameter: pint.Quantity,
+):
+    """Computes how fast the torch has to be moved to fill the gap of the groove.
+
+    Parameters
+    ----------
+    groove: IsoBaseGroove
```
CagtayFabry, line 21:
```suggestion
    groove
```
I suggest we omit the type info from the docstrings in cases where they match the function type hints (like here)
The missing type info will be inherited from type hints in the documentation.
https://weldx--248.org.readthedocs.build/en/248/_autosummary/weldx.welding.utility.compute_welding_speed.html#weldx.welding.utility.compute_welding_speed

This way it is easier to keep everything synced up and the intention is clear when a specific type info is listed in the docstring (showing a "clearer" info than the type hints like for this case)
I like the `pint.Quantity["length"]` approach btw

weldx/welding/utility.py
```diff
@@ -35,7 +32,7 @@ def compute_welding_speed(
     """
     groove_area = groove.cross_sect_area
     wire_area = np.pi / 4 * wire_diameter ** 2
-    weld_speed = wire_area * wire_feed / groove_area * seam_length
```
CagtayFabry: I have removed this, output should now be `[length]/[time]` @marscher

marscher: Thanks. I should also add a small test for that.

CagtayFabry: I feel like it should be safe the way that it is now.
The unit error would raise on the `wx_unit` validation here anyway https://github.com/BAMWelDX/weldx/blob/2bda4dd9e0b0a909a1ba6054a7e3cdd3d5cd9ea9/weldx/asdf/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-1.0.0.schema.yaml#L29-L34

CagtayFabry (approved):
I think this should be good to go now 🚀 
please check my last commits @marscher

marscher:
Thank you very much for your help! Should we put the FFGroove cross section area computation on the issue list?

CagtayFabry:
> Thank you very much for your help! Should we put the FFGroove cross section area computation on the issue list?

Ah yes, that's a good idea actually since we are free to implement a less generic version for these special cases
bat-signaling @mark141 🦇 🦸‍♂️ 😉

PR #265, opened 2021-03-10
Add from_file and write_to_file to SpatialData

vhirtham:
## Changes

- add meshio as new dependency
- add `from_file` and `write_to_file` to `SpatialData`

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/geometry.py
```diff
@@ -2328,6 +2329,26 @@ def __post_init__(self):
             if not self.triangles.ndim == 2:
                 raise ValueError("SpatialData triangulation must be a 2d array")
 
+    @staticmethod
+    def from_file(file_name: str) -> "SpatialData":
```
CagtayFabry, line 2333: `pathlib.Path` also seems to be supported by meshio so we might add that to the type hints

vhirtham: yes. I am currently experimenting and writing the test ;)

marscher (approved):
LGTM!

weldx/geometry.py
marscher: It looks like the library is leaking file handles. But using a sane OS this ain't a problem :p

marscher:
> @vhirtham I stumbled upon [this](https://github.com/nschloe/meshio/blob/main/meshio/_files.py) nice utility in `meshio` that might come in handy for other file operations later, just dropping it here :)

This is actually meant to be used when using their readers/writers to avoid leaking file handles!

CagtayFabry (approved):

PR #267, opened 2021-03-11
fix welding_schema

CagtayFabry:
## Changes
fix imports in `welding_schema.py` script


marscher:
ah sorry, I missed that one.

PR #268, opened 2021-03-11
modify CSM merge for non datetime subsytems

CagtayFabry:
## Changes
- allow time dependent subsystems that don't use absolute times to be merged into CSM instances with a set reference time
- make sure `csm.time_union()` returns correct index type without explicit reference time

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/


CagtayFabry:
@vhirtham the problem is cause when `csm.uses_absolute_times` is true (there is some `DateTimeIndex` involved) but  `csm.reference_time` is false (there is no reference time set on a CSM level)

I think this can only happen with a single reference_time LCS involved but not sure about the checks

This fix will return a valid `pd.DateTimeIndex` in these cases

CagtayFabry:
@vhirtham changed error behavior of `csm.merge()` as discussed

vhirtham (approved):

PR #274, opened 2021-03-12
add ASDF schema examples

### LINKED ISSUE #254: add ASDF schema examples
we should expand the ASDF schemas with examples

ideally at  least one example for all variations in case of `oneOf` etc.

CagtayFabry:
## Changes
- add examples to ASDF schemas

## Related Issues
Closes #254 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/.github/workflows/pytest_asdf.yml
+++ b/.github/workflows/pytest_asdf.yml
@@ -20,4 +20,4 @@ jobs:
           pip install .
       - name: run pytest
         run: |
-          pytest --asdf-tests --ignore=tests/
+          pytest --asdf-tests --ignore=tests/ --no-cov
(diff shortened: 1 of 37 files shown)


vhirtham (approved):
Just had a look at the descriptions since I guess the code is automatically generated. Looks good so far. ❤️

marscher:
It is very nice to have such examples. Just a general question, how would we adopt these examples when the schema changes? If @vhirtham is right about generating the examples from code, it would be cool, if you should share this snippet as well to quickly update them in the future.

CagtayFabry:
They are not automatically generated, I took them from existing files or just quickly created them.
If the schema changes they must be updated accordingly (we would create a new file version string for that anyway)

The good thing is that the asdf-pytest plugin checks if the examples can be (de)serialized so the CI should spot if we break anything

marscher:
On 3/16/21 3:20 PM, Cagtay Fabry wrote:
> They are not automatically generated, I took them from existing files or just quickly created them.
> If the schema changes they must be updated accordingly (we would create a new file version string for that anyway)
> 
> The good thing is that the asdf-pytest plugin checks if the examples can be (de)serialized so the CI should spot if we break anything
> 
That's good news, that CI will break then. Thanks!

PR #275, opened 2021-03-12
Fix a bug in CSM interp_time

vhirtham:
## Changes

Before this PR, calling `CoordinateSystemManager.interp_time` with absolute times would fail if the CSM had coordinate systems without a reference time even though the CSM has one.

## Checks

- [x] ~~updated CHANGELOG.md~~
- [x] updated tests
- [x] ~~updated doc/~~

### DIFF
--- a/tests/test_transformations.py
+++ b/tests/test_transformations.py
@@ -3992,7 +3992,7 @@ def test_coordinate_system_manager_interp_time():
         check_coordinate_systems_close(lcs, exp)
         check_coordinate_systems_close(lcs_inv, exp_inv)
 
-        # Related to pull request #INSERT_NUMBER. This assures that interp time works
+        # Related to pull request #275. This assures that interp time works
         # correctly if some coordinate systems have no own reference time, but the CSM
         # does.
         lcs1 = tf.LocalCoordinateSystem(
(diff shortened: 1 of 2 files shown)


CagtayFabry:
good catch ! 👍

CagtayFabry:
just to clarify: the case where a `CSM`  instance with one or more TimedeltaIndex LCS does not have a specific reference time but more than one LCS with internal DateTimeIndex is forbidden right?

What happens if for the CSM `self.has_reference_time == False` but `self.uses_absolute_times == True`? Do we take the LCS internal Timestamp for reference?

vhirtham:
> just to clarify: the case where a `CSM` instance with one or more TimedeltaIndex LCS does not have a specific reference time but more than one LCS with internal DateTimeIndex is forbidden right?

As far as I remember mixed cases are generally forbidden if the CSM has no reference time.

> What happens if for the CSM `self.has_reference_time == False` but `self.uses_absolute_times == True`? Do we take the LCS internal Timestamp for reference?

Yes, that is handled by the LCS `interp_time` which is called by the CSM. The actual error originated from the LCS function itself since it does not allow mixed cases. So without the new lines, we had absolute times as input to an LCS without reference time. 
This couldn't happen before we updated the CSM's `interp_time`

CagtayFabry (approved):
LGTM 👍

PR #276, opened 2021-03-12
move get_yaml_header to be a public function

CagtayFabry:
## Changes
always define `get_yaml_header` and make public

### DIFF
--- a/weldx/asdf/util.py
+++ b/weldx/asdf/util.py
@@ -13,6 +13,7 @@ __all__ = [
     "read_buffer",
     "write_buffer",
     "write_read_buffer",
+    "get_yaml_header",
     "asdf_json_repr",
     "notebook_fileprinter",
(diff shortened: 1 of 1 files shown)


weldx/asdf/util.py
```diff
@@ -112,6 +113,32 @@ def _write_read_buffer(
     return _read_buffer(buffer, open_kwargs)
 
 
+def get_yaml_header(file) -> str:
```
vhirtham, line 116: Type hints for `file`? What about `Path`?

vhirtham (approved):

marscher (approved):

weldx/asdf/util.py
CagtayFabry: everything that works with `read` should work :)
I will leave it to the docstring for now

PR #279, opened 2021-03-15
allow inline scalar coordinates/variables

CagtayFabry:
## Changes
allows to store scalar values as `number` in `variable-1.0.0.yaml`
This is necessary because scalar values like `{'shape': [], 'data': 1409875200000000000, 'datatype': 'int64'}` are not covered by ASDF `ndarray`  (data must be list)

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/

### DIFF
--- a/weldx/asdf/tags/weldx/core/common_types.py
+++ b/weldx/asdf/tags/weldx/core/common_types.py
@@ -145,6 +145,8 @@ class VariableTypeASDF(WeldxType):
             data = node.data
         dtype = node.data.dtype.str
         data = cls.convert_time_dtypes(data=data)
+        if not data.shape:  # scalar
+            data = np.asscalar(data)
         tree = {
             "name": node.name,
             "dimensions": node.dimensions,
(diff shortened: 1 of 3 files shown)


vhirtham (approved):

weldx/asdf/tags/weldx/core/common_types.py
```diff
@@ -145,6 +145,8 @@ def to_tree(cls, node: Variable, ctx):
             data = node.data
         dtype = node.data.dtype.str
         data = cls.convert_time_dtypes(data=data)
+        if not data.shape:  # scalar
```
marscher, line 148: So data always has a shape information attached, if it is a scalar? Is this yet covered by a test?

marscher: sorry, I haven't checked the coverage report. Thanks for fixing this.

marscher (approved):

weldx/asdf/tags/weldx/core/common_types.py
CagtayFabry: data should always be a `np.ndarray` here, just need to catch scalar values for the correct ASDF serialization

PR #284, opened 2021-03-15
[transformations] use type aliases where appropriate

### LINKED ISSUE #272: use type aliases
We should consider using [type aliases](https://docs.python.org/3.8/library/typing.html#type-aliases) (note we're on py3.8) in some modules with more complex type annotations

especially some common larger unions like `Union[pd.TimedeltaIndex, pd.DateTimeIndex, xr.DataArray]` could become simplified and more explicit where the same inputs are allowed

this will also reduce the work to be done once we switch to python 3.9 type hints

marscher:
Fixes #272

### DIFF
--- a/weldx/constants.py
+++ b/weldx/constants.py
@@ -6,9 +6,15 @@ WELDX_UNIT_REGISTRY = UnitRegistry(
     preprocessors=[lambda string: string.replace("%", "percent")],  # allow %-sign
     force_ndarray_like=True,
 )
+
 WELDX_QUANTITY = WELDX_UNIT_REGISTRY.Quantity
 
 # add percent unit
 WELDX_UNIT_REGISTRY.define("percent = 0.01*count = %")
 # swap plank constant for hour definition
 WELDX_UNIT_REGISTRY.define("hour = 60*minute = h = hr")
+
+__all__ = (
+    WELDX_UNIT_REGISTRY,
+    WELDX_QUANTITY,
+)
--- a/weldx/constants.py
+++ b/weldx/constants.py
@@ -15,6 +15,6 @@ WELDX_UNIT_REGISTRY.define("percent = 0.01*count = %")
 WELDX_UNIT_REGISTRY.define("hour = 60*minute = h = hr")
 
 __all__ = (
-    WELDX_UNIT_REGISTRY,
-    WELDX_QUANTITY,
+    "WELDX_UNIT_REGISTRY",
+    "WELDX_QUANTITY",
 )
--- a/weldx/transformations/rotation.py
+++ b/weldx/transformations/rotation.py
@@ -69,25 +69,27 @@ class WXRotation(_Rotation):
     """Wrapper for creating meta-tagged Scipy.Rotation objects."""
 
     @classmethod
-    def from_quat(cls, quat: np.ndarray) -> "WXRotation":
+    def from_quat(cls, quat: np.ndarray) -> "WXRotation":  # noqa
         rot = super().from_quat(quat)
         setattr(rot, "wx_meta", {"constructor": "from_quat"})
         return rot
 
     @classmethod
-    def from_matrix(cls, matrix: np.ndarray) -> "WXRotation":
+    def from_matrix(cls, matrix: np.ndarray) -> "WXRotation":  # noqa
         rot = super().from_matrix(matrix)
         setattr(rot, "wx_meta", {"constructor": "from_matrix"})
         return rot
 
     @classmethod
-    def from_rotvec(cls, rotvec: np.ndarray) -> "WXRotation":
+    def from_rotvec(cls, rotvec: np.ndarray) -> "WXRotation":  # noqa
         rot = super().from_rotvec(rotvec)
         setattr(rot, "wx_meta", {"constructor": "from_rotvec"})
         return rot
 
     @classmethod
-    def from_euler(cls, seq: str, angles, degrees: bool = False) -> "WXRotation":
+    def from_euler(
+        cls, seq: str, angles, degrees: bool = False
+    ) -> "WXRotation":  # noqa
         rot = super().from_euler(seq=seq, angles=angles, degrees=degrees)
         setattr(
             rot,
--- a/weldx/transformations/rotation.py
+++ b/weldx/transformations/rotation.py
@@ -4,6 +4,7 @@ import numpy as np
 from scipy.spatial.transform import Rotation as _Rotation
 
 from weldx.constants import WELDX_UNIT_REGISTRY as UREG
+from weldx.util import inherit_docstrings
 
 _DEFAULT_LEN_UNIT = UREG.millimeters
 _DEFAULT_ANG_UNIT = UREG.rad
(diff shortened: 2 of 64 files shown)


CagtayFabry:
thanks for starting this @marscher 

I will give these some thought, it should also help streamlining the interfaces a little.
One suggestion I would like to make is a union of types queried like for a `time` axis/attribute (like `xarray` objects but also `LCS`, `CSM`, `TimeSeries` etc.)

weldx/constants.py
```diff
@@ -6,9 +6,15 @@
...
 WELDX_QUANTITY = WELDX_UNIT_REGISTRY.Quantity
 
 # add percent unit
 WELDX_UNIT_REGISTRY.define("percent = 0.01*count = %")
 # swap plank constant for hour definition
 WELDX_UNIT_REGISTRY.define("hour = 60*minute = h = hr")
+
+__all__ = (
+    WELDX_UNIT_REGISTRY,
+    WELDX_QUANTITY,
```
CagtayFabry, line 19: let's see if the docs build with this
```suggestion
    "WELDX_UNIT_REGISTRY",
    "WELDX_QUANTITY",
```

CagtayFabry:
I like how the stacked `Union` gets rendered like [here](https://weldx--284.org.readthedocs.build/en/284/_autosummary/weldx.transformations.CoordinateSystemManager.interp_time.html)

weldx/transformations/util.py
```diff
@@ -286,3 +292,9 @@ def vector_points_to_left_of_vector(vector, vector_reference):
 
     """
     return int(np.sign(np.linalg.det([vector_reference, vector])))
+
+
+coordinate_types = Union[xr.DataArray, np.ndarray, List]
+types_orientation = Union[xr.DataArray, np.ndarray, List[List], Rot]
+time_types = Union[pd.DatetimeIndex, pd.TimedeltaIndex, pint.Quantity]
+types_time_and_lcs = Union[time_types, "weldx.transformations.LocalCoordinateSystem"]
```
vhirtham, line 300: Since we started separating stuff into multiple files, what do you think of adding a file for definitions? 

@marscher @CagtayFabry

vhirtham (approved):

weldx/transformations/util.py
marscher: I can add a new module just for the type aliases, if you think that this would be easier to use.

marscher:
@CagtayFabry I'd be glad, if you could take a look at those sphinx warnings. I think these are related to inferring the return type from the signature (which works for arguments, but somehow not [always] for the return doc string).

CagtayFabry:
> @CagtayFabry I'd be glad, if you could take a look at those sphinx warnings. I think these are related to inferring the return type from the signature (which works for arguments, but somehow not [always] for the return doc string).

I will try to look into it the next couple days
But I've had troubles with indicating the return type as well..

marscher:
Thanks, but I don't want to bother you with this, when you're busy right now. I will dig into it tomorrow.

marscher:
If you want to have a laugh take a look at the sphinx_type_hinting extension. It is extra ordinarily naive. When a signature is multi-line (e.g. is too long and wrapped by black formatter), the return type is simply ignored by the current approach:
https://github.com/agronholm/sphinx-autodoc-typehints/blob/7ad7b209bbd1fe8f76fb0c646fbc08ea5e57c318/sphinx_autodoc_typehints.py#L301
I wonder why they don't use a (multi-line) regular expression to defeat all of these cases. I guess there are more bugs like this within this (horrible) code-base. Is there an alternative to this tool or do we have to stick to it?

vhirtham:
> I wonder why they don't use a (multi-line) regular expression to defeat all of these cases. I guess there are more bugs like this within this (horrible) code-base. Is there an alternative to this tool or do we have to stick to it?

I don't know if there is an alternative, but in my opinion, we don't have to stick with it if you can name a better tool that doesn't take ages to be set up. 😉

CagtayFabry:
For now we can just stick to specifying the return type in the docstring right?

Otherwise I don't mind using a quick fix on a fork or similar if that helps

marscher:
> > I wonder why they don't use a (multi-line) regular expression to defeat all of these cases. I guess there are more bugs like this within this (horrible) code-base. Is there an alternative to this tool or do we have to stick to it?
> 
> I don't know if there is an alternative, but in my opinion, we don't have to stick with it if you can name a better tool that doesn't take ages to be set up. wink

I guess you've chosen this tool since there wasn't an alternative available back then. I've noticed that napoleon does support annotations now and I will test how well this feature works. Another quick question, why do use both napoleon and numpydoc? The latter adds support for rst directives np:function and np:cfunction, which I guess are not used anywhere

marscher:
Well it doesn't infer the return type either and causes other problems as well. It is somehow hard to debug, since the order of extensions is really important and it could happen that another extension kicks out the rtype again.
For this reasons I agree with you @CagtayFabry and will re-include the explicit rtypes again.

weldx/transformations/rotation.py
```diff
@@ -63,45 +64,32 @@ def rotation_matrix_z(angle) -> np.ndarray:
     return _Rotation.from_euler("z", angle).as_matrix()
 
 
+@inherit_docstrings
```
CagtayFabry, line 67: while it would be nice to have the original docstring here we should just stick to linking the scipy docs since sphinx just goes bananas otherwise 🍌

marscher: noticed that as well. I guess scipy does not use warnings as errors policy, but just keeps accumulation wrong formattings since the beginning of time...

CagtayFabry: feel free to use https://github.com/marscher/weldx/pull/1 here, it should be enough for now 👍

CagtayFabry:
I also noticed flake8 is stil pinned to 3.7.* in the static_analysis action
I dont get the errors on 3.8

Should we unpin flake8 or bump it to 3.8/3.9?

weldx/transformations/rotation.py
marscher: Thanks!

marscher:
I don't get this now. Flake8 runs on py3.8 according to the status report on GH.

CagtayFabry:
> 
> 
> I don't get this now. Flake8 runs on py3.8 according to the status report on GH.

Oh I meant the flake8 version (which is dangerously close to the current python versions lol)
We currently `pip install flake8=3.7.*`

CagtayFabry:
I have added intersphinx mappings to `k3d` (removed k3d from `nitpick-ignore`) but still no fix for `matplotlib.axes._axes.Axes` ...

So ignoring them after all @vhirtham @marscher

weldx/transformations/types.py
```diff
@@ -13,8 +13,8 @@
 
 types_coordinates = Union[xr.DataArray, np.ndarray, List]
 types_orientation = Union[xr.DataArray, np.ndarray, List[List], Rotation]
-types_time = Union[pd.DatetimeIndex, pd.TimedeltaIndex, pint.Quantity]
-types_time_and_lcs = Union[types_time, "weldx.transformations.LocalCoordinateSystem"]
+types_timeindex = Union[pd.DatetimeIndex, pd.TimedeltaIndex, pint.Quantity]
```
marscher, line 16: how can a quanity be used as an index?

CagtayFabry: array quantites with dimension ["time"] should get converted to `pd.TimedeltaIndex`

marscher: this means an xarray, right? So a plain quantity is actually misleading here?

weldx/visualization/k3d_impl.py
```diff
@@ -60,14 +62,14 @@ def _create_model_matrix(
...
+    coordinates :
         The coordinates of the origin
-    orientation : numpy.ndarray
+    orientation :
         The orientation of the coordinate system
 
     Returns
     -------
-    numpy.ndarray :
+    matrix :
```
marscher, line 72:
```suggestion
    np.ndarray:
```

Issue #294, opened 2021-03-19
deprecate rotation_matrix_* functions


CagtayFabry:
I think we can deprecate the `transformations.rotation_matrix_*` functions in favor of `WXRotation.as_matrix` to allow more input options and clean up the API a bit

They are mostly used for testing

marscher:
For the aid of deprecating funtions, at least in my experience, a deprecation class/function decorator is very useful. It shows a warning indicating that the called object is deprecated and can be turned into a an error via the warnings module to clean up all usages within the code base. Users will just see the warning for a couple of versions and can adopt their code accordingly. 

Back to topic, I agree that these functions are a indeed a duplication of features and should be replaced by scipy Rotations.

vhirtham:
Since we are the only users so far, I think we can just remove it. However, having a deprecated decorator is probably a nice thing for the future. I'll add one...

CagtayFabry:
yes this should only "break" internal code so removing should be fine :)
we can redefine these in the tests directly to simplify refactoring

PR #295, opened 2021-03-22
Add deprecation decorator

vhirtham:
## Changes

Add a decorator to mark functions as deprecated

## Related Issues

#294 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/weldx/util.py
+++ b/weldx/util.py
@@ -23,6 +23,10 @@ if TYPE_CHECKING:  # pragma: no cover
     import weldx.transformations as tf
 
 
+class WeldxDeprecationWarning(DeprecationWarning):
+    """Deprecation warning type."""
+
+
 def deprecated(since: str = None, removed: str = None, message: str = None) -> Callable:
     """Mark a functions as deprecated.
 
@@ -59,13 +63,7 @@ def deprecated(since: str = None, removed: str = None, message: str = None) -> C
             if message is not None:
                 wm += message
 
-            warnings.simplefilter("always", DeprecationWarning)  # turn off filter
-            warnings.warn(
-                wm,
-                category=DeprecationWarning,
-                stacklevel=2,
-            )
-            warnings.simplefilter("default", DeprecationWarning)  # reset filter
+            warnings.warn(wm, category=WeldxDeprecationWarning, stacklevel=2)
             return func(*args, **kwargs)
 
         return _new_func
--- a/weldx/util.py
+++ b/weldx/util.py
@@ -23,19 +23,24 @@ if TYPE_CHECKING:  # pragma: no cover
     import weldx.transformations as tf
 
 
-def deprecated(func: Callable):
+def deprecated(func: Callable) -> Callable:
     """Mark a functions as deprecated.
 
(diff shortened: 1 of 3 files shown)


vhirtham:
Alternatively, we can use the [deprecation package](https://deprecation.readthedocs.io/en/latest/) which is a better solution I guess.

weldx/util.py
```diff
@@ -23,6 +23,36 @@
...
+    Parameters
+    ----------
+    func :
+        The wrapped function
+
+    """
+
+    @wraps(func)
+    def _new_func(*args, **kwargs):
+        warnings.simplefilter("always", DeprecationWarning)  # turn off filter
```
marscher, line 44: If we go for a custom thing, we should derive from DeprecationWarning in order to not collide with other packages (which might raise a lot of these). Resetting the filter internally ain't a good thing IMO.

marscher: It would also be great to have a simple message prepended which feature to use in the future.

vhirtham: We can build more utility into the function. Just wanted to wait until we decided to go with our own approach before investing more time into it.

weldx/util.py
```diff
@@ -23,6 +23,36 @@
...
+    This decorator emits a warning when the function is used.
+
+    Notes
+    -----
+    Original source: https://stackoverflow.com/a/30253848/6700329
+
+    Parameters
+    ----------
+    func :
+        The wrapped function
```
CagtayFabry, line 38: In general I think we should always start by listing the `Parameters` and `Returns` sections first after the function/class description just to make it easy to spot in the documentation

weldx/util.py
marscher: I think we should go for a custom impl, since this feature demands not too much code and we should avoid adding more dependencies.

vhirtham: Guess you are right. I'll add some useful features to the decorator.

weldx/util.py
```diff
@@ -23,34 +23,54 @@
...
+            if message is not None:
+                wm += message
+
+            warnings.simplefilter("always", DeprecationWarning)  # turn off filter
+            warnings.warn(
+                wm,
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
+            warnings.simplefilter("default", DeprecationWarning)  # reset filter
```
marscher, line 68: This should be the old state of the filter, not "default". This would create a hard to debug situation.

vhirtham: Take a look at [this StackOverflow question](https://stackoverflow.com/questions/2536307/decorators-in-the-python-standard-lib-deprecated-specifically/30253848#30253848). Both of the highest-rated answers do it this way. Can't tell you why, but there must be a reason for it. I am not familiar with the warning system in python so I just copied the corresponding sections.

vhirtham: I removed the filter settings and added our own deprecation warning type as discussed.

marscher (approved):
LGTM

CagtayFabry (approved):
nice :)

PR #299, opened 2021-03-24
Coverage notebooks

### LINKED ISSUE #264: Measure code coverage within notebooks
The codecov service used to gather coverage results from continuous integration (CI) offers the feature of combining multiple reports into a single one. We should use this and merge the (not yet performed) execution of the example notebooks into the main report. We can consider recording coverage while executing the notebooks for the sphinx doc or have a separate job to do it. The latter would be "cleaner", but could cause more delay for CI.

marscher:
## Changes

measure code coverage in notebooks

## Related Issues

Closes #264


marscher:
the individual coverage reports are in place (doc/tutorials) but are not being recognized by the codecov action. Eventually these needs to be converted and merged into a single xml file?

CagtayFabry:
nice to see notebook coverage 👍 

since this currently runs in the docs context it might be difficult to merge the different coverage reports
maybe we could integrate it in the pytest action instead using [nbval](https://github.com/computationalmodelling/nbval/) ?

maybe only use a single pytest setup of the matrix to run the "full" test suite including notebooks?

marscher:
nbval is used to perform regression tests for notebooks (e.g. compare the output stored in the notebook with the actual execution results). Eventually one can turn this feature completely off. I will check it out.

CagtayFabry:
yes we don't need the output comparison, so should `py.test --nbval-lax` be enough https://github.com/computationalmodelling/nbval/#execution

marscher:
I put the coverage measurement in the docs job in order not to increase the testing time, since the notebooks are executed in docs as well. I'm pretty sure that codecov will pick this up correctly, once I sort out the mistake about the missing files.
Using nbval on the other hand has the advantage of an already well established infrastructure.

.github/workflows/pytest.yml
```diff
@@ -42,11 +42,14 @@ jobs:
...
         run: |
-          pip install pytest pytest-cov pytest-xdist setuptools_scm
+          pip install wheel
+          pip install pytest-cov pytest-xdist setuptools_scm nbval
           pip install -e .
       - name: run pytest
         run: |
-          pytest --runslow -n2
+          pytest -n2 --dist loadscope --nbval --cov=weldx --current-env tutorials/
+          pytest --runslow -n2 --cov-append --cov=weldx tests/
```
CagtayFabry, line 51: we do we need to set `--cov=weldx` here? It should get set from `setup.cfg`

marscher: as far as I can see, there is no such setting in setup.cfg. It should contains this, IMO:

```
[coverage:run]
include = weldx
```

I wonder how it worked before, as pytest would have recorded the coverage for all code.

CagtayFabry: it is included in the pytest section
https://github.com/BAMWelDX/weldx/blob/529ced9040111948e28554a0d6f54bc17b368d0b/setup.cfg#L87-L96

marscher: OK, and nbval picks up the settings from pytest-cov. So I can omit it in the pytest invocation.

CagtayFabry:
I tried hard not to have a `pyproject.toml` file in addition to `setup.cfg` but that's fine if this makes it simpler somehow 😉

.github/workflows/pytest.yml
```diff
@@ -34,19 +36,22 @@ jobs:
...
           key: ${{ runner.os }}-pip-${{ hashFiles('setup.cfg') }}
           restore-keys: |
             ${{ runner.os }}-pip-
 
       - name: pip installs
+        if: steps.cache_${{ runner.os }}.outputs.cache-hit != 'true'
         run: |
-          pip install pytest pytest-cov pytest-xdist setuptools_scm
-          pip install -e .
+          pip install -e .[test]
```
CagtayFabry, line 49: How to I interpret this? we don't perform an install when the pip cache gets loaded? why?

.github/workflows/pytest.yml
```diff
@@ -34,19 +36,22 @@ jobs:
...
       - name: run pytest
         run: |
-          pytest --runslow -n2
+          pytest -n2 --dist loadscope --nbval --current-env tutorials/
+          pytest --runslow -n2 --cov-append tests/
```
CagtayFabry, line 53: can we split this into two steps?
First one running the default tests and the second one running the notebooks? Should make it easier to spot where failures happen

I would also be fine with only running the notebooks if the normal tests succeed

marscher: I split these up as demanded. I had to change the way the coverage is combined though

.github/workflows/pytest.yml
marscher: you're right, it doesn't make any sense. I saw this pattern on Stack Overflow, where somebody complained that pip is still downloading things where it should use the cache.

marscher:
using a pyproject.toml file to declare the build system (e.g. setuptools_scm gets installed before accessing the version number), somehow the picked up package structure is faulty, as it only contains the very top level, but not split up modules like vis and trans.

marscher:
I think this should be ready to go. The coverage actually decreases now, because I removed some "pragma: no cover" for tools, which actually can now be tested. I can also try to invoke them in the tutorials at a suitable spot.

marscher:
> I tried hard not to have a `pyproject.toml` file in addition to `setup.cfg` but that's fine if this makes it simpler somehow wink

May I know why did you want to avoid it?

CagtayFabry:
> 
> 
> > I tried hard not to have a `pyproject.toml` file in addition to `setup.cfg` but that's fine if this makes it simpler somehow wink
> 
> May I know why did you want to avoid it?

Just trying to keep everything in one place as much as possible to avoid duplications

CagtayFabry:
my final suggestion would be to only run the notebooks and coverage report/action on a single os (and python version) to save a little time on the slower runs, like so:
```yaml
        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
        run: |
          pytest -n2 --dist loadscope --nbval --current-env --cov=weldx --cov-append tutorials/
          echo "Exited with '$?'"

      - name: coverage xml (combination is performed by pytest-cov)
        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
        run: coverage xml

      - name: codecov.io
        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
        uses: codecov/codecov-action@v1
```

what do you think @vhirtham @marscher ?

CagtayFabry:
I also set the default mutiprocessing option to `-n auto`, I think this is reasonable for local runs as well, what do you think @marscher ?

09dd241d7546fd673cf1b6e2769df3e7b6931655

CagtayFabry (approved):
LGTM

I vote for running the full test suite including notebooks and coverage report only on the relevant `ubuntu-latest` job since we don't have branching code for different OSes

marscher:
I agree, we should run this only the "latest" configuration to save resources.

vhirtham:
Scared about what happens to the coverage afterward, but we can probably remove `- "weldx/visualization/*"` from the `codecov.yml` too

CagtayFabry:
> Scared about what happens to the coverage afterward, but we can probably remove `- "weldx/visualization/*"` from the `codecov.yml` too 😨

only one way to find out .. f5c582c58cddbc71d6cedf1328f84156dc14dcc3 😉

marscher:
What is your concern here? Do you consider "coverage" only unit tested code and therefore an integration test like the notebooks are not in this scope?

vhirtham:
> What is your concern here? Do you consider "coverage" only unit tested code and therefore an integration test like the notebooks are not in this scope?

Think you missed the irony. 😉 I have no concerns. I just wanted to point out, that the visualization package is (was) still ignored due to the `codecov.yml` and that the coverage will further plummet after removing the corresponding lines 😛

vhirtham (approved):

.github/workflows/pytest.yml
```diff
@@ -41,14 +41,23 @@ jobs:
...
+        run: pip install -e .[test]
+
       - name: run pytest
         run: |
-          pytest --runslow -n2
+          pytest -n 2 --runslow tests/
+          echo "Exited with '$?'"
+
+      - name: run pytest on notebooks (append coverage)
+        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
```
marscher, line 52: I just wonder if one could write the following in GH-actions:
```

if: (matrix.os == ...)
   - name: run pytest on notebooks....
   - run: ...
   - name: ... 
```

CagtayFabry: pretty sure that's not possible because it would conditionally change the workflow setup (would be nice though)

Issue #296, opened 2021-03-23
update gas component schema


CagtayFabry:
the gas component schema should have a `wx_unit` validator and use `tag` instead of `$ref`

https://weldx.readthedocs.io/en/latest/generated/weldx.bam.de/weldx/aws/process/gas_component-1.0.0.html

we can also try applying the default min/max validators but not sure if they work with quantities

Issue #207, opened 2020-11-29
add shape and unit validation rotation-1.0.0 schema


CagtayFabry:
just noticed the `core/transformations/rotation-1.0.0` schema is missing `wx_shape` and `wx_unit` validators

units will need #153

PR #303, opened 2021-03-25
replace $ref when tags are used

### LINKED ISSUE #302: replace $ref with tag in schemas
I think there are still some schemas that use `$ref` where `tag` would be sufficient.

This should speed up ASDF validation

CagtayFabry:
## Changes
I replaced some old `$ref` uses in schemas with `tag`, it should be more consistent now
if we're lucky this could even speed up some things

also add `wx_unit` to gas_component schema

## Related Issues

Closes #302, #296, #207

## Checks

- [x] ~updated CHANGELOG.md~
- [ ] ~updated tests~
- [ ] ~updated doc/~
- [ ] ~update example/tutorial notebooks~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - `weldx.util.deprecated` decorator [[#295]](https://github.com/BAMWelDX/weldx/pull/295)
 
+### ASDF
+
+- add validators to `rotation-1.0.0.yaml`
+  & `gas_component-1.0.0.yaml` [[#303]](https://github.com/BAMWelDX/weldx/pull/303)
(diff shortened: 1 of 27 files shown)


vhirtham (approved):

marscher:
I'm curious if this change speed up the asdf test suite :)

CagtayFabry:
> 
> 
> I'm curious if this change speed up the asdf test suite :)

I did a quick test with 2 runs on master and this branch `pytest ./tests/asdf_tests -n1`
master: 51.5s / 49.6s
branch: 48.6s / 49.1s

so not really

PR #314, opened 2021-03-30
move include_package_data back to options

CagtayFabry:
## Changes
fix `setup.cfg` to include asdf schemas


marscher:
So this setting somehow (my whoopsie?) got into the wrong section?

marscher:
I just wonder why there is an discrepancy between pip with and without conda, if this setting was broken.

marscher (approved):

PR #319, opened 2021-03-30
add pytest run to conda build

CagtayFabry:
## Changes
Add default pytest run to the conda build step

also updated the local recipe to match the conda-force recipe more closely
we should add test run to the conda-forge recipe as well once the setup is refined


marscher:
Actually you want to copy the test directory during the test phase of conda build. Have a look here https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-section

conda.recipe/meta.yaml
```diff
@@ -1,34 +1,74 @@
...
 
 test:
-    imports:
-        - weldx
-        - weldx.visualization
+  requires:
+    - pytest
+    - pytest-xdist
+  source_files:
+    - tests/*.py
```
marscher, line 32: Does passing "tests/" here not work?

Why do you need to pass the icons separately? How are they used in the tests?

marscher:
because of #320 you could directly use the test command "pytest --pyargs tests" but I guess we do not really want that.

conda.recipe/meta.yaml
CagtayFabry: they get serialized as external files in a part of the asdf test suite

https://github.com/BAMWelDX/weldx/blob/bd643850b670a9da087e7739406bc7074a0dcc84/tests/asdf_tests/test_asdf_core.py#L416-L430

vhirtham (approved):

conda.recipe/meta.yaml
```diff
@@ -1,34 +1,74 @@
...
+    comparability and reproducibility across different experimental setups can
+    be improved.
+  doc_url: https://weldx.readthedocs.io
+  dev_url: https://github.com/BAMWelDX/weldx/
+
+extra:
+  recipe-maintainers:
+    - marscher
+    - vhirtham
+    - CagtayFabry
```
vhirtham, line 73: Did you put your name on purpose to the bottom of the list? 😛

marscher (approved):
LGTM

conda.recipe/meta.yaml
CagtayFabry: actually that was @marscher 😉

marscher: Well it is not an authorship list of a nature paper. I admit I haven't spent any thought on the order.

conda.recipe/meta.yaml
```diff
@@ -1,34 +1,69 @@
...
-        - weldx
-        - weldx.visualization
+  requires:
+    - pytest
+    - pytest-xdist
+  imports:
+    - weldx
+    - weldx.visualization
+  commands:
+    - pytest --pyargs weldx.tests -n auto
```
CagtayFabry, line 35: this looks much cleaner now after #323 👍

marscher: yes, indeed - very simple :)
And we are testing the packaged code, not the pointer to the working copy, where everything is always in the right spot. This is another benefit.

marscher (approved):

conda.recipe/meta.yaml
vhirtham: I suspected that Cagtay doesn't want to be bothered with maintenance questions 🤣

PR #323, opened 2021-03-31
Move tests folder

CagtayFabry:
## Changes
- move tests directory to `weldx` module
- add static data used for tests to `weldx/tests/data`
- change pytest and coverage config

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/


.github/workflows/pytest.yml
```diff
@@ -52,13 +52,13 @@ jobs:
...
         run: |
-          pytest -n 2 --runslow tests/
+          pytest -n 2 --runslow
           echo "Exited with '$?'"
 
       - name: run pytest on notebooks (append coverage)
         if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
         run: |
-          pytest -n 2 --dist loadscope --nbval --current-env --cov=weldx --cov-append tutorials/
+          pytest -n 2 --dist loadscope --nbval --current-env --cov-append tutorials/
```
CagtayFabry, line 61: `--cov=weldx`should be set in `setup.cfg`

CagtayFabry:
since the test files will get package for pypi/conda I have swapped out the `WelDX_notext.ico` file for the much smaller `WelDX_notext.svg` and included it in the test folder
This is no longer a binary file but I think this shouldn't matter for the tests? @vhirtham

.github/workflows/pytest.yml
marscher: I know, but the way nbval interacts with pytest-cov, it seemed mandatory to explicitly give the argument again. It shouldn't make a difference and I consider it a bug in nbval. You could try omitting it and see what happens (there will be no nb coverage).

marscher: But eventually the cov-append flag also triggers coverage. We will see

marscher: So the cov report is unchanged, which means I added the append flag later on during my last PR. Works like expected.

marscher:
why did you move conftest outside of tests?

CagtayFabry:
> why did you move conftest outside of tests?

I did this so you can run `pytest --runslow` from the base directory and it will result in the same config

before you would have to specify the test directory so that the config would get picked up and the argument recognized py pytest. Leaving `conftest.py` in `weldx/tests` you need to run `pytest --runslow weldx/tests`

See here leaving conftest.py in old folder: https://github.com/BAMWelDX/weldx/runs/2236208545#step:9:1

Basically I want to just type `pytest` from root with all settings preconfigured

marscher:
Ah I see. I guess pytest does not support hidden files, right?

CagtayFabry:
maybe you con specify a custom conftest.py file path but I think this should be ok for now :)

vhirtham:
> since the test files will get package for pypi/conda I have swapped out the `WelDX_notext.ico` file for the much smaller `WelDX_notext.svg` and included it in the test folder
> This is no longer a binary file but I think this shouldn't matter for the tests? @vhirtham

Should be fine

vhirtham (approved):

marscher (approved):
LGTM

PR #326, opened 2021-04-01
63 mc graph backend

### LINKED ISSUE #63: graph backend for measurement chain
implement signal -> transformation -> signal relation (like in measurement chains) as graph (NetworkX)

vhirtham:
## Changes

Rewrite the `MeasurementChain` based on a `networkx.DiGraph` 

## Related Issues

Closes #63

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [x] Reminder: Review tests and remove todo's


weldx/measurement.py
```diff
@@ -55,6 +70,332 @@ class Source:
...
+        self.add_transformation(
+            transformation.name,
+            transformation.error,
+            transformation.output_signal.signal_type,
+            transformation.output_signal.unit,
+            transformation.func,
+            input_signal_source,
+        )
+
+    def get_data(self, name: str) -> xr.DataArray:
```
CagtayFabry, line 313: I think it would be ok if data itself does not have a name but is simply attached to a signal, so either a signal has data or not and just access it via the signal. What do you think?

CagtayFabry:
this is much simpler to use and cleaner than the old class I think 😃 👍 
maybe add the plot function next so we can play around a little more?

weldx/measurement.py
vhirtham: What if a signal has multiple data blocks? For example, expected and measured? Also, the signal also has no name, so we have to use the name of the signal source.

vhirtham:
> this is much simpler to use and cleaner than the old class I think 😃 👍
> maybe add the plot function next so we can play around a little more?

That's what I am currently doing :p

weldx/measurement.py
CagtayFabry: I don't think that fits the current scope of the measurement chain which should describe exactly one measurement.
Not sure how e.g. an expected value would fit into that, maybe that should it described in it's own measurement chain? Or just an external signal?

vhirtham: We can remove the name. I'll change the interfaces accordingly

vhirtham:
While updating the measurement tutorial I came across something we might need to consider/solve differently now:

We had some shared data between the measurement chain and for example the `GernericEquipment` class. Namely the transformation and source class. The latter one is currently completely removed from the measurement chain and the first one is only constructed if desired. It is not used as an internal storage format. The side effect is, that an asdf serialization would store the data twice. We can either use those classes internally, which will probably make the code a bit more complicated/cluttered or we can also rethink how the data is shared between equipment classes and a measurement chain. 

I tend towards rethinking the structure to keep everything simple (code and user experience). For example, the equipment class can be handed over to the measurement chain while creating it (source) or adding a transformation to establish a link. Asdf would then also create a link to the equipment. Additionally, the equipment gets the name of the measurement chain it was added to attached. It which can be used to fetch the chain from the measurement class. We need the name to break the circular dependency here (same problem as in the graph serialization). Maybe we can also directly assign the measurement chain to the equipment class (circular dependency) on the python side for enhanced ease of use, but for the serialization, I think we can't solve it without name referencing on one of both sides.

vhirtham:
I have now implemented a basic coupling between the equipment and the measurement chain. It is still WIP, but I think the direction is somewhat obvious (have a look at the newly added notebook). Still need to solve how we can avoid data duplications in the asdf file, but I have at least some ideas. Let me know what you think @CagtayFabry @marscher

CagtayFabry:
> 
> 
> I have now implemented a basic coupling between the equipment and the measurement chain. It is still WIP, but I think the direction is somewhat obvious (have a look at the newly added notebook). Still need to solve how we can avoid data duplications in the asdf file, but I have at least some ideas. Let me know what you think @CagtayFabry @marscher

Thanks I will take a look into it now
Pretty sure ASDF can handle circular dependencies but I have not found the correct hooks.

We should definitly come up with a good way to handle these bidirectional links at some point

weldx/measurement.py
```diff
@@ -47,55 +55,459 @@ class DataTransformation:
...
 
 
+# equipment ----------------------------------------------------------------------------
 @dataclass
+class GenericEquipment:
+    """Simple dataclass implementation for generic equipment."""
+
+    name: str
+    sources: List = field(default_factory=lambda: [])
+    data_transformations: List = field(default_factory=lambda: [])
```
CagtayFabry, line 84: maybe just `transformations` or `signal_transformations` is more fitting here?

tutorials/measurement_example.ipynb
```diff
@@ -146,7 +146,7 @@
...
     "Now we define the missing elements of our measurement chain and bundle everything together. A core concept of the chain are `signals` that go in and out of `transformations` which define mathematical operations of the signals, forming the chain.\n",
     "\n",
-    "Each measurement chain starts with a `Source` signal. This is the point where our physical process or quantity is initially detected by a sensor. We define a `measurement.Source` object by giving a `name`, defining the `Signal` type and an `error` representing the uncertainty attached to the signal source. In a way, a `Source` is a special type of `transformation` in that it only has an `output_signal` but no `input_signal`.\n",
+    "Each measurement chain starts with a source signal. This is the point where our physical process or quantity is initially detected by a sensor. Since every `MeasurementChain` needs at least a source, we need to provide the corresponding data when creating an instance of this class. The parameters are the `source_name`, the `output_signal_type` and `output_signal_unit` as well as the `source_error` representing the uncertainty attached to the signal source. \n",
```
CagtayFabry, line 149:
```suggestion
    "Each measurement chain starts with a source signal. This is the point where our physical process or quantity is initially detected by a sensor. Since every `MeasurementChain` needs at least a source, we need to provide the corresponding information when creating an instance of this class. The parameters are the `source_name`, the `output_signal_type` and `output_signal_unit` as well as the `source_error` representing the uncertainty attached to the signal source. \n",
```
or maybe `metadata` here

weldx/measurement.py
```diff
@@ -47,55 +55,459 @@ class DataTransformation:
...
+        self,
+        name: str,
+        error: Error,
+        output_signal_type: str = None,
+        output_signal_unit: str = None,
+        output_signal: Signal = None,
+        func: "MathematicalExpression" = None,
+        input_signal_source: str = None,
+        data=None,
+    ):
```
CagtayFabry, line 311: So currentyl this does not create a `SignalTransformation` instance?
How about the orther way around? (attaching `SignalTransformation` to the edge?)

weldx/measurement.py
```diff
@@ -47,55 +55,459 @@ class DataTransformation:
...
+            node_id=source_name,
+            signal_type=output_signal_type,
+            unit=output_signal_unit,
+            signal=output_signal,
+        )
+        if signal_data is not None:
+            self.add_signal_data(signal_data)  # todo : test
+
+    @staticmethod
+    def construct_from_source(name, source: SignalSource):
```
CagtayFabry, line 165: I think these should be classmethods and simply named `from_source` etc. for consistency

weldx/measurement.py
vhirtham: Just didn't bother with renaming stuff so far. But renaming it to transformations is already on my list.

weldx/measurement.py
vhirtham: I did something similar for the Signal. However, what bothers me is that attaching a `Signaltransformation` to an edge will result in storing duplicate information and therefore cluttering the asdf file. Maybe we can discuss this tomorrow. A possible solution might be to encapsulate the non-duplicate stuff in another class, but I am still struggling to see a concept that satisfies my requirements

CagtayFabry: why would it duplicate information?
If the transformation already exists can we not just link it? (for example just passing through an existing `SignalTransformation` as the first argument)

vhirtham: Because the signals are stored as nodes in the graph and inside the transformation itself. They are not necessarily the same instance, so no refs in asdf.

vhirtham: So if we want to store the transformation class directly, the same should probably apply to the source. Maybe we should discuss this before I proceed.

CagtayFabry: > 
> 
> Because the signals are stored as nodes in the graph and inside the transformation itself. They are not necessarily the same instance, so no refs in asdf.

Oh I see is it because the `SignalTransformations` have in/output signals as attributes?
```python
@dataclass
class SignalTransformation:
    name: str
    error: Error
    input_signal: Signal
    output_signal: Signal
    func: "MathematicalExpression" = None
    input_shape: Tuple = None
    output_shape: Tuple = None
```
would that not be covered by the graph now?

vhirtham: It would, but only inside the graph. A transformation at a piece of equipment still needs these information. I already have an idea how to solve this. Would require some restructuring and I would really have your opinion on that @marscher @CagtayFabry before I start it. It basically affects the internal structure of the graph at the python side, not asfd: 

If we would leave the nodes blank (except for the name) and store the signal data at the transformations and the source, we have just a single duplication of each signal. They can still be different objects at the start, but we can introduce a synchronization method for that. Since signals are always encapsulated inside of an `Source` or `SignalTransformation` object, overwriting them will change the state of the encapsulating class. So lets say we have two equipments that are paired in a measurement chain. Their signals are identical but they are not same python object. When adding both to a measurement chain, we just need to copy the signal of one class into the other one to enforce an identical python object before we copy the transformation itself to the measurement chain. Does this make sense? What do you think?

weldx/measurement.py
```diff
@@ -53,55 +60,507 @@ class DataTransformation:
...
+
+    def _raise_if_node_exist(self, node_id: str):
+        """Raise en error if the graph already contains a node with the passed id.
+
+        Parameters
+        ----------
+        node_id :
+            Name that should be checked
+
+        """
```
CagtayFabry, line 266:
```suggestion
        """Raise en error if the graph already contains a node with the passed id."""
```
A short description instead of a full docstring feels more readable for internal functions like this that are very descriptive with type-hints

weldx/measurement.py
```diff
@@ -37,71 +28,684 @@ class Signal:
...
+            self._raise_if_node_does_not_exist(node_name)
+        return node_name
+
+    @staticmethod
+    def _letter_to_signal_type(letter):
+        """Convert a single letter to the corresponding signal type."""
+        if letter == "A":
+            return "analog"
+        elif letter == "D":
+            return "digital"
```
CagtayFabry, line 400:
```suggestion
        return dict(A="analog", D="digital")[letter.upper()]
```

weldx/tests/test_measurement.py
```diff
@@ -1,112 +1,369 @@
...
+        kwargs :
+            A dictionary containing some keyword arguments that should replace the
+            default ones.
+
+        Returns
+        -------
+        Dict :
+            Dictionary with keyword arguments for the `add_transformation` method
+
+        """
```
CagtayFabry, line 123:
```suggestion
    def _default_add_transformation_kwargs(cls, replace_kwargs: Dict = None) -> Dict:
        """Return an updated dictionary of default keyword arguments required by `add_transformation`."""
```
this makes it much easier to scroll through the code :)

weldx/measurement.py
```diff
@@ -207,25 +809,32 @@ def _signal_label(signal):
...
         draw_networkx_edge_labels(graph, positions, edge_labels, ax=axes)
 
         return axes
 
+    def to_tree(self) -> Dict:
+        return dict(
+            name=self._name,
+            data_source=self._source,
+            source_signal=build_tree(self._graph, self._source.name),
+        )
```
CagtayFabry, line 828: I'm wondering if we should use the default `nx.DiGraph` serialization here instead of relying on `build_tree`
it would look a little different (more nested) in the schema but maybe more stable and simplier to maintain in the long run

weldx/measurement.py
```diff
@@ -37,71 +28,684 @@ class Signal:
...
+        else:
+            raise ValueError("The equipment does not have a transformation.")
+
+        input_signal_source = self._check_and_get_node_name(input_signal_source)
+        self.add_transformation(transformation, data, input_signal_source)
+        self._graph.edges[(input_signal_source, self._prev_added_signal)][
+            "equipment"
+        ] = equipment
+
+    def add_signal_data(self, data: "TimeSeries", signal_source: str = None):
```
CagtayFabry, line 561: From going through the code I am not sure if `Signal.data` is supposed to be a `TimeSeries` or `xr.DataArray` (or either?)

weldx/measurement.py
vhirtham: I tried this first but then asdf expects a di_graph as root-node. Would need to investigate how the schema has to be adjusted.

vhirtham:
The tutorial is finished and all features are now implemented if I haven't missed something. Tomorrow, I'll clean up everything, fix the measurement notebook and the tests. I will also check if I have covered everything we discussed here. Then we should be good to go.

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
```diff
@@ -6,19 +6,66 @@ tag: "tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0"
 
 title: |
   Schema that describes a measurement_chain
+description: |
+  A measurement chain describes the different steps that are performed to acquire experimental data. It collects all
+  relevant information about the initial source signal and the involved transformations. In this schema, the fundamental
+  data representation of the measurement chain is a directed graph with the signals being the nodes and the
+  transformations being the edges.
+
+#example: Add when schema is accepted
```
vhirtham, line 15: Reminder

vhirtham:
@marscher @CagtayFabry I am not sure, why the tests for `compare_nested` are currently failing. Seems like there is a problem with a container comparison

~~~
[gw1] linux -- Python 3.8.9 /opt/hostedtoolcache/Python/3.8.9/x64/bin/python
weldx/tests/test_utility.py:621: in test_metadata_modified
    assert not ut.compare_nested(self.a, self.b)
weldx/util.py:1345: in compare_nested
    iterutils.remap(a, visit=visit, reraise_visit=True)
/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/boltons/iterutils.py:1082: in remap
    visited_item = visit(path, key, value)
weldx/util.py:1310: in _visit
    if not _Eq_compare_nested._compare(value, other_value):
weldx/util.py:1277: in _compare
    return x == y
<string>:4: in __eq__
    ???
/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/xarray/core/common.py:130: in __bool__
    return bool(self.values)
E   ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
~~~

Couldn't really get to the root of the problem with the debugger. I don't know if it is a bug in `compare_nested` or if I messed something up. Maybe you have a clue. Note that the "measurement data" is currently a `DataArray` and should be a `TimeSeries`. I will change this and maybe this fixes the tests. But in any case, we should find out why it is currently failing.

CagtayFabry:
> @marscher @CagtayFabry I am not sure, why the tests for `compare_nested` are currently failing. Seems like there is a problem with a container comparison
> 
> ```
> [gw1] linux -- Python 3.8.9 /opt/hostedtoolcache/Python/3.8.9/x64/bin/python
> weldx/tests/test_utility.py:621: in test_metadata_modified
>     assert not ut.compare_nested(self.a, self.b)
> weldx/util.py:1345: in compare_nested
>     iterutils.remap(a, visit=visit, reraise_visit=True)
> /opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/boltons/iterutils.py:1082: in remap
>     visited_item = visit(path, key, value)
> weldx/util.py:1310: in _visit
>     if not _Eq_compare_nested._compare(value, other_value):
> weldx/util.py:1277: in _compare
>     return x == y
> <string>:4: in __eq__
>     ???
> /opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/xarray/core/common.py:130: in __bool__
>     return bool(self.values)
> E   ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> ```
> 
> Couldn't really get to the root of the problem with the debugger. I don't know if it is a bug in `compare_nested` or if I messed something up. Maybe you have a clue. Note that the "measurement data" is currently a `DataArray` and should be a `TimeSeries`. I will change this and maybe this fixes the tests. But in any case, we should find out why it is currently failing.

It looks like some `array-like` object get's compared with `==` where something like `np.allclose()` should be used
My guess: Is there `@dataclass` where one attribute is a `DataArray` that does not define a custom `__eq__`? `Signal` ? In that case the default `__eq__` created by `@dataclass` would probably fail, simply calling `__eq__` on all attributes (which doesn't work for `DataArray`)

vhirtham:
@CagtayFabry 
Found this out too. Here is my minimal example:

~~~ python
from copy import deepcopy
from dataclasses import dataclass

from xarray import DataArray

from weldx.util import compare_nested


class SomeClass:
    def __init__(self, member):
        self.member = member


@dataclass
class SomeDataClass:
    member: DataArray = None


data_array = DataArray(data=[1, 2, 3], dims=["time"])

original = SomeDataClass(member=data_array)
# original = SomeClass(member=DataArray(data=[1, 2, 3], dims=["time"]))
# original = DataArray(data=[1, 2, 3], dims=["time"])
# original = SomeDataClass([1, 2, 3])

other = deepcopy(original)
print(original == other)
print(compare_nested([original], [other]))
~~~

What bothers me, is that python is extremely inconsistent here. If you use a normal class the comparison (and hence compare_nested) will simply return False, which is somehow even worse. This makes it really hard to debug if you do not know the exact problem. 
Now for `compare_nested` the situation is as follows:
- I can use a `DataArray` inside of a Dict, List etc. -> works
- If I use it inside of a normal class (somewhere deeply hidden in a big nested structure) it simply returns `False` -> Have fun finding this one
- If I use it inside of a data class, the function raises an Exception. -> At least I get a hint via debugger, which class causes the problem

I don't know what your opinion is, but I find this behavior extremely confusing with a great potential for frustration. I know this is more a problem that arises from the `DataClass` being default comparable, but I would highly appreciate it, if we can find a good solution here that makes our function behave consistently.

CagtayFabry:
maybe we could "enhance" the `enter` function in a way that it also "enters" dataclasses and calls the custom define comparison functions on each attribute? But I imagine this could lead to some ugly dataclass code...
https://github.com/BAMWelDX/weldx/blob/ab44ef00182cd6c52dd976bc5541f9c8a4ca9121/weldx/util.py#L1237-L1240

vhirtham:
@marscher appeciate your efforts, but there is no need to fix the current state since the data will be replacef by a TimeSeries, which should eliminate the problem anyways. It's more a general concern how we should handle this in the future.

vhirtham:
@CagtayFabry Maybe could take a short look together at the measurement example tutorial. The structure regarding the measurement chain doesn't really fit anymore with the new implementation but I wanted to discuss that before rewriting it.

scripts/welding_schema.py
```diff
@@ -103,119 +103,118 @@ def single_pass_weld_example(out_file="single_pass_weld_example.asdf"):
...
+            "a * x + b", dict(a=Q_(100.0 / 32768.0, "V"), b=Q_(0.0, "V"))
+        ),
     )
+
+    # Define lab equipment
+
+    HKS_sensor = msm.GenericEquipment(
+        name="HKS P1000-S3",
+        sources=[src_current, src_voltage],
+        transformations=[voltage_AD_transform],  # Is this really done by this eq.?
```
vhirtham, line 169: Is the voltage AD conversion done by the HKS sensor?

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
```diff
@@ -6,19 +6,101 @@ tag: "tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0"
...
   Schema that describes a measurement_chain
+description: |
+  A measurement chain describes the different steps that are performed to acquire experimental data. It collects all
+  relevant information about the initial source signal and the involved transformations. In this schema, the fundamental
+  data representation of the measurement chain is a directed graph with the signals being the nodes and the
+  transformations being the edges.
+
+example:
+  -
+    - A measurement chain consisting with a single transformation that performs an AD conversion.
```
vhirtham, line 17:
```suggestion
    - A measurement chain containing a single transformation that performs an AD conversion.
```

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
```diff
@@ -6,19 +6,101 @@ tag: "tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0"
...
   data_source:
     tag: "tag:weldx.bam.de:weldx/measurement/source-1.0.0"
-  data_processors:
-    type: array
-    items:
-      tag: "tag:weldx.bam.de:weldx/measurement/data_transformation-1.0.0"
+  source_equipment:
+    tag: "tag:weldx.bam.de:weldx/equipment/generic_equipment-1.0.0"
+  source_signal:
+     $ref: "#/definitions/signal_node"
```
vhirtham, line 100: Descriptions missing (Also in the defines)

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
+
+        self._name = name
+        self._source = source
+        self._source_equipment = None
+        self._prev_added_signal = None
+        self._graph = DiGraph()
+
+        self._add_signal(node_id=source.name, signal=source.output_signal)
+        if signal_data is not None:
+            self.add_signal_data(signal_data)  # todo : test
```
vhirtham, line 202:
```suggestion
            self.add_signal_data(signal_data)
```

weldx/tests/test_measurement.py
```diff
@@ -1,112 +1,480 @@
...
+            A dictionary with keyword arguments that are used to construct the
+            `SignalTransformation` that is passed to the `add_transformation` method.
+            Missing arguments are added.
+
+        """
+        mc = MeasurementChain(**self._default_init_kwargs())
+
+        mc.add_transformation(self._default_transformation(tf_kwargs))
+
+        # todo: add assertions (check returned signal)
```
vhirtham, line 232: reminder

weldx/util.py
```diff
@@ -421,6 +421,24 @@ def pandas_time_delta_to_quantity(
...
+        First unit
+    unit_2 :
+        Second unit
+
+    Returns
+    -------
+        `True` if both units have the same dimensionality and `False` otherwise.
+
+    """
+    return Q_(1, unit_1).dimensionality == Q_(1, unit_2).dimensionality
```
vhirtham, line 439: If you know a way to do this without creating a dummy quantity, feel free to make a suggestion

vhirtham:
Apart from the necessary update of the measurement tutorial (as mentioned earlier), this should be ready

scripts/welding_schema.py
CagtayFabry: no the HKS sensor outputs 0V - 10V (`"AA"`), `"AD"` happens at `ELM_3104`

vhirtham: As I thought.

CagtayFabry:
I took a quick look over the new tutorial 👍 

a few notes that would make it more accessible imo:
- plot the new measurement chain layout after each change
- add an introduction explaining the concept (without code):
  - what is a signal?
  - what is a transformation?
- leave out analog/digital conversion from introduction as it is too complicated to get started imo
- if possible I would prefer more code segments between the text and shorter text-segments (if it is supposed to focus on code)

also I would like to hear your thoughts @AndreasPittner @marscher **@rwinterk** (fake pint, need to get him on GH 😄 )

weldx/util.py
marscher: There is UnitRegistry.get_dimensionality(input_units) returning a dict base dimensions. This can be checked for quality, but I guess for just two units creating the actual Unit is an easy and readable way.

scripts/welding_schema.py
```diff
@@ -265,8 +263,12 @@ def single_pass_weld_example(out_file="single_pass_weld_example.asdf"):
         reference_timestamp=reference_timestamp,
         equipment=[HKS_sensor, BH_ELM],
         measurements=[welding_current, welding_voltage],
-        welding_current=current_calib_transform.output_signal,
-        welding_voltage=voltage_calib_transform.output_signal,
+        welding_current=welding_current_chain.get_signal(
+            "Calibration current measurement"
```
marscher, line 267: could we also write something like get_signal(-1) to get the last entry of the chain? I think getting by name is nice, but seems a bit unhandy from a programmers point of view.

vhirtham: would rather make it return the last signal of the chain if no name is provided like it is the case for some of the other functions. Since the underlying structure is not a list, a direct access operator doesn't make that much sense, does it? Never tried `dict[-1]`. Is that possible in python?

CagtayFabry: I think it would be best and most intuitive to implement `.signals` and `.transformations` that expose a list of signals/tf or (maybe even better) provide an iterator?

That should enable `.signals[-1]`, `.signals[0:3]` etc

weldx/util.py
CagtayFabry: this seems to be implemented with
```python
from weldx.constants import WELDX_UNIT_REGISTRY as ureg
ureg.is_compatible_with("m*s","inch*min")
```

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
```diff
@@ -6,19 +6,119 @@ tag: "tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0"
...
+      !<tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0>
+        name: Measurement Chain
+        data_source: !<tag:weldx.bam.de:weldx/measurement/source-1.0.0>
+          name: Source
+          output_signal: &id001 !<tag:weldx.bam.de:weldx/measurement/signal-1.0.0>
+            signal_type: analog
+            unit: V
+          error: !<tag:weldx.bam.de:weldx/measurement/error-1.0.0>
+            deviation: !unit/quantity-1.1.0 {unit: percent, value: 0.2}
+        source_signal: !<tag:weldx.bam.de:weldx/core/graph/di_node-1.0.0>
```
CagtayFabry, line 28: for clarity maybe simply use `source` or `source_node` here since technically this isn't a signal?

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
```diff
@@ -6,19 +6,119 @@ tag: "tag:weldx.bam.de:weldx/measurement/measurement_chain-1.0.0"
...
+          name: Source
```
CagtayFabry, line 29: I think it would be helpful to provide a simplified real world example like `welding current` instead of a more generic/abstract one since the schema is rather complex

weldx/asdf/tags/weldx/measurement/signal_transformation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    name = "measurement/signal_transformation"
+    version = "1.0.0"
+    types = [SignalTransformation]
+    requires = ["weldx"]
+    handle_dynamic_subclasses = True
+
+    @classmethod
+    def to_tree(cls, node: SignalTransformation, ctx):
+        """convert to tagged tree and remove all None entries from node dictionary"""
```
CagtayFabry, line 18:
```suggestion
```
This is outdated anyway so we can remove those docstrings in other functions as well

if we haven't already I'm fine with deactivating pydocstyle for all `to_tree` and `from_tree` methods since it's pretty clear what they do

weldx/measurement.py
```diff
@@ -1,34 +1,25 @@
 """Contains measurement related classes and functions."""
 
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Dict, List, Tuple, Union  # noqa: F401
+from typing import TYPE_CHECKING, Dict, List, Tuple  # noqa: F401
+from warnings import warn
 
-import xarray as xr
+from weldx.asdf.tags.weldx.core.graph import build_graph, build_tree
```
CagtayFabry, line 7: so is it not feasible to use the default graph serialization?

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
     name: str
     output_signal: Signal
     error: Error
 
 
+# equipment ----------------------------------------------------------------------------
 @dataclass
-class MeasurementChain:
-    """Simple dataclass implementation for measurement chains."""
+class GenericEquipment:
```
CagtayFabry, line 84: Maybe we should rename this to `MeasurementEquipment` since it is rather specific to this application ?
My initial idea behind `GenericEquipment` was to provide some generic metadata that can be used for all kinds of equipment (robots, power sources, sensors etc.)

So other equipment-classes like `MeasurementEquipment` could inherit from that class

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
+        ----------
+        other :
+            The measurement chain that should be compared.
+
+        Returns
+        -------
+        bool :
+            `True` if 2 measurement chains are equal and `False` otherwise.
+
+        """
```
CagtayFabry, line 217:
```suggestion
        """Return `True` if 2 measurement chains are equal and `False` otherwise."""
```

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
+        return node_name
+
+    @classmethod
+    def _determine_output_signal(
+        cls,
+        transformation: SignalTransformation,
+        input_signal: Signal,
+        data: "TimeSeries",
+    ) -> Signal:
+        """Get the signal that is produced by the provided transformation.
```
CagtayFabry, line 364:
```suggestion
        """Create a signal that is produced by the provided transformation.
```
The Signal doesn't exist in the graph/mc yet if I understand correctly?

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
+        input_unit :
+            The unit of the input signal
+
+        Returns
+        -------
+        str:
+            Unit of the transformations' output signal
+
+        """
+        from weldx import Q_
```
CagtayFabry, line 441: any reason this can not be a top level import?

weldx/measurement.py
```diff
@@ -37,71 +28,705 @@ class Signal:
...
+            raise KeyError(
+                f"The internal graph already contains a node with the id {node_id}"
+            )
+
+    def _raise_if_node_does_not_exist(self, node: str):
+        """Raise a `KeyError` if the specified node does not exist."""
+        if node not in self._graph.nodes:
+            raise KeyError(f"No signal with source '{node}' found")
+
+    def add_signal_data(self, data: "TimeSeries", signal_source: str = None):
```
CagtayFabry, line 480:
```suggestion
    def add_signal_data(self, data: TimeSeries, signal_source: str = None):
```
I think this should work with `from __future__ import annotations`

weldx/measurement.py
```diff
@@ -119,69 +744,63 @@ def plot(self, axes=None):
 
         """
         import matplotlib.pyplot as plt
-        from networkx import DiGraph, draw, draw_networkx_edge_labels
-
-        def _signal_label(signal):
-            return f"{signal.signal_type}\n[{signal.unit}]"
+        from networkx import draw, draw_networkx_edge_labels
```
CagtayFabry, line 747: `networkx` can be a top level import since it gets loaded by `DiGraph`

weldx/measurement.py
```diff
@@ -207,25 +830,76 @@ def _signal_label(signal):
...
+    @property
+    def source(self) -> SignalSource:
+        """Return the source of the measurement chain.
+
+        Returns
+        -------
+        SignalSource :
+            The source of the measurement chain
+
+        """
```
CagtayFabry, line 869:
```suggestion
        """Return the source of the measurement chain."""
```

weldx/measurement.py
```diff
@@ -207,25 +830,76 @@ def _signal_label(signal):
...
+    def source_name(self) -> str:
+        """Get the name of the source.
+
+        Returns
+        -------
+        str :
+            Name of the source
+
+        """
+        return self._source.name
```
CagtayFabry, line 882: not sure if we need a property for this since it is the same length as `.source.name`

weldx/measurement.py
```diff
@@ -207,25 +830,76 @@ def _signal_label(signal):
...
 @dataclass
-class GenericEquipment:
-    """Simple dataclass implementation for generic equipment."""
+class Measurement:
+    """Simple dataclass implementation for generic measurements."""
 
     name: str
-    sources: List = field(default_factory=lambda: [])
-    data_transformations: List = field(default_factory=lambda: [])
+    data: List["TimeSeries"]
```
CagtayFabry, line 904: is this the list that links to the data that is also in the measurement chain?

CagtayFabry:
One more thought regarding the `Signal` class
https://github.com/BAMWelDX/weldx/blob/1444ae18ade34c73633921c7fe552e3fc375ba2d/weldx/measurement.py#L25-L36

could we simply inherit from `TimeSeries` here and add the few custom attributes instead of a "full" class?
That would also remove one more `.data` field 😉

weldx/util.py
vhirtham: Didn't find that one when searching for it. I added it to our [Wiki page about quantities](https://github.com/BAMWelDX/weldx/wiki/Quantities-(pint)) and removed the utility function.

scripts/welding_schema.py
vhirtham: Not sure, why an iterator would be better, but I have added the properties to return a list. We can change that to an iterator if necessary

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
vhirtham: Renamed to `source_node`

weldx/measurement.py
vhirtham: Should be doable. We can change that.

weldx/measurement.py
vhirtham: Should all the files (tag, schemas) be moved to the measurement directory then?

weldx/measurement.py
vhirtham: I think we should stick with the full docstring for public methods since it is also used for the documentation.

weldx/measurement.py
vhirtham: It is consitent with the transformations and I guess it doesn't hurt to have it.

weldx/measurement.py
vhirtham: In theory, but there is no mechanism that enforces a synchronization. I am not sure if we still need this class. It doesn't provide any additional information.

weldx/measurement.py
vhirtham: I am not sure about this particular import, but there are ring dependency issues. Can test it for this case and report back

vhirtham: Confirmed the circular dependency

tutorials/measurement_chain.ipynb
```diff
@@ -0,0 +1,553 @@
+{
```
marscher, line 1: <p>lacking a comma here: the output signal, and the output unit</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

marscher, line 1: <p><span class="markdownDiffSpan" style="background: rgb(209, 255, 221) none repeat scroll 0% 0%; display: inline-block; --darkreader-inline-bgcolor:#114627; --darkreader-inline-bgimage:none;" data-darkreader-inline-bgcolor="" data-darkreader-inline-bgimage="">The only restriction is, that the function only has a single variable which represents the input signal. </span></p><p>In this example our input variable is given by <code>x</code>.</p><p> </p><p>The paragraph about the dimensionality is somehow confusing to people not familiar with pint, I'd say. Firstly say something about the actual or desired output dimensionality, e.g. temperature, voltage, current etc. and then tell what is the expected behaviour for handling these.</p><p>I guess you almost have that, but just clarify on <code>dimensionality</code>a  little bit.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

marscher, line 1: <p>lacking a comma: <span class="markdownDiffSpan" style="background: rgb(209, 255, 221) none repeat scroll 0% 0%; display: inline-block; --darkreader-inline-bgcolor:#114627; --darkreader-inline-bgimage:none;" data-darkreader-inline-bgcolor="" data-darkreader-inline-bgimage="">we demonstrated, required </span></p><p><span class="markdownDiffSpan" style="background: rgb(209, 255, 221) none repeat scroll 0% 0%; display: inline-block; --darkreader-inline-bgcolor:#114627; --darkreader-inline-bgimage:none;" data-darkreader-inline-bgcolor="" data-darkreader-inline-bgimage="">While this is method is -&gt; While this method is</span></p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

marscher, line 1: <p>It is a method, not a function :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

marscher, line 1: <p><code>GenericEquipment -&gt; MeasurementEquipment</code></p><p> </p><p><code><span class="ql-cursor">﻿﻿</span></code><span class="markdownDiffSpan" style="background: rgb(209, 255, 221) none repeat scroll 0% 0%; display: inline-block; --darkreader-inline-bgcolor:#114627; --darkreader-inline-bgimage:none;" data-darkreader-inline-bgcolor="" data-darkreader-inline-bgimage="">define all your instruments and once and reuse their defs -&gt; define all your instruments once, and reuse their defs.</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

marscher, line 1: <p>double article: we add a an equip.</p><p> </p><p>Here you say, that we need to provide the name of the source or transformation, but earlier on get_transformation, you didn't comment on the argument. I recommend doing the same thing there.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

marscher, line 1: <p>The goal sentence is really useful and should be moved to the introduction (and eventually be rephrased here).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

marscher, line 1: <p>I'd like a concluding a small paragraph summarizing what we've just learned.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='50'/>

tutorials/measurement_example.ipynb
```diff
@@ -59,9 +59,8 @@
    "source": [
```
marscher, line 59: <p>I'd at least cross-link both notebooks. This concrete example should be given as "further reading" in the more general notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/BAMWelDX/weldx/pull/326/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

weldx/measurement.py
CagtayFabry: that's weird since this come straight from `constants` ..?

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
vhirtham: I have updated my generation script. But I will wait with the update until the changes to the schemas are applied

weldx/measurement.py
vhirtham: Managed to get it working by importing directly from `weldx.constants` as in the `__init__.py`. So importing `weldx` was the problem. Got to remember this in the future

tutorials/measurement_chain.ipynb
vhirtham: <p>Sure about the comma? I am not and grammrly doesn't complain either.</p>

weldx/measurement.py
vhirtham: > I took a quick look over the new tutorial 👍
> 
> a few notes that would make it more accessible imo:
> 
> * plot the new measurement chain layout after each change
> * add an introduction explaining the concept (without code):
>   
>   * what is a signal?
>   * what is a transformation?
> * leave out analog/digital conversion from introduction as it is too complicated to get started imo
> * if possible I would prefer more code segments between the text and shorter text-segments (if it is supposed to focus on code)
> 
> also I would like to hear your thoughts @AndreasPittner @marscher **@rwinterk** (fake ping, need to get him on GH 😄 )

Rewrote most of the introduction trying to meet your requirements. Regarding the more code, less text bullet point: Please make a suggestion where you think this is necessary. Don't think the focus here is code alone. I tried to give an overall introduction.

tutorials/measurement_chain.ipynb
vhirtham: <p>Partially rewrote the section. Please have another look at it and tell me if it is clearer now</p>

vhirtham: Done

tutorials/measurement_example.ipynb
vhirtham: I will add the link to the measurement example notebook. I have to rewrite it partially anyways.

vhirtham:
@marscher @CagtayFabry I think the current GitHub integration of the ReviewNB comments is rather poor and is currently more a pain to work with, at least at the receiving side:
- Don't see the relevant code section on GH, just `+ {` (first line of the notebook file diff) -> Have to open ReviewNB site
- No specific line given, which makes it hard to find the relevant part in a huge text block
- No code correction suggestions for simple things like a missing comma (see preceding bullet points)
- Conversations marked as resolved on ReviewNB are not resolved on GH
- The other annoying things I didn't write down and can't remember 😉 

Don't know what you think about it, but for commenting I would rather stick to pure GH.

marscher (approved):
Superb code quality @vhirtham! I couldn't find anything to complain about. I guess you already saw my comments about the tutorials. Eventually I should take another look, when you've adopted the measurement notebook again.
So far I'm approving these changes.

marscher:
> Don't know what you think about it, but for commenting I would rather stick to pure GH.

Sorry for having caused such hassle.I will note use ReviewNb again to write comments.

Lately I've seen a tool (can not recall its name right now), which converts notebooks from a plain text format, which is solely stored in the repo. This makes merging/diffing/reviewing super easy. I guess the only downside of this would be to render documentation, and having easy access for users.

vhirtham:
> > Don't know what you think about it, but for commenting I would rather stick to pure GH.
> 
> Sorry for having caused such hassle.I will note use ReviewNb again to write comments.

Hey, no problem. You couldn't have known that before trying it 😉  . Also, I think we should keep an eye on this feature. I guess they are working on improving the GH integration and it might become a valuable tool in the future. Just now I find it to be more of a liability than a helpful addition to our toolbox.

weldx/measurement.py
```diff
@@ -1,34 +1,32 @@
...
-    from networkx import DiGraph
+    from pint import Quantity
 
+    from weldx.core import MathematicalExpression, TimeSeries
 
-# measurement --------------------------------------------------------------------------
-@dataclass
-class Data:
-    """Simple dataclass implementation for measurement data."""
+# todo add examples to docstrings
```
vhirtham, line 20: Reminder

tutorials/measurement_chain.ipynb
marscher: <p>It is clearer now, thank you. </p>

marscher: <p>I guess you're right :)</p>

weldx/asdf/schemas/weldx.bam.de/weldx/measurement/measurement_chain-1.0.0.yaml
vhirtham: done

weldx/measurement.py
vhirtham: It's now implemented using the default graph serialization

vhirtham:
> One more thought regarding the `Signal` class
> https://github.com/BAMWelDX/weldx/blob/1444ae18ade34c73633921c7fe552e3fc375ba2d/weldx/measurement.py#L25-L36
> 
> could we simply inherit from `TimeSeries` here and add the few custom attributes instead of a "full" class?
> That would also remove one more `.data` field 😉

I am not sure about this. Usually, class composition should be favored over inheritance because it keeps things simple and modular. I would rather leave it as it currently is. Think it would also make the schema more complicated by adding another `allOf` pattern as in the measurement chain.

marscher:
> 
> 
> > > Don't know what you think about it, but for commenting I would rather stick to pure GH.
> > 
> > 
> > Sorry for having caused such hassle.I will note use ReviewNb again to write comments.
> 
> Hey, no problem. You couldn't have known that before trying it 😉 . Also, I think we should keep an eye on this feature. I guess they are working on improving the GH integration and it might become a valuable tool in the future. Just now I find it to be more of a liability than a helpful addition to our toolbox.

I think one is considered to look at the review on ReviewNB as well. Of course the restriction that one can only comment on a physical source line also applies there. Just noticed that, when doing another iteration.

vhirtham:
Updated the measurement example notebook. Just need to add the link to the `MeasurementChain` tutorial and some examples to the docstrings and this should finally be finished

CagtayFabry:
> > One more thought regarding the `Signal` class
> > https://github.com/BAMWelDX/weldx/blob/1444ae18ade34c73633921c7fe552e3fc375ba2d/weldx/measurement.py#L25-L36
> > 
> > could we simply inherit from `TimeSeries` here and add the few custom attributes instead of a "full" class?
> > That would also remove one more `.data` field 😉
> 
> I am not sure about this. Usually, class composition should be favored over inheritance because it keeps things simple and modular. I would rather leave it as it currently is. Think it would also make the schema more complicated by adding another `allOf` pattern as in the measurement chain.

I think we leave it as is for now and see :)

CagtayFabry (approved):
just went over the tutorials one last time, very nice !

slight changes:
- shortened transformation names to fit the plots more nicely
- hid the 'under construction' code cells for now (still in the NB)
- slight code cleanup for commented lines

LGTM 🚀

Issue #153, opened 2020-10-02
support quantities in WXRotation.from_euler()


CagtayFabry:
Now that we mostly use quantities we should add support to `WXRotation.from_euler()`

- allow quantities in ASDF schema
- use pint decorator
- maybe deprecate `tf.rotation_matrix_X` functions in favour of `WXRotation`

PR #327, opened 2021-04-01
fix WXRotation unit prefix

CagtayFabry:
## Changes
this fixes unit prefixes like "kilo" that where being ignored when constructing WXRotation

## Related Issues
related #153

## Checks

- [x] updated tests


vhirtham (approved):
Still wondering who would use kilodegrees 😛 millidegrees maybe 😉

CagtayFabry:
> 
> 
> Still wondering who would use kilodegrees 😛 millidegrees maybe 😉

Yeah it's not perfect either since back and forth conversion will map to range 0-360 so some loss is inevitable. But it might happen when the function inputs get generated by a script with unit conversions rather than user input
Aynway, better safe than sorry I guess 😉

PR #328, opened 2021-04-01
Add compare_nested to util

marscher:
## Changes

Added a compare function, which traverses nested data structures (using boltons.iterutils) and checks for equality in
the other data structure. Arrays eq checks are treated separately (np.all and xr.DataArray.identical). 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests

## Closes
#328


marscher:
Actually I wanted to use the schema of single_pass_weld as a testing input. Is there a possibility to create a valid ASDF tree from it - or can't this be done, because the binary data is missing there? I wanted to have some real world examples in my test and adding binary files to repos is almost every time a bad idea.

marscher:
@vhirtham I have a question regarding CSM, when I delete a lcs from two equally constructed (deeply copied) CSMs, I'd expect that they are not equal any more, which they currently are. Is this a bug or somehow expected?

CagtayFabry:
> 
> 
> Actually I wanted to use the schema of single_pass_weld as a testing input. Is there a possibility to create a valid ASDF tree from it - or can't this be done, because the binary data is missing there? I wanted to have some real world examples in my test and adding binary files to repos is almost every time a bad idea.

I don't think you can "open" ASDF files with the binary part missing (maybe you could with bypassing array loading and vlaidation with original ASDF files but some weldx objects will always need to access the binary parts, certainly xarray objects)

However we can produce a valid file that meets the `single_pass_weld` schema requirements by [inlining](https://asdf.readthedocs.io/en/2.7.0/asdf/arrays.html#saving-inline-arrays) the data.
I will create an updated file to merge into this branch

weldx/util.py
```diff
@@ -1201,3 +1202,58 @@ def time_ref(self, value: pd.Timestamp):
...
+            return False
+        return np.all(x == y)
+
+    elif isinstance(x, xr.DataArray):
+        if not isinstance(y, xr.DataArray):
+            return False
+        return x.identical(y)
+
+    else:
+        return x == y
```
CagtayFabry, line 1219: What do you think of something like this ?
```python
_compare_funcs = {
    (np.ndarray, pint.Quantity): lambda x, y: np.all(x == y) if (type(x) == type(y)) else False,
    (xr.DataArray, xr.Dataset): lambda x, y: x.identical(y) if (type(x) == type(y)) else False,
}


def _compare(x, y):
    for key, item in _compare_funcs.items():
        if type(x) in key:
            return item(x, y)
    return x == y
```

marscher:
That's pretty elegant!

On April 2, 2021 9:12:07 AM UTC, Cagtay Fabry ***@***.***> wrote:
>@CagtayFabry commented on this pull request.
>
>
>
>> +def _compare(x, y):
>+    if isinstance(x, np.ndarray):
>+        if not isinstance(y, np.ndarray):
>+            return False
>+        return np.all(x == y)
>+
>+    elif isinstance(x, xr.DataArray):
>+        if not isinstance(y, xr.DataArray):
>+            return False
>+        return x.identical(y)
>+
>+    else:
>+        return x == y
>
>What do you think of something like this ?
>```python
>_compare_funcs = {
>(np.ndarray, pint.Quantity): lambda x, y: np.all(x == y) if (type(x) ==
>type(y)) else False,
>(xr.DataArray, xr.Dataset): lambda x, y: x.identical(y) if (type(x) ==
>type(y)) else False,
>}
>
>
>def _compare(x, y):
>    for key, item in _compare_funcs.items():
>        if type(x) in key:
>            return item(x, y)
>    return x == y
>```
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/BAMWelDX/weldx/pull/328#pullrequestreview-626879912

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

weldx/util.py
marscher: That is pretty elegant :)

vhirtham: nice one, even though I remember that we had some trouble with the xarray `identical` function. But we can sort it out when it gets a problem

vhirtham:
> @vhirtham I have a question regarding CSM, when I delete a lcs from two equally constructed (deeply copied) CSMs, I'd expect that they are not equal any more, which they currently are. Is this a bug or somehow expected?

Should be a bug. Have to take a look into it

vhirtham:
Maybe you can combine some tests using `pytest.parametrize`.

weldx/tests/test_utility.py
```diff
@@ -498,3 +501,137 @@ def test_xr_time_ref():
...
+        "x": {
+            0: [1, 2, 3],
+        },
+        "blub": True,
+    }
+    b = dict(a)
+    assert ut.compare_nested(a, b)
+
+
+class TestWeldxExampleCompareNested(unittest.TestCase):
```
CagtayFabry, line 584: what is the reason fpr using `unittest.TestCase`?
does it play nice with pytest?

marscher: it does play along with pytest nicely. It is used to load the data only once. I could have used a pytest.fixture as well, but I'm more used to this pattern from the past. If you want I can rewrite it using pytest fixtures only.

marscher:
> Maybe you can combine some tests using `pytest.parametrize`.

Yes I will do that. Good suggestion.

weldx/tests/test_utility.py
CagtayFabry: ok sounds nice, maybe you can give a quick intro into this style soon. :)

marscher:
> Maybe you can combine some tests using `pytest.parametrize`.
@vhirtham 
I have to admit, that the readability of the parameterized test (which now only consists out of one line, e.g. the assertion) is not very high. This might be due to the dictionaries being passed, which are kind of large. I could try to shrink the test data without sacrificing a case, but the other style of one function per case makes it more understandable IMO. This is due to having a proper name for the case and pythonic data definitions. In this case it is not even possible to manipulate the data easily, but you really have to duplicate it to modify it (which could also be a good thing in terms of side-effects).
What do you think?

vhirtham:
> > Maybe you can combine some tests using `pytest.parametrize`.
> > @vhirtham
> > I have to admit, that the readability of the parameterized test (which now only consists out of one line, e.g. the assertion) is not very high. This might be due to the dictionaries being passed, which are kind of large. I could try to shrink the test data without sacrificing a case, but the other style of one function per case makes it more understandable IMO. This is due to having a proper name for the case and pythonic data definitions. In this case it is not even possible to manipulate the data easily, but you really have to duplicate it to modify it (which could also be a good thing in terms of side-effects).
> > What do you think?

Well, it isn't very readable indeed. However, I think there are some things left we can do about it. I'll have a look later this day.

weldx/tests/test_utility.py
marscher: I can do this on our jour fixe, like 15 minutes should be sufficient to get the full idea since all of us are familiar with unit testing, right?

marscher:
Would it be useful to also pass the comparator function to compare_nested? So we could not only compare for equality but for "less", "greater equal", and so forth.
@vhirtham, @CagtayFabry

weldx/tests/test_utility.py
CagtayFabry: great 👍

CagtayFabry:
> 
> 
> Would it be useful to also pass the comparator function to compare_nested? So we could not only compare for equality but for "less", "greater equal", and so forth.
> @vhirtham, @CagtayFabry

Not sure where we would actually use this right now

marscher:
> 
> 
> > Would it be useful to also pass the comparator function to compare_nested? So we could not only compare for equality but for "less", "greater equal", and so forth.
> > @vhirtham, @CagtayFabry
> 
> Not sure where we would actually use this right now

Me neither, it would just render this function more versatile. But if we do not need it, we shouldn't do it.

.github/workflows/pytest.yml
```diff
@@ -56,7 +56,7 @@ jobs:
 
       - name: run pytest
         run: |
-          pytest -n 2 --runslow
+          pytest -n 2 --runslow --dist=loadfile
```
CagtayFabry, line 59: should we put this as default in `setup.cfg` ?

marscher: I just tested this distribution method with the notebooks and it works. But if we put this option, we would also have to specify a parallel execution method (e.g. the number of nodes --n auto or --n1 etc.). Is this still an option for setup.cfg then?

marscher: There are some design patterns in test_transformations, which prohibit to run the tests in an arbitrary parallel order. So we have to enforce this particular distribution method.

marscher: @vhirtham fyi

marscher: My only concern about adding this as a default option is that we would also need to add an option for the number of hosts (e.g. -n1 etc.), because without it you receive an error message like this:
```ERROR: MISSING test execution (tx) nodes: please specify --tx```

Comment (CagtayFabry) [reply .github/workflows/pytest.yml]: yes if this need `-n` than keep the current implementation specified in GitHub actions 👍

Comment (marscher) [reply .github/workflows/pytest.yml]: I decided to put -n1 as default, which can be overridden from the outside. Do you agree?

**Comment by marscher** [comment]
This is finished, imho :)

File: .github/workflows/pytest.yml (hunk 3 above)
Comment (CagtayFabry) [reply .github/workflows/pytest.yml]: yes that is a nice solution!

File: weldx/constants.py [hunk 4]
```diff
@@ -1,11 +1,11 @@
 """Define constants for global library use."""
-from pathlib import Path
```
CagtayFabry, line 2: I think using `Path` is common here, any collisions?

weldx/util.py
```diff
@@ -1201,3 +1193,78 @@ def time_ref(self, value: pd.Timestamp):
                 self._obj.time.attrs["time_ref"] = value  # set new time_ref value
             else:
                 self._obj.time.attrs["time_ref"] = value
+
+
+class _Eq_compare_nested:
+    """Compares nested data structures like lists, sets, tuples, arrays, etc."""
+
+    compare_funcs = {
+        (np.ndarray, pint.Quantity): lambda x, y: np.all(x == y),
```
CagtayFabry, line 1202:
```suggestion
        (np.ndarray, pint.Quantity, pd.Index): lambda x, y: np.all(x == y),
```
this one should be pretty common as well

weldx/util.py
```diff
@@ -1201,3 +1193,78 @@ def time_ref(self, value: pd.Timestamp):
...
+        1. If the path does not exist in `b` a KeyError will be raised.
+        2. If the index `k` does not exist an IndexError is raised.
+        3. If the other path exists, a comparison will be made using `_compare`.
+           When the elements are not equal traversing `a` will be stopped
+           by raising a ValueError.
+        """
+        other_value = iterutils.get_path(b, p)[k]
+        if not _Eq_compare_nested._compare(v, other_value):
+            raise ValueError
+        return True
```
CagtayFabry, line 1230:
```suggestion
        if not iterutils.default_enter(p, k, v)[1]:
            if not _Eq_compare_nested._compare(v, other_value):
                raise ValueError(f"{v=}\n{other_value=}")
        return True
```
We need to make sure not call the compare function on items that should be traversed deeper because we would end up calling `__eq__` on these as well (and failing)

try the following:
```python
a = {"l1": {"l2": np.arange(5)}}
b = {"l1": {"l2": np.arange(5)}}
assert ut.compare_nested(a, b)
```

The solution above is a little ugly but works...

Also the ValueError should report the compared instances for clarity

weldx/util.py
```diff
@@ -1201,3 +1193,78 @@ def time_ref(self, value: pd.Timestamp):
...
+            When a or b is not a nested structure.
+
+        """
+        visit = functools.partial(_Eq_compare_nested._visit, b=b)
+
+        try:
+            iterutils.remap(a, visit=visit, reraise_visit=True)
+        # Key not found in b, values not equal, more elements in a than in b
+        except (KeyError, ValueError, IndexError):
+            return False
```
CagtayFabry, line 1263:
```suggestion
            raise
```
I would prefer just to raise the error because otherwise this is really hard to debug if the Exception gets lost

weldx/util.py
```diff
@@ -1201,3 +1193,78 @@ def time_ref(self, value: pd.Timestamp):
...
+
+    compare_funcs = {
+        (np.ndarray, pint.Quantity): lambda x, y: np.all(x == y),
+        (xr.DataArray, xr.Dataset): lambda x, y: x.identical(y),
+    }
+
+    @staticmethod
+    def _compare(x, y):
+        if type(x) is not type(y):
+            return False
```
CagtayFabry, line 1209:
```suggestion
        x, y = (np.asarray(v) if isinstance(v, NDArrayType) else v for v in (x, y))
        if type(x) is not type(y):
            return False
```
When reading ASDF trees numpy arrays get loaded as `NDArrayType` which can be confusing
This should be a simple fix to catch that case

CagtayFabry (changes requested):
I will add another complex test case to use this function

.github/workflows/pytest.yml
CagtayFabry: I noticed a problem with PyCharm integration
When using `xdist` the debugger does not attach to the test instances, anyone else? This happens even for using a single xdist runner :(

So in that case I would suggest to remove xdist arguments from the default config to make it easier to use (not ideal I know)

marscher: Actually PyCharms debugger should work fine with subprocesses. If multiple test workers are an issue for the debugger, I agree to be explicit for this option in CI environments.

weldx/util.py
marscher: I raise the value error just to break out of the iteration, not to make any useful error messages for the user, since the outer function just returns a boolean (I'd prefer such an interface for simplicity).

weldx/util.py
marscher: Key and Index errors are being raised if something is not accessible in the other data structure. IMO returning false here is correct. TypeErrors are re-raised.

weldx/util.py
marscher: The asdf tree test cases did not fail for NDArrayType, so I guess this is fine.

CagtayFabry: I think this is because the example does not include "raw" numpy arrays that ar enot nested in dataclasses/xarray or simliar

.github/workflows/pytest.yml
CagtayFabry: I'm not sure what works exactly, we can take a look at it together tomorrow :)

weldx/util.py
marscher: Does ASDF perform this wrapping only, if arrays are being red lazily?

CagtayFabry: good question, I did not look into it that deeply but it sounds reasonable

weldx/tests/asdf_tests/test_asdf_validators.py
```diff
@@ -163,6 +164,8 @@ def test_shape_validation_error_exception(shape, exp, err):
         assert _val(shape, exp)
 
 
+@pytest.mark.parametrize("copy_arrays", [True, False])
+@pytest.mark.parametrize("lazy_load", [True, False])
```
CagtayFabry, line 168: I have added the different loading parameters here for testing but it seems like that is not the issue
arrays apparently get read as `NDArrayType` @marscher

weldx/tests/test_utility.py
```diff
@@ -498,3 +499,117 @@ def test_xr_time_ref():
...
+    a, b = _default_dicts
+    del b["x"]
+    assert not ut.compare_nested(a, b)
+
+
+def test_key_changed2(_default_dicts):
+    a, b = _default_dicts
+    x = b.pop("x")
+    b["y"] = x
+    assert not ut.compare_nested(a, b)
```
vhirtham, line 574: Maybe group them all in a class `TestCompareNested`.

marscher: will do. Thanks.

marscher: any idea, if there is a good workaround for the need of parameterized methods to be "staticmethod"?

weldx/util.py
vhirtham: Agree with @marscher here. Exceptions that are raised during the comparison due to missing keys in one of the compared dicts shouldn't be propagated. The function should simply return not equal e.g. `False`.

weldx/tests/test_utility.py
marscher: ah it simply works with passing self. Sorry for the noise.

vhirtham: No, but why is this a problem?

marscher: When you parameterize tests without marking them static, the first argument gets interpreted as self and collides with the pytest.mark logic. But the parameterize decorator is aware of the self argument (when you give it).

weldx/constants.py
marscher: It did not cause any collisions, I just wanted to cleanup the namespace a bit. Some IDEs/frontends also show all "public" attributes of a module and ignore the meaning of `__all__`

weldx/tests/test_utility.py
vhirtham: I know, just wanted to why not simply make them static unless they need access to the test class ;)

marscher: a class containing only static methods defeats the purpose of a class IMO (like I did it for this comperator thing here :D). Its also pretty redundant to declare all methods static. If we consider using unittest.TestCase as base, it would make even more sense to have non-static members.

CagtayFabry: I like having test classes for the convenience of grouping tests (and running them all) as well so 👍 for using that

marscher:
These types are now covered by the test suite:
```
{<class 'weldx.welding.groove.iso_9692_1.VGroove'>,
 <class 'asdf.tags.core.ndarray.NDArrayType'>,
 <class 'str'>,
 <class 'NoneType'>,
 <class 'int'>,
 <class 'float'>,
 <class 'bool'>,
 <class 'pandas._libs.tslibs.timestamps.Timestamp'>,
 <class 'pandas.core.indexes.datetimes.DatetimeIndex'>,
 <class 'pandas.core.indexes.timedeltas.TimedeltaIndex'>,
 <class 'pint.quantity.build_quantity_class.<locals>.Quantity'>,
 <class 'weldx.core.TimeSeries'>,
 <class 'weldx.transformations.local_cs.LocalCoordinateSystem'>,
 <class 'weldx.transformations.cs_manager.CoordinateSystemManager'>,
 <class 'weldx.welding.processes.GmawProcess'>,
 <class 'weldx.asdf.tags.weldx.aws.process.shielding_gas_for_procedure.ShieldingGasForProcedure'>,
 <class 'weldx.measurement.Measurement'>,
 <class 'weldx.measurement.Signal'>,
 <class 'weldx.measurement.GenericEquipment'>,
 <class 'numpy.ndarray'>}
```

weldx/util.py
```diff
@@ -1205,15 +1205,34 @@ class _Eq_compare_nested:
 
     @staticmethod
     def _compare(x, y):
-        if type(x) is not type(y):
-            return False
+        # treat special cases first
+        # rely upon the fact, that comparison would fail, if y is of an incompatible type.
```
CagtayFabry: I think the question here is if we want to test for equality (meaning in types as well) or "just" compare compatibilty with `__eq__`?
Why not test for type equality?
I would prefer having the security for cases where it would matter (ASDF file roundtrips)

weldx/util.py
```diff
@@ -1222,12 +1241,19 @@ def _visit(p, k, v, b):
...
+        other_value = other_data_structure[key]
+        if not _Eq_compare_nested._enter(None, key, value)[1]:
+            # check lengths of Sequence types first and raise prior starting a more expensive comparison!
+            if isinstance(other_data_structure, Sequence) and len(
+                other_data_structure
+            ) != len(iterutils.get_path(a, path)):
+                raise RuntimeError
+
+            if not _Eq_compare_nested._compare(value, other_value):
+                raise RuntimeError
```
CagtayFabry, line 1248: Even if it doesn't show up, can you add a small description of the error case in the `RuntimeError` here?

weldx/util.py
marscher: Type equality is not very pythonic. We already encountered the case of the ndarray wrapper, for which we would have to create an ugly workaround. How does exact type matching increase security and what are ASDF file roundtrips?

weldx/util.py
marscher: ok, but I honestly do not see why it would be useful :)

weldx/tests/test_utility.py
```diff
@@ -498,3 +499,125 @@ def test_xr_time_ref():
...
+            "foo": np.arange(3),
+            "x": {
+                0: [1, 2, 3],
+            },
+            "bar": True,
+        }
+        b = copy.deepcopy(a)
+        return a, b
+
+    @pytest.mark.parametrize(["a", "b"], [("asdf", "foo"), (b"asdf", b"foo"), (1, 2)])
```
vhirtham, line 519: Might be picky and a matter of taste, but I like one line per test case. If the auto-formatting is an issue, just put a `,` behind the last tuple

marscher: yes, this has been formatted by black. Thanks for the comma hint ;)

weldx/util.py
```diff
@@ -1201,3 +1193,97 @@ def time_ref(self, value: pd.Timestamp):
...
+    def _compare(x, y):
+        # 1. treat special cases first
+        # 2. rely upon the fact, that comparison would fail,
+        #    if y is of an incompatible type.
+        for types, func in _Eq_compare_nested.compare_funcs.items():
+            if isinstance(x, types):
+                return func(x, y)
+            return x == y
+
+        return x == y
```
vhirtham, line 1216: Maybe it is the missing coffee, but I don't get this. To my understanding, the loop can't execute more than once, since the body always returns. So it is obsolete. As a further consequence, the `return` after the loop should be unreachable. What am I missing?

weldx/util.py
CagtayFabry: If I save a dictionary into an ASDF file and read it back into a new dict it would be good to know everything gets deserialized into the original type (in some cases)
personally I would like to rely on the method to test for this as well and would prefer to catch a few known exceptions over having some ambiguity

weldx/util.py
marscher: I was missing coffee too, I just spotted it as well. The return inside the loop made no sense.

weldx/util.py
marscher: So the only case here (so far) is the asdf ndarray wrapping, am I right?

weldx/tests/asdf_tests/test_asdf_validators.py
CagtayFabry, line 168: we can remove this before merging @marscher

weldx/util.py
CagtayFabry: yes not aware of anything else so far

weldx/tests/asdf_tests/test_asdf_validators.py
marscher: Why, didn't this cover the ndarray/NDArrayType pair?

weldx/util.py
```diff
@@ -1201,3 +1193,97 @@ def time_ref(self, value: pd.Timestamp):
...
+
+    @staticmethod
+    def _enter(path, key, value):
```
vhirtham, line 1219: Maybe thats just me, but I like to have docstrings and return type hints on private class methods using the same style as for public methods. Makes it easier to understand for someone who might touch the code in the future (or reviews it 😛).

marscher: These methods override functionality of boltons.iterutils, which doesn't use type annotations as well. I can add more documentation, if you feel like it is non yet sufficient (but I'd be delighted, if you could point me towards it, as I'm already too deeply involved with this code).

weldx/tests/asdf_tests/test_asdf_validators.py
marscher: Ah I see, it also in another spot. Thanks

weldx/util.py
vhirtham: I guess we would have seen it in the coverage if we hadn't spotted it before 😄

weldx/util.py
marscher: I've added a list of tuples to handle future cases of enforced type equalities.

weldx/util.py
vhirtham: I am not familiar with itertools. I can somehow anticipate what the inputs are, but what is the purpose of the second return value (bool)? I can google it, but the point is: 
I like having docstrings on internal functions in general until they are simple one-liners. We all have different knowledge bases and something that is super obvious to someone might be a riddle to another person (Including the future you 😛 ). I know that @CagtayFabry has a somewhat different opinion regarding this topic and we should probably discuss a general procedure. So no need to change something here and treat it as an annotation

marscher: Thanks! I added more comments due to your remarks.
I totally agree to have some kind of documentation to aid the understanding of the code.
It is always a tradeoff how much to document. A quick line comment should sufficient for private methods in my opinion, if it is not too complex (e.g. weird behavior or lots of code). Large numpydoc styled strings for very simple functionality blows up the code and reading the actual code instead of the docstring would lead to a quicker understanding. This is of course not the case for something non-obvious or weird behavior, but then some comment lines should also do the trick.

CagtayFabry (approved):

PR #329, opened 2021-04-06
clean notebook IDs

CagtayFabry:
## Changes
adds a small function `util._clean_notebook_ids` to remove the autogenerated JuypterLab `ID` field from notebooks for better diffs when needed.


vhirtham:
Maybe we should put this into the scripts folder and add a script that automatically clears all our notebooks

CagtayFabry:
> 
> 
> Maybe we should put this into the scripts folder and add a script that automatically clears all our notebooks

good idea :)

marscher (approved):
LGTM

scripts/clean_notebooks.py
```diff
@@ -0,0 +1,11 @@
+# clean up jupyter notebooks in main directory
+if __name__ == "__main__":
+    from pathlib import Path
+
+    from weldx.util import _clean_notebook_ids
+
+    p = Path(__file__).parents[1]
+    print(p)
```
vhirtham, line 8: Supposed to stay? 😛

vhirtham (approved):
Maybe we can also execute

~~~
jupyter nbconvert --clear-output --inplace NOTEBOOKNAME.ipynb
~~~

per notebook to clear the cells too

scripts/clean_notebooks.py
vhirtham, line 8:
```suggestion
```

CagtayFabry:
> 
> 
> Maybe we can also execute
> 
> ```
> jupyter nbconvert --clear-output --inplace NOTEBOOKNAME.ipynb
> ```
> 
> per notebook to clear the cells too

I like it, I removed the outputs (and execution_count) manually so that there would be no need for nbconvert as a dependency

vhirtham (approved):
Really nice. Maybe I will set it up as a local pre-commit hook. Will definitely save some time.

marscher:
nbdime will be your friend in terms of pre-commit hooks (and notebook diffs etc.)

PR #332, opened 2021-04-06
fix docs build failing

CagtayFabry:
not sure why docs builds are failing suddenly so testing some older versions

### DIFF
--- a/doc/rtd_environment.yml
+++ b/doc/rtd_environment.yml
@@ -15,6 +15,12 @@ dependencies:
   - pydata-sphinx-theme=0.5.0
   - numpydoc>=0.5
   - sphinx-autodoc-typehints
+  # debug
+  - decorator=4.4.2
+  - docutils=0.17
+  - nbformat=5.1.2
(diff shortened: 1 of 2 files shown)


CagtayFabry:
seems like docutils 0.17 broke some things in the notebooks :/

CagtayFabry:
The docs build is broken with the new `docutils` 0.17 so pinned to 0.16 for now to keep it running @vhirtham @marscher 

We should open an issue and look into it at some point, error messages can be seen in the Actions

marscher (approved):
Thanks for taking care of this. Has this already been reported upstream? So the problem is, that the titles of the notebooks are not being recognized, right?

CagtayFabry:
> 
> 
> Thanks for taking care of this. Has this already been reported upstream? So the problem is, that the titles of the notebooks are not being recognized, right?

I did a quick google search but couldn't find anything so not sure
But I didn't dig into the problem either, maybe something is wrong with our notebook formatting? (I don't think so but who knows)

PR #336, opened 2021-04-13
[ci/appveyor] use conda-recipe for bld/test

marscher:
## Changes

- moved environment.yml to devtools (if still desired). I remember Rene asked to keep it.
- use conda-recipe for building/testing in Appveyor.

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/


marscher:
Should I add an changelog entry for this?

vhirtham:
> Should I add an changelog entry for this?

Don't think so since it has no effect on the package.

marscher:
It puzzles me, that the build_env.yaml is being looked up in the root directory. I thought it should always lie next to meta.yaml

```Error: ENOENT: no such file or directory, open '/home/runner/work/weldx/weldx/conda.recipe/build_env.yml'```

**Comment by marscher** [comment]
Eventually this is cached somehow?

**Comment by CagtayFabry** [comment]
update the recipe for new dir layout with `devtools`, somehow `conftest.py` is not beeing found..?

**Comment by CagtayFabry** [comment]
maybe we should remove the appveyor CI, what do you think @vhirtham ?

File: devtools/conda.recipe/meta.yaml [hunk 1]
```diff
@@ -2,18 +2,18 @@
 
 package:
   name: weldx
-  version: {{ data["version"] }}
+  version: {{ data.get("version") }}
```
marscher, line 5: I changed it, because locally I received an error like this (with a very recent version of conda-build):
```INFO:conda_build.metadata:Attempting to finalize metadata for weldx
Error: Failed to render jinja template in /home/marscher/sources/weldx/devtools/conda.recipe/meta.yaml:
'data' is undefined
```

CagtayFabry (approved):
looking good now I think :)

marscher:
Thanks. Now what about Appveyor? Should it stay or should it go? :)

vhirtham:
> Thanks. Now what about Appveyor? Should it stay or should it go? :)

Not sure. In contrast to Travis it is not utterly useless to us since it doesn't have the same build time limits. As long as there is no further maintenance overhead I am fine with both (stay or go).

marscher:
I guess the overhead will be low, now that we can steer everything from the conda recipe

CagtayFabry:
yes should be minimal overhead now :)

PR #338, opened 2021-04-14
Get yaml header read only header

marscher:
## Changes

The get_yaml_header function used to read the whole file into memory (which is prohibitive for large files).
Now it only reads until the header is finished, e.g. the binary part of the ASDF file starts.
Optionally the function now returns the parsed YAML header as a dictionary.
Also added a test for allowed inputs and the new parse option.


CagtayFabry:
this makes a lot of sense 👍 

please update the CHANGELOG with the new option of returning a parsed dict

marscher:
Thanks, I will update the changelog then.

marscher:
this should be ready.

weldx/asdf/util.py
```diff
@@ -141,30 +141,38 @@ def write_read_buffer(
...
     else:
         with open(file, "rb") as f:
-            code = f.read()
+            code = read_header(f)
 
-    parts = code.partition(b"\n...")
-    code = parts[0].decode("utf-8") + parts[1].decode("utf-8")
-    return code
+    if parse:
+        return yaml.load(code, yaml.BaseLoader)
```
CagtayFabry, line 174:
```suggestion
        return asdf.yamlutil.load_tree(code)
```
I have been wondering about this for a while, turns out it is quite simple

With the change above we will return a `asdf.tagged.TaggedDict` which is basically a dict that keeps all the ASDF `tag` definitions in a `_tag` attribute (for all nested types)
This allows us to keep all the tag/class information from the yaml header as well 🚀

CagtayFabry (changes requested):

weldx/asdf/util.py
marscher: That's really nice, cheers!

CagtayFabry: we should keep this in mind when handling the extracted YAML sections from files in KISA, this should give us all the original file information needed without the binary bits @marscher

marscher:
The tag information of the TaggedDict is private and so I guess is only meant to be used internally. But if we really want, we could use some day. Lets just leave it like this for now.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 ### changes
 - `WXRotation.from_euler()` now accepts a `pint.Quantity` as input. [[#318]](https://github.com/BAMWelDX/weldx/pull/318)
 - move tests folder to `weldx/tests` [[#323]](https://github.com/BAMWelDX/weldx/pull/323)
+- `get_yaml_header` received a new option parse, which optionally 
+  returns the parsed YAML header as dictionary. [[#338]](https://github.com/BAMWelDX/weldx/pull/338)
```
CagtayFabry, line 14:
```suggestion
- `get_yaml_header` received a new option parse, which optionally returns the parsed YAML
  header as `asdf.tagged.TaggedDict`. [[#338]](https://github.com/BAMWelDX/weldx/pull/338)
```

marscher:
That's more precise, thank you!

CagtayFabry:
> 
> 
> The tag information of the TaggedDict is private and so I guess is only meant to be used internally. But if we really want, we could use some day. Lets just leave it like this for now.

yes I'm sure it will come in handy later

CagtayFabry (approved):

weldx/asdf/util.py
marscher: Yes. I changed this function precisely for the broad usage in KISA. When we will be handling lots of large files it would have been overkill to read them all to memory just for the sake of extracting the small header.

PR #341, opened 2021-04-16
Add weldx file wrapper

### LINKED ISSUE #271: asdffile wrapper discussion
We should wrap the most common functionalities needed by users/KISA project.
This far from complete lists contains:
- creating file handles 
- reading the tree
- updating the tree
- adding history information (we should discuss here as well, what would be useful)
- validate against a schema

marscher:
## Changes

This adds an ASDF file wrapper to ease some 

### Features
- Dictionary interface.
- Provides an optional interface to access the dictionary via attributes.
- Handle file objects internally, e.g. create buffers automatically.
- Context manager interface, synchronizing file contents.
- Expose only the most mandatory ASDF interfaces to users.
- Easy visualization of the Asdf header.
- History entries with custom software entries.

## Related Issues

Closes #271 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/asdf/file.py
```diff
@@ -0,0 +1,357 @@
...
+        software_history_entry: Mapping = None,
+    ):
+        # TODO: default asdf_args?
+        # e.g: asdf_args = {"copy_arrays": True}
+        if asdf_args is None:
+            asdf_args = {}
+
+        self._quality_standard = (
+            asdf_args["custom_schema"] if "custom_schema" in asdf_args else None
+        )
```
CagtayFabry, line 86:
```suggestion
        self._custom_schema = (
            asdf_args["custom_schema"] if "custom_schema" in asdf_args else None
        )
```
might get confusion with the "full" quality standards later so I suggest sticking to ASDF naming

marscher: OK thanks, I rename it to custom_schema then.

weldx/asdf/file.py
```diff
@@ -0,0 +1,471 @@
...
+
+        header = get_yaml_header(self.file_handle, parse=True)
+        tree = Tree()
+        for x in header:
+            node = Node(x)
+            tree.add_node(node)
+            for child in x:
+                child_node = Node(child)
+                node.add_node(child_node)
+        return tree
```
CagtayFabry, line 441: I think using `ipytree` is the best way to go for more complex and interactive file browsing. I looked into it a little at the beginnig but the `JSON` display was just convenient

one idea could be to implement `_as_ipytree` methods for our custom classes like LCS, CSM etc. that attach their own subtrees with custom information

marscher: Building the tree incrementally for different high-level objects seems like a very good idea. I was just looking for a more "interactive" widget to handle large trees. I thought we already had a widget doing this job, but was unable to find it.

tutorials/GMAW_process.ipynb
```diff
@@ -145,20 +146,15 @@
...
-   "source": [
-    "buffer = weldx.asdf.util._write_buffer(tree)\n",
-    "weldx.asdf.util.notebook_fileprinter(buffer)"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
+   "metadata": {
```
marscher, line 149: pycharm sucks for notebooks :)

marscher:
somehow there is not autosummary being generated for the top level types module, and I do not know why. I tried importing it in the top level init (which is undesired). Importing the types somewhere else does not enforce the generation of a stub. If I place the rst file for autodoc manually everything works like expected.

marscher:
cd'ing out of test source directory helps to solve this import conflict thing pytest raised, but now we have to decide how we proceed with the missing "scripts" folder. Currently the single_pass_weld.py is used as a fixture in several tests. When this is not deployed/installed, we should decide where to move these scripts, so that they can be used in an installed version of weldx.
I'd suggest to have entry points for these scripts (so they can be used from the command line) and move the contents somewhere within the core package.

@CagtayFabry, @vhirtham what do you think where we should keep these scripts?

CagtayFabry:
> cd'ing out of test source directory helps to solve this import conflict thing pytest raised, but now we have to decide how we proceed with the missing "scripts" folder. Currently the single_pass_weld.py is used as a fixture in several tests. When this is not deployed/installed, we should decide where to move these scripts, so that they can be used in an installed version of weldx.
> I'd suggest to have entry points for these scripts (so they can be used from the command line) and move the contents somewhere within the core package.
> 
> @CagtayFabry, @vhirtham what do you think where we should keep these scripts?

If this is solved by #350 that's fine by me

weldx/tests/asdf_tests/test_weldx_file.py
```diff
@@ -0,0 +1,307 @@
...
+
+class WritableFile:
+    """Example of a class implementing SupportsFileReadWrite."""
+
+    def __init__(self):
+        """."""
+        self.to_wrap = BytesIO()
+
+    def read(self, *args, **kwargs):
+        """."""
```
CagtayFabry, line 46: I guess these are to silence pydocstyle? I would prefer to set the corresponding `# noqa: D...` instead for each function (or at top level in the file/class even if that is supported)

weldx/util.py
```diff
@@ -1335,3 +1336,17 @@ def compare_nested(
 
 
 compare_nested = _Eq_compare_nested.compare_nested
+
+
+def is_interactive_session() -> bool:
```
CagtayFabry, line 1341: I would be really interested to see if/how we can distinguish between jupyter notebook/lab here

CagtayFabry (approved):
I didn't get to review all code segments in detail but I think we should start using this as soon as possible to get a feel for it :)

weldx/util.py
marscher: It doesn't seem to be possible as both environments seem to be equal from within the ipython object. 
There has been a request to enable the display of JSON, but there was no response since last year. So I guess this has no priority at all. https://github.com/jupyter/notebook/issues/5446

tutorials/GMAW_process.ipynb
vhirtham: Can't we add a proper treatment of that to the `clean_notebooks` script? Maybe we can also handle this python version ping pong too.

weldx/asdf/file.py
```diff
@@ -0,0 +1,514 @@
...
+        self, fd: Optional[types_path_and_file_like] = None, **write_args
+    ) -> Optional[BytesIO]:
+        """Write current weldx file to given file name or file type.
+
+        Parameters
+        ----------
+        fd :
+            May be a string path to a file, or a Python file-like
+            object. If a string path, the file is automatically
+            closed after writing. If None is given, write to a new buffer.
```
vhirtham, line 387:
```suggestion
            closed after writing. If `None` is given, write to a new buffer.
```

weldx/asdf/file.py
```diff
@@ -0,0 +1,514 @@
...
+         'software': {'name': 'MyTool', 'author': 'MeSoft', \
+         'homepage': 'https://startpage.com', 'version': '1.0'}}
+
+        """
+        return self._DEFAULT_SOFTWARE_ENTRY
+
+    @software_history_entry.setter
+    def software_history_entry(self, value):
+        if value is None:
+            from weldx import __version__ as version
```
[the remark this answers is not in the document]
marscher: eventually, because it is a property setter?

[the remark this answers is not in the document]
vhirtham: Didn't know that pydocstyle ignores setters in general

marscher:
A setter will never be rendered in sphinx, probably that's why.

vhirtham:
> A setter will never be rendered in sphinx, probably that's why.

You are right: https://stackoverflow.com/questions/17478040/how-can-i-generate-documentation-for-a-python-property-setter-using-sphinx

Good to know.

marscher:
Unfortunately there is no way to determine, if a code is running in jupyterlab or notebook. So I suggest, we fall back to set use_widgets to false and lab users can enable it on their behalf.

PR #344, opened 2021-04-26
[tutorials] remove cd cmd prior imports

### LINKED ISSUE #342: notebooks should not cd into parent directory
I think it is confusing to have this optional instruction for users and is not best practice. Output files will be written (if created), will not be written next to the notebooks, but in the parent, which is unexpected.
We should ensure, that the package is installed prior execution. So we should mention to install (which now should be super easy with pip/conda). On binder it is already installed, right?

marscher:
## Changes

The cd cmd prior import is historical, when there was no installation
available.

## Related Issues

Closes #342

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/

### DIFF
--- a/tutorials/GMAW_process.ipynb
+++ b/tutorials/GMAW_process.ipynb
@@ -17,18 +17,6 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "nbsphinx": "hidden"
-   },
-   "outputs": [],
-   "source": [
-    "# if the package is not installed in your python environment, run this to execute the notebook directly from inside the GitHub repository\n",
-    "%cd -q .."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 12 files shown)


marscher:
note that this collapsed state thing was removed by nbconvert --clear-output

CagtayFabry:
> 
> 
> note that this collapsed state thing was removed by nbconvert --clear-output

do we have to remove it? Not that it's all that important just wondering

marscher:
> 
> 
> > note that this collapsed state thing was removed by nbconvert --clear-output
> 
> do we have to remove it? Not that it's all that important just wondering

We do not have to remove, but I guess since we agreed to always clear the notebooks output, it will be removed in the future anyways. I also think that the collapsed state is nothing which should be stored.

CagtayFabry:
> > > note that this collapsed state thing was removed by nbconvert --clear-output
> > 
> > 
> > do we have to remove it? Not that it's all that important just wondering
> 
> We do not have to remove, but I guess since we agreed to always clear the notebooks output, it will be removed in the future anyways. I also think that the collapsed state is nothing which should be stored.

fine by me, can you check since this is in the geometry tutorials @vhirtham ?

CagtayFabry (approved):

vhirtham (approved):

vhirtham:
I think the collapsed field is just leftover from the time we didn't have dedicated plot functions and wanted to hide those big code blocks in the tutorial. So it should be safe to remove them.

PR #350, opened 2021-04-30
minor fixes

marscher:
## Changes

- enable doc tests for pytest runs (setup.cfg)
- fix conftest import conflict for pip dist install
- tidy up scripts directory


marscher:
This import conflict somehow makes sense to me. There is the installed location containing conftest.py (e.g. the distribution which just got installed via pip prior invoking pytest), then the source directory (from which we invoke pytest).
It is unclear, why this just happens now. I think pip or pytest has somehow changed upstream.

The simplest workaround would be to invoke pytest for the distribution test from another directory than the source directory.

.github/workflows/build.yml
```diff
@@ -118,7 +118,9 @@ jobs:
       - name: run pytest
         run: |
           pip install pytest pytest-cov pytest-xdist
-          pytest -n2 --dist=loadfile
+          cd /tmp
+          pytest -n2 --dist=loadfile --pyargs weldx
+          cd -
```
CagtayFabry, line 123: can you share why this is necessary?

CagtayFabry (approved):

.github/workflows/build.yml
marscher: https://github.com/BAMWelDX/weldx/runs/2475826648#step:62 Because of this error. Somehow, if we run the tests from a dist install within the source directory, an import conflict appears. Running the tests elsewhere cures this.

PR #351, opened 2021-05-03
Fix clean_notebooks

vhirtham:
## Changes

Fix path in clean_notebooks


marscher (approved):
Ah so someone is using this script :) Thanks for fixing the path.

vhirtham:
> Ah so someone is using this script :) Thanks for fixing the path.

When working on the tutorials, it is really useful

PR #353, opened 2021-05-07
346 TimeSeries interp_time

### LINKED ISSUE #346: TimeSeries.interp_time should return a new TimeSeries
Currently, `TimeSeries.interp_time` returns an `xarray.DataArray`. I think it does make more sense to return a new `TimeSeries` instead. If the user really wants to get the underlying xarray object, he can fetch it after the interpolation.

vhirtham:
## Changes

`TimeSeries.interp_time` now returns a new `TimeSeries` instead of an `xarray.DataArray`. Additionally, a warning is emitted when already interpolated data is interpolated another time.

## Related Issues

Closes #346 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [ ] save interpolation counter to asdf?


vhirtham:
create new issue

weldx/core.py
```diff
@@ -470,55 +534,28 @@ def interp_time(
 
         Returns
         -------
-        xarray.DataArray:
+        TimeSeries :
             A data array containing the interpolated data.
```
CagtayFabry, line 538:
```suggestion
            A new `TimeSeries` object containing the interpolated data.
```

weldx/core.py
```diff
@@ -470,55 +534,28 @@ def interp_time(
...
+            dax = self._interp_time_discrete(time)
+        else:
+            dax = self._interp_time_expression(time, time_unit)
 
-        # create data array
-        if not np.iterable(data):  # make sure quantity is not scalar value
-            data = np.expand_dims(data, 0)
+        interpolation = self.interpolation
+        if interpolation is None:
+            interpolation = "linear"
```
CagtayFabry, line 554: I don't think we have to check for None here any longer
maybe we should add a setter for `self.interpolation` that also checks for the valid interpolation styles?

weldx/core.py
```diff
@@ -385,6 +387,68 @@ def __repr__(self):
...
+            if self._data.attrs["interpolation"] == "linear" or self.shape[0] == 1:
+                return ut.xr_interp_like(
+                    self._data,
+                    {"time": time},
+                    assume_sorted=False,
+                    broadcast_missing=False,
+                )
+
+            dax = self._data.reindex({"time": time}, method="ffill")
+            return dax.fillna(self._data[0])
```
CagtayFabry, line 415: since the outside branch was the "step" interpolation anyways (I think), can we simplify this to just use `util.xr_interp_like` now for all interpolation methods?

and if we do that it should also allow us to enable all xarray supported interpolation methods as well?

CagtayFabry (changes requested):

CagtayFabry:
what do you think about adding a `.time` property to `TimeSeries` directly @vhirtham ? In line with LCS and similar classes

vhirtham:
> what do you think about adding a `.time` property to `TimeSeries` directly @vhirtham ? In line with LCS and similar classes

Not sure what you intend, cause the `TimeSeries` already has a `time` property (line 501)

weldx/core.py
vhirtham: Well, it is needed in the current version because TS based on mathematical expressions have no interpolation and creating a new TS with `None` as interpolation raises an exception. However, I think the problem lies in the init method. I will rewrite it.

A setter for the interpolation is certainly useful. Will add it.

CagtayFabry:
> 
> 
> > what do you think about adding a `.time` property to `TimeSeries` directly @vhirtham ? In line with LCS and similar classes
> 
> Not sure what you intend, cause the `TimeSeries` already has a `time` property (line 501)

exactly that, so disregard :wink:
just wondering why we now need to call `data_array.time` here https://github.com/vhirtham/weldx/blob/3b06b74a800d2ac45b9794dfd81d49670d63f4e9/weldx/tests/test_core.py#L416-L419

vhirtham:
> > > what do you think about adding a `.time` property to `TimeSeries` directly @vhirtham ? In line with LCS and similar classes
> > 
> > 
> > Not sure what you intend, cause the `TimeSeries` already has a `time` property (line 501)
> 
> exactly that, so disregard 😉
> just wondering why we now need to call `data_array.time` here https://github.com/vhirtham/weldx/blob/3b06b74a800d2ac45b9794dfd81d49670d63f4e9/weldx/tests/test_core.py#L416-L419

The reason was that `.time` returns `None` if the internal time data has only one value, e.g the TS is constant. But we checked is the stored value matches the passed one. Don't know if that made sense, but I modified the test in favor of using `.time`

weldx/core.py
vhirtham: I added the other interpolation methods manually but I think it would be nicer to extract them somehow from xarray. A quick search didn't bring up a list that I could use to update ours. Didn't search very hard though. If you know where to find one, let me know and I will add it.

CagtayFabry (approved):
LGTM 👍

weldx/core.py
CagtayFabry: I couldn't find a list from xarray (also didn't look too hard) but I think it's fine now this way :)

weldx/core.py
CagtayFabry: ah yes I forgot about the math expression, makes sense
looks good this way imo

PR #358, opened 2021-05-18
[rtd] fix install of packages (allow pinning)

marscher:
## Changes

On readthedocs we first create a conda environment containing the dependencies to build the documentation (sphinx etc),
then we install weldx via pip. The settings of the pip invocation are set to upgrade all previously installed packages, which
prevents pinning package versions. This ultimately leads to version incompatibilities (e.g. nbpshinx needs jinja2 < 3.0, but the conda installed version 2.11 gets upgraded to 3 again in the pip install call).

To prevent this, we do not install weldx via RTD directly, but via the pip section in doc/rtd_environment.yml

### DIFF
--- a/doc/rtd_environment.yml
+++ b/doc/rtd_environment.yml
@@ -23,3 +23,4 @@ dependencies:
   - pip
   - pip:
     - git+https://github.com/CagtayFabry/sphinx-asdf.git@sphinx-weldx
+    - -e ../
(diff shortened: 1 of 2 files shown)


marscher:
firstly the correct jinja version is being installed via conda, then it gets upgraded when pip installs sphinx-asdf. I'll create a conda package for sphinx-asdf to circumvent this hassle.

doc/rtd_environment.yml
```diff
@@ -23,3 +23,4 @@ dependencies:
   - pip
   - pip:
     - git+https://github.com/CagtayFabry/sphinx-asdf.git@sphinx-weldx
+    - -e ../
```
CagtayFabry, line 26: this is amazing 🏅

Issue #110, opened 2020-08-21
unwanted name attribute from asdf NDArrayType


CagtayFabry:
There appears to happen some unlucky attribute propagation when reading from asdf `ndarray` -> `pint.Quantity` -> `xarray.DataArray` which makes the following fail:

~~~python
ts = TimeSeries(data=Q_([40.0,20.0], "V"),time=Q_([0.0,10.0],"s"))
data = weldx.asdf.utils._write_read_buffer({"root":ts})
data["root"] == ts
>>> False
~~~
Both `TimeSeries` have identical values but `==` fails because `xarray.DataArray.identical` checks every attribute including `DataArray.name`

When reading from the ASDF file in this case the name attribute gets erroneously applied:
~~~python
print(ts.data_array.name)
>>> None
print(data["root"].data_array.name)
>>> core/ndarray
~~~

The problem occurs because when creating a `pint.Quantity` directly from an ndarray it actually gets constructed from the asdf `NDArrayType` which apparently carries `.name='core/ndarray'`
This propagates over to the `pint.Quantity` and finally `xarray.DataArray`.

Here is a quick fix for the `TimeSeries` class https://github.com/BAMWelDX/weldx/pull/104/commits/1dd1f23ef76c59cbc8c198d00bdf97d049e3c3f7 but we should be aware of the issue and discuss how to handle this in general.

marscher:
Ain't an issue any more.

CagtayFabry:
I think we fixed this in the `TimeSeries` asdf implementation?
but the general problem should still persist.

One possible idea could be to adapt asdf implementation of our quantity to remove the attribute.

CagtayFabry:
cref https://github.com/BAMWelDX/weldx/pull/364#discussion_r655330492

vhirtham:
So reopen?

CagtayFabry:
yes I think I'd like to run some more tests until finally closing this one

marscher:
Sorry, if this caused extra work now. I just ran the attached snippet and it didn't showed the described behavior any more.

CagtayFabry:
it's good to keep in mind, I'll have to think of a proper snipped for debugging

marscher:
I guess this has something to do with attributes being discarded if set to None during serialization. A name of an xarray got set to None, to prevent the automatic derivation from the underlying array (NDArrayType of asdf has a name attribute), will be lost.

[...]

PR #364, opened 2021-06-08
Add spatial_data to csm serialization

### LINKED ISSUE #348: CSM SpatialData serialization
I didn't debug in detail but it appears the CSM serialization does not serialize `SpatialData` that is attached?

vhirtham:
## Changes

The data that is stored in the CSM is now also written to the asdf file. 

## Related Issues

Closes #348

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/asdf/tags/weldx/core/transformations/coordinate_system_hierarchy.py
```diff
@@ -458,4 +466,9 @@ def from_tree(cls, tree, ctx):
         cls._add_coordinate_systems_to_subsystems(tree, csm, subsystem_data_list)
         cls._merge_subsystems(tree, csm, subsystem_data_list)
 
+        spatial_data = tree.get("spatial_data")
+        if spatial_data is not None:
```
CagtayFabry, line 470:
```suggestion
        if (spatial_data := tree.get("spatial_data")) is not None:
```
py38 🥳

weldx/asdf/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
```diff
@@ -97,7 +97,20 @@ properties:
...
+    type: array
+    items:
+      type: object
+      properties:
+        coordinate_system:
+          type: string
+        name:
+          type: string
+        data:
+          tag: "tag:weldx.bam.de:weldx/core/geometry/spatial_data-1.0.0"
```
CagtayFabry, line 110:
```suggestion
          tag: "tag:weldx.bam.de:weldx/core/geometry/spatial_data-1.*"
```

weldx/geometry.py
```diff
@@ -2330,6 +2330,16 @@ def __post_init__(self):
...
+    # def __eq__(self, other):
+    # """Compare to another object."""
+    # if not isinstance(other, SpatialData):
+    #    return False
+    # if not self.coordinates.data == other.coordinates.data:
+    #    return False
+    # if self.triangles and not self.triangles == other.triangles:
+    #    return False
+    # if not self.attributes ==
+
```
CagtayFabry, line 2342: I guess default `__eq__` doesn't work out of the box because of the arrays?

in that case we should implement a basic `__eq__` function, using `util.compare_nested` for the attributes?

weldx/tests/asdf_tests/test_asdf_core.py
```diff
@@ -403,13 +403,52 @@ def test_coordinate_system_manager_time_dependencies(
...
+        tree, open_kwargs={"copy_arrays": copy_arrays, "lazy_load": lazy_load}
+    )
+    csm_buffer = buffer["csm"]
+
+    for data_name in csm.data_names:
+        sd = csm.get_data(data_name)
+        sd_buffer = csm_buffer.get_data(data_name)
+        assert np.allclose(sd.coordinates.data, sd_buffer.coordinates.data)
+        if sd.triangles is not None:
+            assert np.allclose(np.array(sd.triangles), np.array(sd_buffer.triangles))
```
CagtayFabry, line 449: use `==` here once implemented

CagtayFabry (changes requested):
asdf implementation LGTM 👍 (should be even easier once switching to graph backend for serialization)

I think it's a good time to implement `__eq__` for `SpatialData`

weldx/geometry.py
vhirtham: I ll try it out

weldx/tests/asdf_tests/test_asdf_core.py
vhirtham: Doesn't compare equal:

~~~ 
SpatialData(coordinates=<xarray.DataArray 'core/ndarray' (n: 2, c: 3)>
vs.
SpatialData(coordinates=<xarray.DataArray (n: 2, c: 3)>
~~~

The data seems to be equal, but since compare nested is using xarrays `identical` function, we get a `False` returned.
I remember that we had trouble with the `identical` function before, since it also considers the order of dimensions, coordinates etc. Alternatively, we can use `equals` for a more relaxed comparison but I am not sure if this is really what we want.

CagtayFabry: I think this happens because of the `.name` property that get's created when reading arrays from asdf files (due to `NDArrayType ` wrapper)

see #110 and a possible workaround (reading explicitly as `np.asarray` from tree here https://github.com/BAMWelDX/weldx/commit/1dd1f23ef76c59cbc8c198d00bdf97d049e3c3f7)

maybe we should ask for a solution over at asdf

vhirtham: I'll try the workaround for now.

vhirtham: Works now. Think we can merge

CagtayFabry (approved):

Issue #86, opened 2020-07-24
implement custom shape validation for specific tags


CagtayFabry:
Some tags might not implement a `shape` property but should validate against `wx_shape`

We should find a nice way to implement this on a per class basis
For a hacky implementation see #85 

https://github.com/BAMWelDX/weldx/blob/28d3aedb68526faf192f400405cec9aa3f477eb4/weldx/asdf/validators.py#L438-L445

CagtayFabry:
New approach in #176 

implementation moved to `weld.validators._get_instance_shape`

CagtayFabry:
A few notes:

- We should probably define the interfaces in the `WeldxType` base class

- this part
  ```python
     elif isinstance(instance_dict, asdf.types.tagged.Tagged): 
         # add custom type implementations 
         if "weldx/time/timedeltaindex" in instance_dict._tag: 
             return TimedeltaIndexType.shape_from_tagged(instance_dict) 
         elif "weldx/time/datetimeindex" in instance_dict._tag: 
             return DatetimeIndexType.shape_from_tagged(instance_dict)
  ```
  should perform a lookup to find the type based on the tag and try to run `shape_from_tagged()`

- implementation should be based on ASDF 2.8 API

CagtayFabry:
I have not looked into the changes to validators with the new asdf 2.8 API changes but we should think about this

Issue #51, opened 2020-06-23
support callable coordinate/rotation objects


CagtayFabry:
Similar to the astropy implementation of [`DynamicMatrixTransform`](https://docs.astropy.org/en/stable/api/astropy.coordinates.DynamicMatrixTransform.html#astropy.coordinates.DynamicMatrixTransform) I think we should consider supporting callable functions for defining  translations and rotations as a future addition

CagtayFabry:
we should be closer to this with #76

vhirtham:
So just to get this sorted. We are aiming for some interfaces like this?

~~~ python
t = DynamicTranslation(x="2*t + 3", y="3", z="2*sin(t)")
m = t.interp_time(Q_([1, 2, 3], "s"))
~~~

where `m` is an instance of `LocalCoordinateSystem`?
Or for a rotation:

~~~ python
r = DynamicRotation(sequence = "xy", angles=["10*t", "5*t"], degrees=True) 
~~~

CagtayFabry:
ideally something like:
```python
# omitting `TimeSeries` args but of course there would have to be a check on dimensions etc.
lcs_0 = LCS(coordinates=TimeSeries(*args), orientation=TimeSeries(*args))
```
only interpolate when requested by CSM or otherwise.
Of course mix/match between discrete translation/rotation for LCS should also be viable

Creating something like `DynamicRotation` would also make sense when subclassing `TimeSeries` with additional restrictions (or just "easily/safely" create proper a `TimeSeries` )

vhirtham:
Not sure about the `TimeSeries` here. If we have dedicated objects with an overloaded call operator I think it is okay, because it is their purpose. However, the `TimeSeries` has multiple other functionalities, and making it callable would be somehow strange. Feels unintuitive. Maybe we can simply use a function of the same name that is implemented for all objects that should support this feature?

CagtayFabry:
my thought was that `TimeSeries` is the default object representing time dependent data so it would fit here
We only really need `interp_time` from the `TimeSeries` object for this

That should allow something like
```python
ts_sine = weldx.util.sine(f=Q_(0.5 * 2 * np.pi, "Hz"), amp=Q_([[0, 0.75, 0]], "mm"))
tcp_sine = lcs(coordinates=ts_sine)
```
instead of
```python
ts_sine = util.sine(f=Q_(0.5 * 2 * np.pi, "Hz"), amp=Q_([[0, 0.75, 0]], "mm"))
[...]

PR #366, opened 2021-06-10
Support TimeSeries as possible input to LCS.__init__ 

vhirtham:
## Changes

`TimeSeries` can be used as coordinates of a `LocalCoordinateSystem`

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [x] ~~update example/tutorial notebooks~~ Combine with update of orientations
- [x] asdf serialization
- [x] copy date from time series with discrete inputs to coordinates
- [x] ~~deprecate `util.lcs_coords_from_ts`~~ created an issue
- [x] ~~fix todo in time_ref (see discussion below)~~ created an issue


vhirtham:
I am currently struggling with the `invert` method. This is the current implementation

~~~ python
    def invert(self) -> "LocalCoordinateSystem":
                orientation = ut.xr_transpose_matrix_data(self.orientation, dim1="c", dim2="v")
        coordinates = ut.xr_matmul(
            self.orientation,
            -self.coordinates,
            dims_a=["c", "v"],
            dims_b=["c"],
            trans_a=True,
        )
        return LocalCoordinateSystem(
            orientation, coordinates, self.time, self.reference_time
        )
~~~

The problem here is, that the coordinates get multiplied with a matrix. Since the coordinates are a TimeSeries now, the current code obviously doesn't work. So we need to approach it a bit differently if a TimeSeries is used. The first idea would be to create a new TimeSeries with updated formula. Something like this (pseudo-code-math-mixture):

~~~
new_expression = M * (-1 * old_expression)
new_ts = TimeSeries(new_expression, parameters=dict(M=mat, **old_parameters))
~~~

I tested the matrix implementation of sympy and this should work. However, In case the orientation is time-dependent, the expression will return a list of vectors instead of a single one for each time input, which isn't what we want and will probably crash all of our other code. The workaround for time-dependent systems would be, to turn the equation into explicit values before inverting. However, this will remove the equation and cause deviations when further interpolating the result. For static systems, the expression can be preserved using the above technique. 

Alternatively, we can store in the LCS that it has been inverted and all necessary information to perform the actual value interpolation later. This will allow us to evaluate the equation for each required time_step and invert the coordinates afterward. Therefore, we can keep the expression. The problem here is, that we probably have to check if the system was inverted at multiple places in the code. I can't really tell yet if this gets messy or if we are talking about less than 10 new lines. But it certainly makes the already complicated code even more difficult to understand. We can also decide to take this route only for time-dependent systems and use the previous approach for static systems.

Lastly, we can simply forbid inversions of LCS with TimeSeries as coordinates. I have to check how frequently we use the `invert` function. So far I have only seen it in the `__sub__` method of the LCS and we can rewrite it in a way that it is not affected by this restriction. However, I have to check the CSM if that would cause trouble there, but I guess it shouldn't. Maybe we can talk about this on Friday. 

Still have some other ideas that I haven't thought through. If I can come up with another solution, I'll let you know.

CagtayFabry:
regarding the inversion problem:
I can see how this makes the code/math really complicated.

From the way I see it I think we should take a careful look on when `invert` is actually called. It is not called when simply constructing a `LCS` right? Does the `CSM` automatically generate the inverted instances? (I think but have to check again in detail)

How about this workaround:
Since in my eyes the support for `TimeSeries` is more of an intermediate step and should not really be used as a tool to do 'actual' math operations (we will leave that to discrete values) we do not support direct `LCS` inversion for `TimeSeries` (raise an exception).
Instead the code/user should ensure to only call `invert` on 'discretized' `LCS` instances, for example by calling `interp_like` first. Implementing this in the `CSM` should not be too complicated for us I think.
Of course we could also add an optional `time` keyword argument providing time steps to `invert` for simplicity.

weldx/asdf/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-1.0.0.yaml
```diff
@@ -105,15 +105,18 @@ properties:
...
   orientations:
     description: |
       An ndarray containing the orientations.
     tag: "tag:weldx.bam.de:weldx/core/variable-1.0.0"
 
 wx_shape:
-  (coordinates):
-    data: [(t), 3]
+  # (coordinates):
+    # data: [(t), 3]
```
vhirtham, line 119: We need to decide if and how we want to use shape validation if the coordinates are a `TimeSeries`

CagtayFabry: cref #86

weldx/core.py
```diff
@@ -132,7 +132,11 @@ def equals(
                 equality = simplify(self.expression - other.expression) == 0
 
             if check_parameters:
-                equality = equality and self._parameters == other.parameters
+                from weldx.util import compare_nested
```
CagtayFabry, line 135: do we have some ring dependencies on top level import here?

vhirtham: Can't tell for this particular import, but I had some ring dependency issues. Maybe we should start transferring the code to have one file per class like in the transformations package.

CagtayFabry:
can you update the changes section of the PR to reflect the current implementation level? @vhirtham 

- `TimeSeries` works as coordinates?
- CSM can handle/add `TimeSeries` LCS?
- CSM interpolation works (with the right inputs)?

Can we add `TimeSeries` orientations in a simple form as well?

vhirtham:
> can you update the changes section of the PR to reflect the current implementation level? @vhirtham
> 
> * `TimeSeries` works as coordinates?
> * CSM can handle/add `TimeSeries` LCS?
> * CSM interpolation works (with the right inputs)?

Will do that, once the PR is ready. It is still WIP. I have to check the CSM if there is anything missing. Pretty sure it isn't, but I haven't double-checked it so far.

> Can we add `TimeSeries` orientations in a simple form as well?

As stated in #51 , I would use a new branch after this one is merged.

weldx/core.py
CagtayFabry: unless it is indeed necessary here I would prefer to have these kind of imports at the file scope for consistency

vhirtham: After a short test, this causes ring dependency issues.

weldx/transformations/local_cs.py
```diff
@@ -58,7 +58,9 @@ def __init__(
             provided as orientation.
             Passing a scipy.spatial.transform.Rotation object is also supported.
         coordinates :
-            Coordinates of the origin
+            Coordinates of the origin. If the passed type is a `TimeSeries` with
+            discrete values and the ``time`` parameter is set, it will be interpolated
+            the values of ``time``.
```
vhirtham, line 63: We need to decide which behavior we want to have if the TimeSeries does not match the provided time (needed for time dep. orientations). Currently, the data of the time series is just interpolated, which isn't such a good solution since the LCS might be interpolated too later resulting in double interpolation. Alternatively, we can raise an exception during construction, which is probably better.

vhirtham:
Lest remaining issue is the shape during validation. Then this PR should be ready.

weldx/transformations/local_cs.py
CagtayFabry: Do we even want to allow `TimeSeries` with discrete values?

In that case it is basically just another xarray input so we should just convert (extract xarray) and use the original constructor behaviour?

vhirtham: That's exactly what happens internally. The data from the `TimeSeries` is copied. Still, the problem with time is remaining.

CagtayFabry: from looking into the code it is now basically the same behavior as before with two different xarray structures, right? (unifying time axis)
So I think just keep it that way

weldx/transformations/local_cs.py
```diff
@@ -303,6 +305,26 @@ def _build_coordinates(coordinates, time: pd.DatetimeIndex = None):
...
+                    f"{coordinates.shape}"
+                )
+            coordinates = coordinates.data_array
+            coordinates = coordinates.rename({coordinates.dims[1]: "c"})
+            coordinates = coordinates.assign_coords(dict(c=["x", "y", "z"]))
+            # This is a workaround to remove the warning about stripped units. This line
+            # should be removed once we add/require units in the lcs
+            # Additionally, the correct unit should be checked for TimeSeries
+            # (also expressions)
+            coordinates.data = coordinates.data.m
```
vhirtham, line 323: See comment

weldx/transformations/local_cs.py
vhirtham: Fixed as discussed offline

weldx/transformations/local_cs.py
CagtayFabry: I think we have some code that uses units with LCS coords

for now I think the most consistent way in these cases would be, to convert to our 'default' unit of `"mm"`?
This would also raise an exception if units where not of type `"[length]"`

vhirtham: It "somehow" works with units, but it isn't inherently supported. Converting to `mm` first is a good suggestion

weldx/core.py
CagtayFabry: we should take a look into this @marscher @vhirtham soon

CagtayFabry (changes requested):
I think it would be convenient to add a property like `has_expression` to `LCS` that can later also cover the orientation

CagtayFabry:
There seems to be an issue with plotting the CSM graph when a `TimeSeries` is present, have a look @vhirtham 
```python
import numpy as np
import weldx
from weldx import LocalCoordinateSystem as lcs
from weldx import Q_
from weldx import util

csm = weldx.CoordinateSystemManager("base")
ts_sine = util.sine(f=Q_(0.5 * 2 * np.pi, "Hz"), amp=Q_([[0, 0.75, 0]], "mm"))
csm.add_cs(
    coordinate_system_name="sine",
    reference_system_name="base",
    lcs=lcs(coordinates=ts_sine),
)
csm.plot_graph()
>>> Exception: The chosen transformation is time dependent, but no time is given. This is usually the case if the time dependencies are only described by mathematical expressions. Provide the desired time using the corresponding parameter to solve this issue.
```

weldx/transformations/local_cs.py
```diff
@@ -303,6 +306,29 @@ def _build_coordinates(coordinates, time: pd.DatetimeIndex = None):
...
+            # This is a workaround to remove the warning about stripped units. This line
+            # should be removed once we add/require units in the lcs
+            # Additionally, the correct unit should be checked for TimeSeries
+            # (also expressions)
+            coordinates.data = coordinates.data.to(UREG.mm).m
+            if coordinates.data.shape[0] == 1:
+                coordinates = coordinates.data.reshape(3)
+            else:
+                coordinates = coordinates.rename({coordinates.dims[1]: "c"})
+                coordinates = coordinates.assign_coords(dict(c=["x", "y", "z"]))
```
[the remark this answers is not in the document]
vhirtham: I will put it into a static method. Also, I would deprecate `utils.lcs_coords_from_ts`, because all its functionality (except for the interpolation) is now handled by the LCS

vhirtham:
> There seems to be an issue with plotting the CSM graph when a `TimeSeries` is present, have a look @vhirtham
> 
> ```python
> import numpy as np
> import weldx
> from weldx import LocalCoordinateSystem, CoordinateSystemManager, Q_, util
> 
> csm = weldx.CoordinateSystemManager("base")
> ts_sine = util.sine(f=Q_(0.5 * 2 * np.pi, "Hz"), amp=Q_([[0, 0.75, 0]], "mm"))
> csm.add_cs(
>     coordinate_system_name="sine",
>     reference_system_name="base",
>     lcs=LocalCoordinateSystem(coordinates=ts_sine),
> )
> csm.plot_graph()
> >>> Exception: The chosen transformation is time dependent, but no time is given. This is usually the case if the time dependencies are only described by mathematical expressions. Provide the desired time using the corresponding parameter to solve this issue.
> ```

This was caused by calling `get_cs` internally. We needed that to call `is_time_dependent` on the lcs to distinct time-dependent and static systems. However, the fix probably made the function a lot faster since `get_cs` performs a lot of checks that are not needed here

weldx/transformations/local_cs.py
```diff
@@ -799,9 +907,23 @@ def interp_time(
...
+            if isinstance(time_interp, pd.DatetimeIndex):
+                time_interp = time - self.reference_time
 
-        return LocalCoordinateSystem(orientation, coordinates, time_ref=time_ref)
+            coordinates = self._coords_from_discrete_time_series(
+                self.coordinates.interp_time(time_interp)
+            )
+
+            if self.has_reference_time:
+                coordinates.weldx.time_ref = self.reference_time
```
vhirtham, line 920: The condition wasn't necessary before using the weldx accessor. The problem is, that it doesn't support `None`. There is a to-do in the implementation of the accessor addressing this:

~~~ python
@time_ref.setter
    def time_ref(self, value: pd.Timestamp):
        """Convert INPLACE to new reference time.

        If no reference time exists, the new value will be assigned
        TODO: should None be allowed and pass through or raise TypeError ?
        """
        if "time" in self._obj.coords:
            value = _as_valid_timestamp(value)
            if self._obj.weldx.time_ref and is_timedelta64_dtype(self._obj.time):
                if value == self._obj.weldx.time_ref:
                    return
                _attrs = self._obj.time.attrs
                time_delta = value - self._obj.weldx.time_ref
                self._obj["time"] = self._obj.time.data - time_delta
                self._obj.time.attrs = _attrs  # restore old attributes !
                self._obj.time.attrs["time_ref"] = value  # set new time_ref value
            else:
                self._obj.time.attrs["time_ref"] = value
~~~

Do we want to make a decision and clear this to-do?

CagtayFabry: what was the previous behavior passing `None`?

vhirtham: Crash. I believe it was the timestamp conversion that failed.

CagtayFabry: what do you think default behavior should be when passing `None`?

remove an existing reference time (how else would we do that) or just do nothing?

vhirtham: Initially, I added 

~~~ python
if value is None:
    return
~~~

but I wasn't sure because this caused a Failure in the exception tests of the `reset_reference_time` method. Currently, we expect it to fail if None is passed. Would be conservative about it and make raising an exception the default behavior.

vhirtham:
Trying to figure out why the k3d plot fails. Any clue?

weldx/transformations/cs_manager.py
```diff
@@ -1955,7 +1955,7 @@ def _get_lcs(edge):
                 return self.graph.edges[(edge[1], edge[0])]["lcs"]
 
             lcs_list = [_get_lcs(edge) for edge in list_of_edges]
-            lcs_list = [lcs for lcs in lcs_list if lcs.time is not None]
+        lcs_list = [lcs for lcs in lcs_list if lcs.time is not None]
```
vhirtham, line 1958: So this was the line that caused the trouble. It was only included in the branch that was executed if one specified a list of edges. So for the default case, an LCS with a TimeSeries wouldn't be sorted out and cause a `None` value in the time union. It's funny, that the less used case (providing a list of edges) was actually indirectly covered, while the default case wasn't. So it slipped through except for the notebook. I now added a set of dedicated tests 😉

vhirtham:
So everything is more or less ready. I added 2 bullet points into the checklist at the top, but those are more or less unrelated issues and we can fix them in another branch. Just wanted to have a reminder and it is a bit late now to write a full issue text 😄

weldx/transformations/cs_manager.py
```diff
@@ -1377,6 +1393,19 @@ def get_cs(
...
+
+            if invert:
+                if isinstance(lcs.coordinates, TimeSeries):
+                    raise Exception(
+                        "The chosen transformation is time dependent, but no time is "
+                        "given. This is usually the case if the time dependencies are "
+                        "only described by mathematical expressions. Provide the "
+                        "desired time using the corresponding parameter to solve this "
+                        "issue."
+                    )
```
CagtayFabry, line 1405: I think we can skip the test here since `lcs.invert()` will basically raise the same exception.

CagtayFabry:
> I think it would be convenient to add a property like `has_expression` to `LCS` that can later also cover the orientation

can we add this and use it in places with `isinstance(lcs.coordinates, TimeSeries):` @vhirtham

weldx/transformations/cs_manager.py
vhirtham: Just thought that the exception raised by the lcs might be a bit cryptic for a CSM user without deeper knowledge of the underlying mechanisms. Does it hurt here?

vhirtham:
> > I think it would be convenient to add a property like `has_expression` to `LCS` that can later also cover the orientation
> 
> can we add this and use it in places with `isinstance(lcs.coordinates, TimeSeries):` @vhirtham

"Yes, we can!" - but I am not sure if we shouldn't postpone it to the implementation of `TimeSeries` based orientations because this will change some aspects and you need to differentiate between the cases pure expression based time dependency and mixed time dependency with discrete values and expressions.

CagtayFabry:
> > > I think it would be convenient to add a property like `has_expression` to `LCS` that can later also cover the orientation
> > 
> > 
> > can we add this and use it in places with `isinstance(lcs.coordinates, TimeSeries):` @vhirtham
> 
> "Yes, we can!" - but I am not sure if we shouldn't postpone it to the implementation of `TimeSeries` based orientations because this will change some aspects and you need to differentiate between the cases pure expression based time dependency and mixed time dependency with discrete values and expressions.

ok then we lave it like this for now

weldx/transformations/cs_manager.py
CagtayFabry: I would like to keep the code and branches as short as possible, we will have to change `isinstance(lcs.coordinates, TimeSeries):` here again later anyway for example

then see what really is confusing or an issue later
but let's just leave it in here for now

CagtayFabry (approved):
Some of the plots in the notebooks look a bit off (the plot limits don't seem to work for one plot and funny colors in one CSM graph)

but we can refine these later, looks good 👍

vhirtham:
> Some of the plots in the notebooks look a bit off (the plot limits don't seem to work for one plot and funny colors in one CSM graph)
> 
> but we can refine these later, looks good 👍

Can you gather the corresponding plots in an issue? I will have look as soon as 0.4.0 is released

CagtayFabry:
> > Some of the plots in the notebooks look a bit off (the plot limits don't seem to work for one plot and funny colors in one CSM graph)
> > but we can refine these later, looks good 👍
> 
> Can you gather the corresponding plots in an issue? I will have look as soon as 0.4.0 is released

we can look into it when reviewing the tutorials
would be nice to have a unique color for TimeSeries LCS instances in the CSM plot as well

PR #377, opened 2021-06-21
Compare nested check keys

### LINKED ISSUE #376: Compare_nested doesn't take additional dict items of RHS into account
~~~ python
from weldx.util import compare_nested

d1 = dict(a=1)
d2 = dict(a=1, b=1)

print(compare_nested(d1, d2))
print(compare_nested(d2, d1))
~~~

returns:

~~~
True
False
~~~

So the additional member `b` of `d2` is not considered when `d2` is the rhs operand.

CagtayFabry:
## Changes
check if keys in mappings are equal with `util.compare_nested`

## Related Issues
Closes #376

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/


vhirtham (approved):
Thanks for the fast fix 👍

PR #380, opened 2021-06-22
Simplify dataclass serialization

vhirtham:
## Changes

Adds a new utility function that automatically generates the asdf serialization class for python dataclasses. 

## Checks

- [x] updated CHANGELOG.md
- [x] ~~updated tests~~
- [x] ~~updated doc/~~

### DIFF
--- a/weldx/asdf/util.py
+++ b/weldx/asdf/util.py
@@ -1,5 +1,4 @@
 """Utilities for asdf files."""
-from copy import deepcopy
 from io import BytesIO
 from pathlib import Path
 from typing import Callable, Tuple, Type, Union
@@ -341,7 +340,7 @@ def dataclass_serialization_class(
 
         @classmethod
         def to_tree(cls, node, ctx):
-            return to_tree_mod(deepcopy(node.__dict__))
+            return to_tree_mod(node.__dict__)
 
         @classmethod
         def from_tree(cls, tree, ctx):
--- a/weldx/asdf/util.py
+++ b/weldx/asdf/util.py
@@ -288,3 +288,61 @@ def view_tree(file: types_path_and_file_like, path: Tuple = None, **kwargs):
 def asdf_json_repr(file: Union[str, Path, BytesIO], path: Tuple = None, **kwargs):
     """See `view_tree` function."""
     return view_tree(file, path, **kwargs)
+
+
+def asdf_dataclass_serialization(original_class):
+    """Add :code:`to_tree` and :code:`from_tree` methods for dataclass serialization.
+
+    Parameters
+    ----------
+    original_class:
+        Original data class to decorate
+
+    Returns
+    -------
+    type :
+        The class with added :code:`to_tree` and `from_tree` function.
+
+    Examples
+    --------
+    Here is an example how to generate a simple serialization class for a given
+    dataclass::
+
+        @dataclass
+        class MyClass:
+            a: np.ndarray
+
+        @asdf_dataclass_serialization
+        class MyClassTypeASDF(WeldxType):
+            name = "my_class"
+            version = "1.0.0"
+            types = [MyClass]
+            requires = ["weldx"]
+            handle_dynamic_subclasses = True
+
+    """
+    from copy import deepcopy
+
+    from weldx.asdf.types import from_tree_metadata, to_tree_metadata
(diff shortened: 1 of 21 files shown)


weldx/asdf/util.py
```diff
@@ -288,3 +288,69 @@ def view_tree(file: types_path_and_file_like, path: Tuple = None, **kwargs):
...
+    class _SerializationClass(WeldxType):
+        name = class_name
+        version = "1.0.0"
+        types = [class_type]
+        requires = ["weldx"]
+        handle_dynamic_subclasses = True
+
+        @classmethod
+        def to_tree(cls, node, ctx):
+            return to_tree_mod(deepcopy(node.__dict__))
```
CagtayFabry, line 346: I don't think `deepcopy` should be the default, would that not break links/refs in asdf files?

weldx/asdf/util.py
```diff
@@ -288,3 +288,69 @@ def view_tree(file: types_path_and_file_like, path: Tuple = None, **kwargs):
...
+    -------
+    Type :
+        A new asdf serialization class.
+
+    """
+    from copy import deepcopy
+
+    import numpy as np
+
+    from weldx.asdf.types import WeldxType
```
CagtayFabry, line 327: any reason these have to be imported here instead of at the top of the file?

weldx/asdf/util.py
```diff
@@ -288,3 +288,69 @@ def view_tree(file: types_path_and_file_like, path: Tuple = None, **kwargs):
...
+
+        @classmethod
+        def to_tree(cls, node, ctx):
+            return to_tree_mod(deepcopy(node.__dict__))
+
+        @classmethod
+        def from_tree(cls, tree, ctx):
+            for k, v in tree.items():
+                if isinstance(v, np.ndarray):
+                    tree[k] = np.asarray(v)
```
CagtayFabry, line 352: I like the idea catching errors here
this probably 'breaks' lazy loading or memory mapping of arrays though, so we should keep that in mind (and decide if we really want that as default behaviour)

CagtayFabry:
good idea! 🚀

vhirtham:
Initially, I used a decorator (first commit), but I find the solution using a factory function to be much more flexible, easier to implement, and yielding fewer lines of code per dataclass.

weldx/asdf/util.py
vhirtham: The problem here is that if you modify the dict, you also modify the dataclass. Due to `to_tree` being called twice per write, this causes errors whenever you need to modify the dict before passing it to asdf. However, we can probably remove it as default and just add the copy whenever we provide a modification function.

weldx/asdf/util.py
vhirtham: Not really, think we can move it to the top

weldx/asdf/util.py
vhirtham: If we can get the asdf maintainers to fix the NumPy name issue, we can remove it. For now, I would just leave it in place and reevaluate it later, as you said.

vhirtham: Update: I removed it because it didn't do what it was supposed to do (see failed tests). The reason was probably that I checked for a `numpy.ndarray` while I should have checked for the asdf `NDArrayType`. But since `SpatialData` is currently the only class in need of this workaround, I just fixed it for this one.

weldx/asdf/util.py
```diff
@@ -288,3 +289,60 @@ def view_tree(file: types_path_and_file_like, path: Tuple = None, **kwargs):
...
+    Parameters
+    ----------
+    class_type :
+        The type of the dataclass
+    class_name :
+        The value that should ba stored as the classes name property
+    version :
+        The version number
+    to_tree_mod :
+        A method that applies additional modifications to the tree during the `to_tree`
```
CagtayFabry, line 312:
```suggestion
        A method that applies additional modifications to the tree during the ``to_tree``
```
I think this or
```
:code:`to_tree`
```
should get rid of the sphinx link warnings :)

CagtayFabry (approved):
LGTM

vhirtham:
RTD build passed but doesn't seem to report back. I ll just merge this PR

PR #384, opened 2021-06-29
[asdf/xarray.DataSet|DataArray] fix serialization of attributes

marscher:
## Changes

The wrapping of `Variables` inside the `DataSet` serialization did not contain the `attrs` property. Additionally, the `coordinates`
of a `DataArray` can also enclose attributes, these were also not being serialized.

Refactored some for loops in favor of list comprehensions.

## Related Issues

Closes #349

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/
- [ ] update example/tutorial notebooks

PR #390, opened 2021-07-05
Update variable schema with attrs field

CagtayFabry:
## Changes
- update schema to include `attrs` 
- `attrs` field no longer gets serialized when empty

## Related Issues
see #384

### DIFF
--- a/weldx/asdf/schemas/weldx.bam.de/weldx/core/variable-1.0.0.yaml
+++ b/weldx/asdf/schemas/weldx.bam.de/weldx/core/variable-1.0.0.yaml
@@ -34,6 +34,10 @@ properties:
     description: |
       Optional field describing the unit of the data as valid string representation.
     type: string
+  attrs:
+    description: |
+      Additional attributes metadata associated with the variable.
+    type: object
 
 required: [name, dimensions, dtype, data]
 propertyOrder: [name, dimensions, dtype, unit, data]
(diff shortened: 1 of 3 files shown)


weldx/asdf/schemas/weldx.bam.de/weldx/core/variable-1.0.0.yaml
```diff
@@ -34,6 +34,10 @@ properties:
     description: |
       Optional field describing the unit of the data as valid string representation.
     type: string
+  attrs:
+    description: |
+      Additional attributes metadata associated with the variable.
```
vhirtham, line 39:
```suggestion
      Additional attributes/metadata associated with the variable.
```

vhirtham (approved):

CagtayFabry:
just noticed that the `attrs` field gets serialized when empty as `attrs: {}`

Will add a fix to correctly set `None` (and auto-remove) in that case later

CagtayFabry:
similar to how we auto-remove `None` we could also handle other 'empty' fields such as empty dicts, might be worth considering for the future

marscher:
Just out of curiosity, whats is the benefit of removing "empty" properties, except for saving a few bytes of metadata?

CagtayFabry:
I think it might be more distinct/intuitive in combination with the quality standards.
With this implementation, simply adding any field to the `required` list of a schema will also ensure that there is actual data stored in the field. But I'm not 100% sure if this would not also get triggered by empty/None fields (this was also the default behavior for asdf up until 2.6 or so)

PR #402, opened 2021-07-07
allow str quantities for MathExpr

### LINKED ISSUE #387: Support for string quantities
Pint supports strings as a method to create a new Quantity. Functions that use the ureg decorator already convert strings correctly into a quantity, but there are several functions in our packages that do not support it. I think it is a nice feature that makes user code much more readable and is easier to use. So I suggest, that all public interfaces support strings as input method.

CagtayFabry:
## Changes
allow passing `str` parameters to Mathexpression

## Related Issues
Closes #387 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@
 - Add `MeasurementChain.output_signal` property that returns the output signal of the `MeasurementChain`
   [[#394]](https://github.com/BAMWelDX/weldx/pull/394)
 
-
 ### changes
 
 - `WXRotation.from_euler()` now accepts a `pint.Quantity` as input. [[#318]](https://github.com/BAMWelDX/weldx/pull/318)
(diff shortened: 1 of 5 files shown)


CagtayFabry:
I quickly added str support in a few places (MathExpression), a more general solution would probably be more involved but this should be fine for now I think @vhirtham

vhirtham:
> I quickly added str support in a few places (MathExpression), a more general solution would probably be more involved but this should be fine for now I think @vhirtham

Okay, you got me a bit confused. You do not mean `MathematicalExpression`, do you? Because I see no changes to it in the diff. Only to the utility functions `sine` and `compute_welding_speed`.

CagtayFabry:
> > I quickly added str support in a few places (MathExpression), a more general solution would probably be more involved but this should be fine for now I think @vhirtham
> 
> Okay, you got me a bit confused. You do not mean `MathematicalExpression`, do you? Because I see no changes to it in the diff. Only to the utility functions `sine` and `compute_welding_speed`.

for some reason I forgot to commit the one thing I actually wanted to fix ;)
should work now

vhirtham (approved):

vhirtham:
Watch out that the merge will close the issue. Not sure if we want that yet.

CagtayFabry:
> Watch out that the merge will close the issue. Not sure if we want that yet.

maybe list a few other functions that should support this behavior in that issue?
We can reopen it afterwards again

PR #405, opened 2021-07-08
add test path to pytest config

CagtayFabry:
## Changes
set the pytest path in setup.cfg

### DIFF
--- a/.github/workflows/pytest_asdf.yml
+++ b/.github/workflows/pytest_asdf.yml
@@ -17,6 +17,6 @@ jobs:
       - name: pip installs
         run: |
           pip install -e .[test]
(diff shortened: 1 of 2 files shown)


CagtayFabry:
apparently the `welding_schema.py.py` script no longer gets executed as script, do you remember which test case/file coveres that @marscher ?

CagtayFabry:
i ended up also specifying the path for the separate asdf schema tests since pytest kept picking up multiple other files otherwise, see https://towardsdatascience.com/build-a-stunning-readme-for-your-github-profile-9b80434fe5d7

I think this is now the intended behaviour?

vhirtham (approved):

PR #420, opened 2021-07-13
Timeseries plot function

### LINKED ISSUE #393: TimeSeries plot function
Just realized that our `TimeSeries` has no plot function and I think it should add it. Most of the code should already be there in the helper functions file of the reference data notebook.

vhirtham:
## Changes

Add a plot function to the `TimeSeries`

## Related Issues

Closes #393 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [x] add `Signal.plot`


weldx/core.py
```diff
@@ -601,6 +602,52 @@ def interp_time(
...
+            return self.interp_time(time).plot(
+                axes=axes, title=title, data_name=data_name, **mpl_kwargs
+            )
+
+        axes.plot(self.time.values / 1e9, self._data.data.m, **mpl_kwargs)
+        axes.set_title(title)
+        axes.set_xlabel("t in s")
+        y_unit_label = ""
+        if self.units in ["", "dimensionless"]:
+            y_unit_label = f"in {self.units}"
```
CagtayFabry, line 647:
```suggestion
            y_unit_label = f"in {self.units:~}"
```
pint short notation formatting

weldx/core.py
```diff
@@ -601,6 +602,52 @@ def interp_time(
...
+
+        """
+        if axes is None:
+            _, axes = plt.subplots()
+        if self.is_expression or time is not None:
+            return self.interp_time(time).plot(
+                axes=axes, title=title, data_name=data_name, **mpl_kwargs
+            )
+
+        axes.plot(self.time.values / 1e9, self._data.data.m, **mpl_kwargs)
```
CagtayFabry, line 642: can we make the time conversion more safe/explicit and unit-based? Maybe so that the user can provide a base time unit/scale to convert to?

vhirtham: Sure. Maybe it is better to convert to pint units first. Was just dealing with the fact that `.time` returns a `TimedeltaIndex`. Would also be beneficial for the labeling of the x-axis.

weldx/core.py
```diff
@@ -624,7 +671,7 @@ def units(self) -> str:
         Returns
         -------
         str:
-            Unit sting
+            Unit string
```
CagtayFabry, line 674: This functions seems to return a `pint.Unit`? That is probably even better than working with strings

weldx/core.py
```diff
@@ -601,6 +602,52 @@ def interp_time(
...
+        """
+        if axes is None:
+            _, axes = plt.subplots()
+        if self.is_expression or time is not None:
+            return self.interp_time(time).plot(
+                axes=axes, title=title, data_name=data_name, **mpl_kwargs
+            )
+
+        axes.plot(self.time.values / 1e9, self._data.data.m, **mpl_kwargs)
+        axes.set_title(title)
```
CagtayFabry, line 643: does passing `None` here remove an existing title?
in that case I would prefer having the plot function only dealing with the line and labels, leaving title etc. to other functions (so that it is easier to add TimeSeries to existing axes)

weldx/core.py
vhirtham: Didn't check the function, just fixed a typo 😉. However, I think like with the `time` property, we should define what a `units` property is supposed to return and keep the interfaces aligned. A `pint.Unit` might be a more flexible option if it automatically converts to a string during a `print` or inside of a f-string. I think that most of the time the "user" calls this function he wants to print the unit in some way. The interfaces should be tailored to match the users' expectations.

weldx/core.py
vhirtham: Well, this is easy to catch with a single `if`. I will add the check.

vhirtham: Removed the parameter as discussed verbally.

weldx/core.py
vhirtham: Updated the docstring to reflect the actual behavior

weldx/core.py
vhirtham: Fixed it and added a `time_unit` parameter to switch to the desired unit, if the default one isn't playing nice with the plot

tutorials/timeseries_01.ipynb
```diff
@@ -133,7 +118,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "plot_ts(ts_constant.interp_time(time=t))"
+    "ts_constant.interp_time(time=t).plot(marker='o')"
```
CagtayFabry, line 121:
```suggestion
    "ts_constant.plot(time=t, marker='o')"
```
Does this work?
The less interp in user code the better 😃

vhirtham: Don't know why I missed that 😆

CagtayFabry (approved):
thanks for including `Signal` as well 👍

PR #428, opened 2021-07-15
[WeldxFile] fix show header in non interactive sessions

marscher:
## Changes
In non interactive sessions the output was not printed, but simply returned. This made it harder to write code running in different environments. Now WeldxFile.show_asdf_header() prints.

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 
 ### fixes
 
+- `WeldxFile.show_asdf_header` prints output on console, before it only returned the header
+  as parsed dict and string representation [[#]]().
+
 ### documentation
 
 ### ASDF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ### fixes
 
 - `WeldxFile.show_asdf_header` prints output on console, before it only returned the header
-  as parsed dict and string representation [[#]]().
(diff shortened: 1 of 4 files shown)


CagtayFabry (approved):
we can disable doctest for this example, I think this works well enough 👍

CagtayFabry:
what do you think about using `asdf.info` for console outputs? @marscher 
https://asdf.readthedocs.io/en/al-503-document-2.8-features/asdf/features.html#rendering-asdf-trees

marscher:
> what do you think about using `asdf.info` for console outputs? @marscher
> https://asdf.readthedocs.io/en/al-503-document-2.8-features/asdf/features.html#rendering-asdf-trees

Yes, that produces pretty output. I will stick to that and remove the doctest or think of another example. Thanks for your feedback!

CagtayFabry:
> 
> 
> > what do you think about using `asdf.info` for console outputs? @marscher
> > https://asdf.readthedocs.io/en/al-503-document-2.8-features/asdf/features.html#rendering-asdf-trees
> 
> Yes, that produces pretty output. I will stick to that and remove the doctest or think of another example. Thanks for your feedback!

agreed, for a simple console output that is pretty nice
should be simple to get the full yaml style output simply using `print()` when desired

PR #429, opened 2021-07-15
388 timeseries init

### LINKED ISSUE #388: TimeSeries __init__ should understand xarray structures
As the title says, a `TimeSeries` should be constructible from a xarray structure that matches the internal format. This is currently not possible but would make it much easier to convert the data from our lab.

vhirtham:
## Changes

`TimeSeries.__init__` accepts `xarray.DataArray` as `data` parameter.

## Related Issues

Closes #388 

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -352,11 +352,11 @@ class TimeSeries:
         if "time" != data_array.dims[0]:
             raise ValueError("The first dimension of the DataArray must be 'time'.")
 
-        if (time_coords := data_array.coords.get("time")) is None:
+        if "time" not in data_array.coords:
             raise ValueError("The 'DataArray' does not specify time values.")
 
         try:
-            ut.to_pandas_time_index(time_coords.data)
+            ut.to_pandas_time_index(data_array.coords.get("time").data)
         except TypeError:
             raise TypeError(
                 "The time values of the 'DataArray' must be convertible to a "
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -349,8 +349,8 @@ class TimeSeries:
     @staticmethod
     def _check_data_array(data_array: xr.DataArray):
         """Raise an exception if the 'DataArray' can't be used as 'self._data'."""
-        if "time" != data_array.dims[0]:
-            raise ValueError("The first dimension of the DataArray must be 'time'.")
+        if "time" not in data_array.dims:
+            raise ValueError("The 'DataArray' must have a dimension called 'time'.")
 
         if "time" not in data_array.coords:
             raise ValueError("The 'DataArray' does not specify time values.")
@@ -379,6 +379,10 @@ class TimeSeries:
 
         if isinstance(data, xr.DataArray):
             self._check_data_array(data)
+            if data.dims[0] != "time":
+                data = data.transpose(
+                    *ut.swap_list_items(data.dims, data.dims[0], "time")
+                )
             self._data = data
         else:
             # expand dim for scalar input
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -294,7 +294,7 @@ class TimeSeries:
         self._units = None
         self._interp_counter = 0
 
-        if isinstance(data, pint.Quantity):
+        if isinstance(data, (pint.Quantity, xr.DataArray)):
             self._initialize_discrete(data, time, interpolation)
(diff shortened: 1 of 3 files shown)


weldx/core.py
```diff
@@ -346,9 +346,29 @@ def __repr__(self):
...
+            ut.to_pandas_time_index(data_array.coords.get("time").data)
+        except TypeError:
+            raise TypeError(
+                "The time values of the 'DataArray' must be convertible to a "
+                "'pandas.TimedeltaIndex'."
+            )
+
+        if not isinstance(data_array.data, pint.Quantity):
+            raise TypeError("The data of the 'DataArray' must be a 'pint.Quantity'.")
+
```
CagtayFabry, line 368: I think this should be a place to use `util.xr_check_coords`

weldx/core.py
```diff
@@ -346,9 +346,29 @@ def __repr__(self):
             )
         return representation + f"Units:\n\t{self.units}\n"
 
+    @staticmethod
+    def _check_data_array(data_array: xr.DataArray):
+        """Raise an exception if the 'DataArray' can't be used as 'self._data'."""
+        if "time" != data_array.dims[0]:
+            raise ValueError("The first dimension of the DataArray must be 'time'.")
```
CagtayFabry, line 353: not sure if this is what you meant here but does the first dimension really have to be the time?
In that case we should just transpose the xarray correctly after the checks

weldx/core.py
vhirtham: Not sure about this. We have to create a reference `DataArray` just to replace 3 (the time position case will be removed) simple checks. Feels to be too much overhead to me. Additionally, the error messages are probably less clear to the user.

weldx/core.py
vhirtham: That's the internal convention for the `TimeSeries`. But you are right, if that convention is not met, we can simply transpose the dimensions. Will change it.

weldx/core.py
CagtayFabry: why would we have to create another `DataArray`?
```python
xr_check_coords(data_array, {time={"dtype": ["timedelta64[ns]"]})
```

I prefer this for consistency over individual implementations just for different error messages (maybe throw one meaningful error when `xr_check_coords` fails, that should be enough)
If the error messages turn out not to be helpful enough we can reassess later

vhirtham: > why would we have to create another `DataArray`?

I somehow misread the docstring 🤦 My bad...

weldx/core.py
```diff
@@ -357,24 +373,32 @@ def _initialize_discrete(
...
-            raise ValueError(
-                '"time" must be a time quantity or a "pandas.TimedeltaIndex".'
-            )
+        if isinstance(data, xr.DataArray):
+            self._check_data_array(data)
+            if data.dims[0] != "time":
+                data = data.transpose(
+                    *ut.swap_list_items(data.dims, data.dims[0], "time")
+                )
+            self._data = data
```
CagtayFabry, line 382: can't look up the code right now but there is a syntax that should work like:
~~~python
self. _data = data.transpose(..., "time") 
~~~
(check some LCS code)

CagtayFabry (changes requested):

weldx/core.py
vhirtham: works... good to know

CagtayFabry: I think the if check is not necessary with this

weldx/core.py
```diff
@@ -357,24 +373,30 @@ def _initialize_discrete(
...
-        if isinstance(time, pint.Quantity):
-            time = ut.to_pandas_time_index(time)
-        if not isinstance(time, pd.TimedeltaIndex):
-            raise ValueError(
-                '"time" must be a time quantity or a "pandas.TimedeltaIndex".'
-            )
+        if isinstance(data, xr.DataArray):
+            self._check_data_array(data)
+            if data.dims[0] != "time":
+                data = data.transpose("time", ...)
```
vhirtham, line 379:
```suggestion
            data = data.transpose("time", ...)
```

CagtayFabry (approved):

Issue #354, opened 2021-05-07
CSM graph serialization


vhirtham:
Now that we have a basic graph serialization function and schema, we should also apply this to the CSM. I think this would extremely simplify the implementation of the tag and should more or less solve #348 automatically. 

The only thing I am not sure about now are the subsystems, but I will have a deeper look into that.

PR #430, opened 2021-07-16
Sort csm subsystem lists

### LINKED ISSUE #423: sort string list output in asdf tree
in some places we store lists of strings in asdf files, most for the CSM and related functions notably here:
https://github.com/BAMWelDX/weldx/blob/9ffdba4c3594fe9bd73f214005d2a6ab8314369d/weldx/asdf/tags/weldx/core/transformations/coordinate_system_hierarchy.py#L83-L92

The sorting of these list items seem to change randomly when (re)creating files.
For consistency and better diffs it would be nice to sort these kind of lists before writing them to asdf.

Need to check if the order might be relevant to the implementation but I guess not for most cases?

vhirtham:
## Changes

Sort string lists before storing them in a WelDX file

## Related Issues

Closes #423

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/weldx/asdf/tags/weldx/core/transformations/coordinate_system_hierarchy.py
+++ b/weldx/asdf/tags/weldx/core/transformations/coordinate_system_hierarchy.py
@@ -91,6 +91,11 @@ class CoordinateSystemManagerSubsystem:
     subsystems: List[str]
     members: List[str]
 
+    def __post_init__(self):
+        """Sort the string lists."""
+        self.subsystems = sorted(self.subsystems)
+        self.members = sorted(self.members)
+
 
 class CoordinateSystemManagerSubsystemASDF(WeldxType):
     """Serialization class for a CoordinateSystemManagerSubsystem instance"""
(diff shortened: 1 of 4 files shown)


vhirtham:
I adjusted the CSM serialization and our `weldx.asdf.util.dataclass_serialization_class` function. That should cover most cases. Didn't find any further corresponding type hints in the asdf directory and searching for

~~~
type: array
    items:
      type: string
~~~

in the schemas only brought up an additional case in the CSM I missed initially. I guess we covered everything.

weldx/asdf/tags/weldx/core/transformations/coordinate_system_hierarchy.py
```diff
@@ -421,7 +421,7 @@ def to_tree(cls, node: CoordinateSystemManager, ctx):
         tree = {
             "name": node.name,
             "reference_time": node.reference_time,
-            "subsystem_names": sorted(subsystems),
+            "subsystem_names": subsystems,
```
vhirtham: Okay, seems that order actually matters here for nested subsystems, but I think we should tackle this when we work on #354

CagtayFabry: yes lets try this first and see later

CagtayFabry (approved):

PR #433, opened 2021-07-19
Time class

### LINKED ISSUE #367: Time class
We are currently supporting multiple time formats in several places:
- pandas (`DateTimeIndex`, `TimedeltaIndex`, `Timestamp`)
- numpy (time dtypes)
- pint quantities
- xarray structures with "time" and "time_ref" fields

Dealing with time as an additional dimension can actually be quite challenging when trying to enhance, understand or write new code. But also having to deal with those different types turns it into hell. For example, if I just want to compare if the last entry of a time array is less than the first entry of another one I need to figure out if I have relative (time delta) or absolute (dates) values and if they have matching types. Then I need to treat all cases separately. This bloats the code and makes it hard to understand. 

Furthermore, the `time` property of our classes is not consistent regarding its return type. For example, the LCS returns a xarray structure, while the `TimeSeries` returns a `TimedeltaIndex`. The LCS also provides a special method to get time as a quantity to take away the need for type casts from the user.
 
My suggestion is, that we define a `Time` wrapper class, that serves multiple purposes. First, it should be an interface class that is always used when any public function returns a time type. Second, it should be constructable from all the classes that we currently use and provide methods to cast back to all those types like `WXRotation` does for rotations. Lastly, it should provide all the functionalities that we frequently need like calculating time unions and synchronizing reference times.

I would also like to keep time deltas and date times in a single class instead of having two separate classes (possibly three with timestamps). The class method implementations should deal with the compatibility of relative and absolute times so that we do not have to cover it separately every time we deal with time formats.

To make a suggestion of some of the possible interfaces:

~~~ python
class WXTime:
   
    def __init__(
        self,
        time : Union[DateTimeTypes, TimedeltaTypes, TimestampTypes, WXTime],
        time_ref : Union[TimestampTypes, WXTime],
    )

    def to_pandas(self) -> Union[TimedeltaIndex, DateTimeIndex, Timestamp]:

    def to_quantity(self) -> Quantity:

    @property
    def length(self) -> int:

    @property
    def has_reference_time(self) -> bool:

    @property
    def is_timestamp(self) -> bool:
        return self.lengt == 1 and self.has_reference_time

    @property
    def reference_time() -> WXTime:

    @reference_time.setter
    def reference_time( time_ref : Union[TimestampTypes, WXTime]):

    @staticmethod
    def time_union(times = List[WXTime]) -> WXTime:

    def __eq__(self, other:  Union[DateTimeTypes, TimedeltaTypes, TimestampTypes, WXTime]) ->bool:

    def  __le__(self, other:  Union[DateTimeTypes, TimedeltaTypes, TimestampTypes, WXTime]) ->bool:

    def all_close(self, other:  Union[DateTimeTypes, TimedeltaTypes, TimestampTypes, WXTime]) ->bool:

    def overlaps_with(self, other:  Union[DateTimeTypes, TimedeltaTypes, TimestampTypes, WXTime]) ->bool:
~~~

I think this will make our life and of our possible users a lot easier since it bundles all the time-related stuff in one place and unifies the code. What do you think?

vhirtham:
## Changes

Add `weldx.time.Time` 

## Related Issues

Closes #367

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


weldx/time.py
```diff
@@ -0,0 +1,125 @@
...
+    from pint import Quantity
+    from xarray import DataArray
+
+    from weldx.types import types_time_like, types_timestamp_like
+
+
+class Time:
+    """Provides a unified interface for time related operations."""
+
+    def __init__(
```
CagtayFabry, line 17: I think one of the core question to decide is what datatype(s) to use for storing the actual data inside the class

I would suggest using the basic `pandas` types here (`Timedelta`, `TimedeltaIndex`, `Timestamp`, `DateTimeIndex`) and storing them in a simple unified place like `_data` (original, i know 😉)
Every input gets converted into the appropriate pandas type

That way, most of the actual operations can be implemented using basic pandas syntax which should be very easy to implemented (and also throw reasonable exceptions for most cases without us having to worry about them).
I also expect the performance to be superior to using `xarray`.

We should also discuss and define some very basic implementation details of the class:
- do we allow duplicate values (I'd say no)
- do we enforce the time values to be sorted (I'd say yes!)


I added this comment to the `__init__` function so that we can mark it as 'resolved' later

weldx/time.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def max(self) -> "Time":
+        """Get the maximal time of the data."""
+
+    def min(self) -> "Time":
+        """Get the minimal time of the data."""
+
+    @staticmethod
+    def union(times=List[Union[types_time_like, "Time"]]) -> "Time":
+        """Calculate the union of multiple `Time` instances.
```
CagtayFabry, line 113: It would be awesome to have this work as a class method and as a regular method (creating union of existing Time instance with other)

maybe this can get us there:
https://stackoverflow.com/a/28238212

weldx/time.py
vhirtham: The docstring of the `__init__` methods' `time` parameter already agrees with your assessment of the two bullet points ("ascending order"). Since I wrote it I guess it is settled then 😉 

Regarding the pandas time types: This is probably the way that makes the most sense performance-wise and in every other context. 

I think we should also review the usage of the special xarray time structure in other classes. Most of the use cases are related to having a unified interface, that will now be provided by this class. But this is something for later, if the class itself is finished

weldx/time.py
vhirtham: Nice idea/find. We will get it done 😉 But first I have to [learn more about descriptors](https://docs.python.org/3/howto/descriptor.html)

weldx/time.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def as_datetime_index(self) -> DatetimeIndex:
+        """Return the data as `pandas.DatetimeIndex`."""
+
+    def as_timestamp(self) -> Timestamp:
+        """Return the data as `pandas.Timestamp`."""
+
+    @property
+    def data_array(self) -> DataArray:
+        """Return the internally stored data."""
```
vhirtham, line 81:
```suggestion
    def as_data_array(self) -> DataArray:
        """Return the data as `xarray.DataArray`."""
```

weldx/time.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def as_time_delta_index(self) -> TimedeltaIndex:
+        """Return the data as `pandas.TimedeltaIndex`."""
+
+    def as_datetime_index(self) -> DatetimeIndex:
+        """Return the data as `pandas.DatetimeIndex`."""
+
+    def as_timestamp(self) -> Timestamp:
+        """Return the data as `pandas.Timestamp`."""
+
```
vhirtham, line 78:
```suggestion
    def as_quantity(self) -> Quantity:
        """Return the data as `pint.Quantity`."""

    def as_time_delta_index(self) -> TimedeltaIndex:
        """Return the data as `pandas.TimedeltaIndex`."""

    def as_datetime_index(self) -> DatetimeIndex:
        """Return the data as `pandas.DatetimeIndex`."""

    def as_timestamp(self) -> Timestamp:
        """Return the data as `pandas.Timestamp`."""

    def as_pandas(self):
        pass
     

```

vhirtham:
Minor thing:

Should the `__init__` support something like this?

~~~ python
Time(["2000", "2001", "2002"])
~~~

I think this can be passed directly to `DatetimeIndex` and should just take 2 lines at max. But --> reaaaaaally low priority

CagtayFabry:
> Minor thing:
> 
> Should the `__init__` support something like this?
> 
> ```python
> Time(["2000", "2001", "2002"])
> ```
> 
> I think this can be passed directly to `DatetimeIndex` and should just take 2 lines at max. But --> reaaaaaally low priority

it is currently supported via the existing utility functions so I would suggest yes
(also works for Timedeltas: `Time(["2s", "3s", "5s"])`)

vhirtham:
Follow up:

Do we want to allow this?

~~~ python
Time("10s")
~~~

Makes it easier to use but might conflict with:

~~~ python
Time("2002-01-01")
~~~

But nothing that a `try` `except` block can't handle.

vhirtham:
> > Minor thing:
> > Should the `__init__` support something like this?
> > ```python
> > Time(["2000", "2001", "2002"])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this can be passed directly to `DatetimeIndex` and should just take 2 lines at max. But --> reaaaaaally low priority
> 
> it is currently supported via the existing utility functions so I would suggest yes
> (also works for Timedeltas: `Time(["2s", "3s", "5s"])`)

Okay, then I have to add it as a testcase

weldx/tests/test_time.py
```diff
@@ -0,0 +1,127 @@
...
+            (np.array(["2000", "2010"], dtype="datetime64"), None, True, "2000"),
+            (np.array(["2000", "2010"], dtype="datetime64"), "2005", True, "2005"),
+            (Timestamp("2000"), None, True, "2000"),
+            (Timestamp("2000"), "2005", True, "2000"),
+            (Timedelta(1, "d"), None, False, None),
+            (Timedelta(1, "d"), "2005", True, "2005-01-02"),
+            (np.datetime64("2000"), None, True, "2000"),
+            (np.datetime64("2000"), "2005", True, "2000"),
+            (np.timedelta64(1, "s"), None, False, None),
+            (np.timedelta64(1, "s"), "16:00", True, "16:00:01"),
```
vhirtham, line 53: We should rediscuss the behavior in this case here. It differs from the array behavior. We put in a time delta with a reference time, which gets turned into a Timestamp. Now we ignore the reference time. If we would call `as_timedelta`, it would return `"0s"`. Get's even worse if we have a single value array like `np.timedelta64([1], "s")` which will also be converted to a timestamp. On the other hand, `np.timedelta64([1, 2], "s")` will return `"1s"` and `"2s"` . This inconsistency can lead to serious hard to find bugs in some kind of automatic processing. So it boils down to also storing and handling reference times for timestamps.

CagtayFabry: reference time should also be respected for timestamps, true 👍 
should be fixed now

weldx/time.py
CagtayFabry: I added a check for monotonic increasing values after conversion to the equivalent pandas types

that should be safe enough since we don't do any kind of sorting beforehand

weldx/tests/test_time.py
```diff
@@ -0,0 +1,296 @@
...
+            (Q_, "timedelta"),
+            TimedeltaIndex,
+            Timedelta,
+            np.timedelta64,
+            (str, "datetime"),
+            (Time, "datetime"),
+            (Q_, "datetime"),
+            DatetimeIndex,
+            Timestamp,
+            np.datetime64,
```
vhirtham, line 150: The list of our supported time types (all test pass 😄 )

weldx/tests/test_time.py
```diff
@@ -0,0 +1,296 @@
...
+        val = val[0] if data_was_scalar else val
+        exp_timedelta = (
+            Timedelta(val, "s") if data_was_scalar else TimedeltaIndex(val, "s")
+        )
+
+        # expected datetime
+        exp_datetime = None
+        if exp_is_absolute:
+            time_ref = Timestamp(time_ref if time_ref is not None else abs_val[0])
+            exp_datetime = time_ref + exp_timedelta
```
vhirtham, line 123: This function describes the behavior that we expect from our `Time` class for the different input types. 
If `exp_datetime` is `None` we expect a `TypeError` 😉

weldx/tests/test_time.py
vhirtham: Had to remove one more line in the `__init__`.

CagtayFabry (approved):
this looks ready to me for an initial implementation 👍

vhirtham:
> this looks ready to me for an initial implementation 👍

I will add some more info to the docstrings and merge later this day

vhirtham:
> > this looks ready to me for an initial implementation 👍
> 
> I will add some more info to the docstrings and merge later this day

Have to postpone the merge to monday.

PR #434, opened 2021-07-19
Update changelog action

vhirtham:
## Changes

Disables the changelog action if the pull request is marked as "draft" and/or not targetting our master branch.


CagtayFabry (approved):
makes sense 💯 👍

CagtayFabry:
wondering why it failed here though 🤔

vhirtham:
There are multiple runs of this action listed. The first one failed because of the missing tag. After I added the tag, the action was restarted and skipped correctly. Since there is no new commit in conjunction with the relabeling, it looks a bit confusing because the failed run is still tied to the last commit, even though it does not represent the final outcome of the check.

PR #435, opened 2021-07-19
Add mypy

### LINKED ISSUE #427: Add mypy to CI pipelines
`mypy` is a tool that checks if the correct types are used (for example during a function call). It uses python type annotations and compares them against the actual type in use. 

Here is the [GitHub repo](https://github.com/python/mypy) and the [documentation](https://mypy.readthedocs.io/en/stable/). Since Guido van Rossum (somebody told me this guy knows a bit about python 😉) participates in the project, I guess it might be a good choice over other alternatives. I just tried it on our `core.py`. Here is the output:

~~~
weldx\core.py:298: error: Argument 3 to "_initialize_discrete" of "TimeSeries" has incompatible type "Optional[str]"; expected "str"  [arg-type]
weldx\core.py:326: error: "None" has no attribute "identical"  [attr-defined]
weldx\core.py:376: error: Argument 1 to "rename" of "DataArray" has incompatible type "Dict[str, str]"; expected "Union[Hashable, Mapping[Hashable, Hashable], None]"  [arg-type]
weldx\core.py:376: note: Following member(s) of "Dict[str, str]" have conflicts:
weldx\core.py:376: note:     __hash__: expected "Callable[[], int]", got "None"
weldx\core.py:436: error: Argument 1 to "xr_interp_like" has incompatible type "None"; expected "DataArray"  [arg-type]
weldx\core.py:438: error: Argument "method" to "xr_interp_like" has incompatible type "Optional[str]"; expected "str"  [arg-type]
weldx\core.py:468: error: "None" has no attribute "evaluate"  [attr-defined]
weldx\core.py:476: error: Argument 1 to "rename" of "DataArray" has incompatible type "Dict[str, str]"; expected "Union[Hashable, Mapping[Hashable, Hashable], None]"  [arg-type]
weldx\core.py:476: note: Following member(s) of "Dict[str, str]" have conflicts:
weldx\core.py:476: note:     __hash__: expected "Callable[[], int]", got "None"
weldx\core.py:618: error: Incompatible return value type (got "None", expected "Tuple[Any, ...]")  [return-value]
weldx\core.py:632: error: Incompatible return value type (got "None", expected "str")  [return-value]
~~~

We will certainly have to put a bit of work into fixing all issues in the package and tweaking the tool, but most of the stuff should be simply adjusting wrong type annotations due to interface changes, etc.

I think it will further improve the quality of our package. What do you think?

vhirtham:
## Changes

Adds `mypy` to the CI pipeline and fixes all type-hint errors

## Related Issues

Closes #427

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] ~~updated doc/~~


vhirtham:
After understanding the error messages, their origins, and some of the options that we can use to configure `mypy`, there were only 4 lines to change in the core module. 2 of the fixes were just missing type hints. In the other 2 cases, `mypy` couldn't follow the full logic to exclude a type from the possible options so I had to tell it to ignore the line. I guess updating the other modules shouldn't be too much pain.

CagtayFabry:
Thanks for looking into this 👍 

I think we should review this after refactoring towards the release of `0.5.0`

vhirtham:
> Thanks for looking into this 👍
> 
> I think we should review this after refactoring towards the release of `0.5.0`

Fixing the issues is rather easy. I think I will fix 3-5 per day when I need to clear my head. I'll keep the branch updated and postpone stuff that will be changed in `0.5.0` anyway. We can then perform the final review after everything is refactored for the `0.5.0` release, as you suggested.

weldx/core.py
```diff
@@ -289,8 +289,8 @@ def __init__(
             'step', 'linear'.
 
         """
-        self._data = None
-        self._time_var_name = None
+        self._data = None  # type: Union[MathematicalExpression, xr.DataArray]
```
marscher, line 292: why do you choose such a format here? You could directly attach the type annotation to the variable, right?

marscher:
```suggestion
        self._data: Union[MathematicalExpression, xr.DataArray] = None
```

weldx/core.py
```diff
@@ -324,7 +324,7 @@ def __eq__(self, other: Any) -> bool:
         if not isinstance(self.data, MathematicalExpression):
             if not isinstance(other.data, pint.Quantity):
                 return False
-            return self._data.identical(other.data_array)
+            return self._data.identical(other.data_array)  # type: ignore
```
marscher, line 327: shouldn't mypy inspect the return type of the identical method then?

setup.cfg
```diff
@@ -128,3 +128,39 @@ exclude_lines =
...
+allow_redefinition = True
+show_error_context = True
+show_error_codes = True
+show_column_numbers = True
+
+[mypy-weldx.asdf.*,weldx.tests.*]
+ignore_errors = True
+
+# fix current errors, remove module from ignore list, repeat
+[mypy-weldx.visualization.*]
```
marscher, line 148: and fixing these is a week of work, isn't it?

vhirtham: No

weldx/core.py
vhirtham: Maybe. Works for dataclasses. Not sure if it works for local vars too

weldx/core.py
vhirtham: The problem here is a bit different. `self._data` isn't guaranteed to be a xarray at this point, or at least, mypy doesn't see it that way. So it complains about something like "`Any` has no method `identical`". We could change the code to give this guarantee, but this might have side effects requiring more attention. If the code does what it is meant to do I acknowledge that there might be an issue by just adding `# type: ignore`. Not every reported issue is really relevant and needs to be taken care of.

weldx/time.py
```diff
@@ -245,8 +245,8 @@ def __init__(
         if isinstance(time, pd.Index) and not time.is_monotonic_increasing:
             raise ValueError("The time values passed are not monotonic increasing.")
 
-        self._time = time
-        self._time_ref = time_ref
+        self._time: Union[pd.TimedeltaIndex, pd.DatetimeIndex] = time
+        self._time_ref: pd.Timestamp = time_ref
```
CagtayFabry, line 249: I had no idea we could use type hints just like that in code 👀

vhirtham: Me neither, but I find it to be really useful.

setup.cfg
```diff
@@ -128,3 +128,30 @@ exclude_lines =
...
+show_error_codes = True
+show_column_numbers = True
+
+[mypy-weldx.asdf.*]
+ignore_errors = True
+[mypy-weldx.tests.*]
+ignore_errors = True
+# remove after refactoring the geometry module
+[mypy-weldx.geometry.*]
+ignore_errors = True
```
vhirtham, line 150: Tests are ignored. Asdf for now too. Don't know if we want to include it at some point. Geometry should be fixed when it has been refactored.

vhirtham:
Also, notice our [discussion on how to fix mypy errors](https://github.com/BAMWelDX/weldx/discussions/446)

CagtayFabry (approved):
nice 🏅 👍

PR #440, opened 2021-07-20
Timeseries absolute times

CagtayFabry:
## Changes
demo implementation to support absolute times in `TimeSeries`


CagtayFabry:
I had to avoid some import troubles but the implementation is pretty straight forward
not really fleshed out but the idea seems to work well enough, let me know what you think @vhirtham

weldx/core.py
```diff
@@ -385,10 +390,16 @@ def _initialize_discrete(
...
-                time = pd.TimedeltaIndex([0])
+                time = pd.Timedelta(0)
+
+            time = Time(time)
+
+            if time.is_absolute:
+                self._reference_time = time.reference_time
+                time = time - time.reference_time
+
+            time = time.as_pandas_index()
```
vhirtham, line 401: My idea would be to rework the internal structure of the `TimeSeries` to actually use `Time` as `self._time` instead of just using it for conversions. We can still provide the combined xarray structure as a possible output to the user, but I think it is a bit unwieldy for internal use compared to `Time`. But I have not fully looked into the code yet. It might require approaching interpolations differently. I have to think about it.

CagtayFabry: I don't think that works because xarray will create it's own internal array anyway
anything not xarray at this point would be way too much overhead imo (and also differ from GenericSeries eventually anyway)

using the unified interfaces should make a lot of stuff way easier

CagtayFabry: for what it's worth, we can simply return a `Time` object for the `.time` property so that would basically have the same effect for user facing interfaces (and we would do that throughout `weldx`)

vhirtham: After some thinking, you might be right. Doesn't make that much sense.

> for what it's worth, we can simply return a `Time` object for the `.time` property so that would basically have the same effect for user facing interfaces (and we would do that throughout `weldx`)

That should be the case and was one of the motivations. The `Time` class should be our main interface object.

vhirtham:
Deleting the target branch closed this PR automatically and somehow I can't reopen it with a new target. So just create a new PR, please.

vhirtham:
> Deleting the target branch closed this PR automatically and somehow I can't reopen it with a new target. So just create a new PR, please.

Somehow, I overlooked the "restore branch" button :smile:

vhirtham:
Still trying to figure out what causes the one failure in the tests. The debugger shows that time is an array of `Timestamp` instead of a `DatetimeIndex`. Maybe this is the reason. Have to check why this happens. Any clues?

CagtayFabry:
so far the timeseries always falls back to timedelta values even when working with absolute times
if you want to use this we should probably implement the correct behaviour @vhirtham 

Or da that later?

vhirtham:
> so far the timeseries interpolation always falls back to timedelta values even when interpolating against absolute times
> if you want to use this we should probably implement the correct behaviour @vhirtham
> 
> Or da that later?

I will have a detailed look tomorrow.

CagtayFabry:
It's a similar situation like the CSM:
if `TimeSeries` is defined without reference time, what do we do when interpolating against absolute times (currently simply using the interpolation reference time as reference time)

This could be tricky when evaluating the same `TimeSeries` against different times and expecting the same reference time (without explicitly setting it) but I think it is ok for now

vhirtham:
> It's a similar situation like the CSM:
> if `TimeSeries` is defined without reference time, what do we do when interpolating against absolute times (currently simply using the interpolation reference time as reference time)
> 
> This could be tricky when evaluating the same `TimeSeries` against different times and expecting the same reference time (without explicitly setting it) but I think it is ok for now

I would simply forbid that and require the user to set a reference time before using absolute times. But we can do this later. We should just create a reminder issue

weldx/core.py
```diff
@@ -270,8 +270,9 @@ class TimeSeries:
     def __init__(
         self,
         data: Union[pint.Quantity, MathematicalExpression],
-        time: Union[None, pd.TimedeltaIndex, pint.Quantity] = None,
+        time: Union[types_time_like, Time] = None,
```
vhirtham, line 273: This is totally unrelated to this PR, but we should find a solution without ring dependencies to include `Time` into `types_time_like` and other fitting time type definitions

weldx/core.py
```diff
@@ -270,8 +270,9 @@ class TimeSeries:
...
         interpolation: str = None,
+        reference_time: pd.Timestamp = None,
```
vhirtham, line 275: Also not related to this PR, but a reminder to find a common interface name for the reference time and apply it to all functions

weldx/core.py
```diff
@@ -386,10 +388,16 @@ def _initialize_discrete(
...
+                self._reference_time = time.reference_time
+                time = time - time.reference_time
+
+            time = time.as_pandas_index()
 
-            if isinstance(time, pint.Quantity):
-                time = ut.to_pandas_time_index(time)
             if not isinstance(time, pd.TimedeltaIndex):
                 raise ValueError(
                     '"time" must be a time quantity or a "pandas.TimedeltaIndex".'
```
vhirtham, line 403:
~~~
                )
            dax = xr.DataArray(data=data)
            self._data = dax.rename({"dim_0": "time"}).assign_coords({"time": time})
~~~

(Couldn't mark the lines above, so I added them to avoid confusion)

Maybe it does make sense to create an extra generalized `util.xr_` function for this. This basically just adds a Time object as a dimension to a xarray. Guess we have already used something similar elsewhere?

vhirtham: Also, I guess the exception is obsolete if we use the `as_timedeltaindex` from the other PR hera

weldx/core.py
```diff
@@ -496,7 +484,7 @@ def _interp_time_expression(
             data = np.expand_dims(data, 0)
 
         dax = xr.DataArray(data=data)  # don't know exact dimensions so far
-        return dax.rename({"dim_0": "time"}).assign_coords({"time": time_pd})
+        return dax.rename({"dim_0": "time"}).assign_coords({"time": time})
```
vhirtham, line 487: Here is at least a partial repetition of the mentioned code above. Not sure if we can use it.

weldx/core.py
```diff
@@ -615,10 +608,21 @@ def interp_time(
...
+        else:
+            time_interp = time
+        time_interp = time_interp.as_pandas_index()
+
         if isinstance(self._data, xr.DataArray):
-            dax = self._interp_time_discrete(time)
+            dax = self._interp_time_discrete(time_interp)
         else:
-            dax = self._interp_time_expression(time, time_unit)
+            dax = self._interp_time_expression(time_interp, time_unit)
```
vhirtham, line 625: If you don't mind, I will try to further simplify this by passing the `Time` instance directly to the internal interpolation methods.

CagtayFabry: It was like that before (passing Time) but really didn't help because you just end up doing the conversion in both internal functions

weldx/core.py
CagtayFabry: cant we just define `types_time_like` (and other time related types) in the `.time` module would also make sense to put it there imo

weldx/core.py
vhirtham: Just have a look, at commit 9e95f7497948a2600ec6398e3e8c309ab6d44b5d 
Maybe my comment was lacking some information to understand what I had in mind

weldx/core.py
vhirtham: Yes, that's a good option I also thought of. But I wasn't sure about it since we have the dedicated types module. Still not sure how to add `Time` to `types_time_like` while using it all over the place in the classes type hints. Does Python have some sort of forward declaration (will check that)?

vhirtham (approved):

weldx/core.py
vhirtham: Done in PR #468

weldx/core.py
vhirtham: Resolved in PR #445

weldx/core.py
vhirtham: Resolved in PR #445

PR #442, opened 2021-07-21
add "units" and "dimensionality" options to `xr_check_coords`

CagtayFabry:
## Changes
add "units" and "dimensionality" options to `xr_check_coords`

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### added
 
+- added "units" (exact) and "dimensionality" (dimensionality compatible) checking options
+  to `util.xr_check_coords` [[#442]](https://github.com/BAMWelDX/weldx/pull/442)
+
 ### removed
 
 ### changes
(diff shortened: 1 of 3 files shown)


vhirtham (approved):
LGTM, just wondering, if it would make sense to put all the sub functionalities into extra functions like `xr_check_coord_dimensionality` etc. so that the code gets a bit more relaxed and we can pick specialized functions if we only want to check one specific characteristic of a coordinate.

CagtayFabry:
> ... and we can pick specialized functions if we only want to check one specific characteristic of a coordinate.

just pass only the desired characteristic into `xr_check_coords`

vhirtham:
> > ... and we can pick specialized functions if we only want to check one specific characteristic of a coordinate.
> 
> just pass only the desired characteristic into `xr_check_coords`

The point was just that the function implementation looks a bit more cleaned up when it gets subdivided into multiple smaller ones. Also, I guess the interfaces/overhead gets probably a bit lighter for the subfunctions. Yes, you can always use the "I do everything"-function for a smaller task, but having the opportunity to choose isn't too bad. But it was just an idea I had during the review.

PR #448, opened 2021-07-26
Use `Time` in `LocalCoordinateSystem` and `CoordinateSystemManager`

vhirtham:
## Changes

Use the new `Time` class in the `LocalCoordinateSystem` and the `CoordinateSystemManager`

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [x] ~~update example/tutorial notebooks~~ refactor in an extra PR


weldx/transformations/cs_manager.py
```diff
@@ -1356,24 +1366,21 @@ def get_cs(
         if time is None:
             time_ref = None  # ignore passed reference time if no time was passed
             time = self.time_union(path_edges)
-
         elif isinstance(time, str):
+            # todo: this branch conflicts a bit with our other str inputs for time.
+            #       We should catch that with a try, except
             parent_name = self.get_parent_system_name(time)
             if parent_name is None:
                 raise ValueError("The root system has no time dependency.")
```
vhirtham, line 1374: Should fix this before merging

weldx/transformations/cs_manager.py
```diff
@@ -1387,12 +1394,14 @@ def get_cs(
...
-                    lcs.reset_reference_time(time_ref_interp)
+                    lcs = lcs.interp_time(time - self.reference_time)
+                    lcs = LocalCoordinateSystem(
+                        lcs.orientation.data, lcs.coordinates.data, time
+                    )
                 else:
-                    lcs = lcs.interp_time(time_interp, time_ref_interp)
+                    if time_ref is None and len(path_edges) == 1:
+                        time_ref = lcs.reference_time
+                    lcs = lcs.interp_time(time, time_ref)
```
vhirtham, line 1404: The missing branch was actually an untested bug in the previous implementation. Maybe it does make sense to handle the special case that we have just a single edge outside the loop. Should yield better performance and also makes the code less hard to understand.

weldx/transformations/types.py
```diff
@@ -13,6 +13,7 @@
 
 types_coordinates = Union[xr.DataArray, np.ndarray, List]
 types_orientation = Union[xr.DataArray, np.ndarray, List[List], Rotation]
+# todo: remove next two definitions if possible
```
vhirtham, line 16: Reminder

weldx/transformations/types.py
```diff
@@ -13,6 +13,7 @@
...
 types_timeindex = Union[pd.DatetimeIndex, pd.TimedeltaIndex, pint.Quantity]
 types_time_and_lcs = Union[
     types_timeindex, "weldx.transformations.LocalCoordinateSystem"
```
vhirtham, line 19: Reminder

weldx/transformations/local_cs.py
```diff
@@ -329,19 +309,29 @@ def _build_coordinates(cls, coordinates, time: pd.DatetimeIndex = None):
...
-        time: types_timeindex = None,
-        time_ref: pd.Timestamp = None,
-    ) -> Tuple[pd.TimedeltaIndex, pd.Timestamp]:
-        if (
+    def _build_time(
+        coordinates: Union[types_coordinates, TimeSeries, None],
+        time: Union[types_time_like, Time, None],
+        time_ref: Union[types_timestamp_like, Time, None],
+    ) -> Union[Time, None]:
+        # check if this function can be refactored with the TimeSeries supporting Time
```
vhirtham, line 317: Revisit if corresponding branch gets merged

weldx/transformations/local_cs.py
```diff
@@ -905,27 +912,21 @@ def interp_time(
...
         if isinstance(self.coordinates, TimeSeries):
-            time_interp = time
-            if isinstance(time_interp, pd.DatetimeIndex):
-                time_interp = time - self.reference_time
+            time_interp = Time(time, self.reference_time)
 
+            # todo: remove as_timedelta when TimeSeries supports Time
             coordinates = self._coords_from_discrete_time_series(
-                self.coordinates.interp_time(time_interp)
+                self.coordinates.interp_time(time_interp.as_timedelta())
```
vhirtham, line 922: Reminder

weldx/transformations/local_cs.py
```diff
@@ -889,14 +895,15 @@ def interp_time(
...
 
-        # use LCS reference time if none provided
-        if isinstance(time, LocalCoordinateSystem) and time_ref is None:
-            time_ref = time.reference_time
-        time = ut.to_pandas_time_index(time)
+        # handle LCS as time
+        if isinstance(time, LocalCoordinateSystem):
+            if time_ref is None:
+                time_ref = time.reference_time
+            time = time.time
```
vhirtham, line 902: This should be revisited if we add a `TimeDependent` metaclass. `Time` should be capable to handle it on its own

vhirtham:
I suggest we merge the updated `TimeSeries` first. Afterwards, I will review the reminders above before merging this PR. I also think that we should update the interfaces (`CLASS.time` always returns a `Time` instance) in a follow-up PR. Since most interfaces already support `Time` as input, the additional work shouldn't be too much. 

The `plot` functions still need to be updated. Should we do it in this PR or a separate one?

weldx/transformations/cs_manager.py
```diff
@@ -787,7 +780,8 @@ def create_cs_from_euler(
         angles,
         degrees: bool = False,
         coordinates: types_coordinates = None,
-        time: pd.DatetimeIndex = None,
+        time: Union[types_time_like, Time, None] = None,
+        time_ref: Union[types_timestamp_like, None] = None,
```
CagtayFabry, line 784: Since we want to get rid of the `time_ref` parameter in the long run anyway I would prefer not to introduce this for existing functions, the user should supply a proper `Time` object here

vhirtham: I am not 100% sure if we should really get rid of the `time_ref` since it gives the user the flexibility to pass in whatever he wants without any need to involve another class. The inputs are forwarded to `Time` anyways. I like this kind of flexibility.

Apart from that, I think we should remove or merge most of the LCS construction methods (`from_xyz` etc.) and the corresponding `create` methods on the CSM.

CagtayFabry: > I am not 100% sure if we should really get rid of the `time_ref` since it gives the user the flexibility to pass in whatever he wants without any need to involve another class. The inputs are forwarded to `Time` anyways. I like this kind of flexibility.

imo it bloats the parameters and docstrings of many functions while providing the exact functionality of the `Time` class, maybe keep it as short as possible and add some more flexibility later where we really need it?
We could also support tuple arguments for `time` where the second item is the reference time so no explicit `Time()` call needed (we should add that to the `Time` constructor)

> 
> Apart from that, I think we should remove or merge most of the LCS construction methods (`from_xyz` etc.) and the corresponding `create` methods on the CSM.

I am also for removing them

vhirtham: I think that this is a discussion for the future and not tied to this PR. The `create_` methods were just thin wrappers around the LCS `from_` functions. The modified places missed the `time_ref` parameter that is part of the corresponding `from_` function. I added it because I had to fix the type hints of all those functions. I do not see the point in inverting the modifications when the `create_` functions still possess this parameter.

weldx/transformations/local_cs.py
```diff
@@ -257,33 +259,21 @@ def _comp_coords():
...
             if isinstance(orientation, Rot):
                 orientation = orientation.as_matrix()
             elif not isinstance(orientation, np.ndarray):
                 orientation = np.array(orientation)
 
-            if orientation.ndim == 3:
-                time_orientation = time
+            time_orientation = (
+                time.as_timedelta_index() if orientation.ndim == 3 else None
+            )
```
CagtayFabry, line 276: in cases like these the 'old' implementation could be beneficial over the one-liner for seeing the branch coverage in tests

weldx/transformations/local_cs.py
```diff
@@ -292,35 +282,25 @@ def _build_orientation(
...
         if not isinstance(coordinates, xr.DataArray):
-            time_coordinates = None
             if not isinstance(coordinates, (np.ndarray, pint.Quantity)):
                 coordinates = np.array(coordinates)
-            if coordinates.ndim == 2:
-                time_coordinates = time
+
+            time_coordinates = (
+                time.as_timedelta_index() if coordinates.ndim == 2 else None
+            )
```
CagtayFabry, line 303: I think we should simply add support for passing `Time` in `xr_3d_vector` and `xr_3d_matrix` instead of handling this here

vhirtham: > I think we should simply add support for passing `Time` in `xr_3d_vector` and `xr_3d_matrix` instead of handling this here

Changed it as suggested. Have a look into both of the updated functions

weldx/transformations/local_cs.py
vhirtham: Yes, that is actually a good point.

CagtayFabry (approved):

weldx/transformations/local_cs.py
vhirtham: No real optimization possibility found here

weldx/transformations/local_cs.py
vhirtham: Done in #464

weldx/transformations/local_cs.py
vhirtham: Done in #464

weldx/transformations/types.py
vhirtham: Done in PR #468

weldx/transformations/cs_manager.py
vhirtham: Done in PR #468

weldx/transformations/cs_manager.py
vhirtham: Done in PR #468

PR #449, opened 2021-07-27
[csm/k3d-viz] downcast coordinates to float32

marscher:
## Changes

Internally k3d expects a single precision array for the coordinates and emits a warning. This is fixed by this change.


vhirtham (approved):

CagtayFabry:
the status of the `Confirm changelog entry ` is really confusing me, does it always look like this when creating a draft first?

vhirtham:
> the status of the `Confirm changelog entry ` is really confusing me, does it always look like this when creating a draft first?

Only if there was no follow up commit after the tag was added and only after you merged. Before the merge, you only see the last execution.

PR #450, opened 2021-07-27
Install guide

### LINKED ISSUE #396: add jupyter setup documentation
we could add a small section to the documentation that describes a complete jupyter(lab) installation with conda, including the additional tools we currently use such as `k3d`, `ipympl` etc.

marscher:
## Changes

Added a draft for an installation guide, which includes the creation of two separate environments. The jupyter lab installation is separated from weldx itself. The mandatory extensions are installed (currently k3d) and the weldx env is exported as a kernel into jupyter lab.

## Related Issues

Closes #396 

## Checks

- [x] updated CHANGELOG.md
- [x] updated doc/


doc/install.rst
```diff
@@ -0,0 +1,64 @@
...
+all subsequent commands will not be executed.
+
+using conda::
+    conda create -n weldx -c conda-forge weldx
+    conda activate weldx
+    python -m ipykernel install --user --name weldx --display-name "Python (WeldX)"
+    conda create -n jlab -c conda-forge jupyter-lab
+    conda activate jlab
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager
+    jupyter labextension install k3d
```
CagtayFabry, line 64:
```suggestion
using conda::

    conda create -n weldx -c conda-forge weldx
    conda activate weldx
    python -m ipykernel install --user --name weldx --display-name "Python (WeldX)"
    conda create -n jlab -c conda-forge jupyter-lab
    conda activate jlab
    jupyter labextension install @jupyter-widgets/jupyterlab-manager
    jupyter labextension install k3d
```

CagtayFabry:
thank you for starting this 🚀 

we should also add the required windows fix for running kernels, I will update later

maybe also move it over to the top of the 'Getting Started' section instead of the main index?

marscher:
> thank you for starting this rocket
> 
> we should also add the required windows fix for running kernels, I will update later
> 
> maybe also move it over to the top of the 'Getting Started' section instead of the main index?

I think most projects have the installation as a separate topic to make it easier to spot.

CagtayFabry:
> > thank you for starting this rocket
> > we should also add the required windows fix for running kernels, I will update later
> > maybe also move it over to the top of the 'Getting Started' section instead of the main index?
> 
> I think most projects have the installation as a separate topic to make it easier to spot.

in that case I would move it behind the tutorials in the toctree since I find it a little too technical to be the first item

marscher:
I mean "Getting started" is the heading of our tutorial list. Eventually we should restructure that as well. As you suggested the installation guide could be a subtopic of the tutorials list, but then IMO it's hard to find. And first things first, if you do not have an installation, you cannot use the software (in theory we have binder, which is linked in the readme (first tutorial)).

vhirtham:
In my opinion, the installation is the first thing a user has to bother with, so it should either be the first item on the getting started page or at a premium position in the toctree. I don't think that an installation guide is too technical for someone who intends to use a python package. Also, it's not the case that the title is somewhat vague about what to expect when I click on it. So if I just want to see some features of the package, I'll simply skip it and move on with the next item. In any case, it's easy to change, so if we can't decide now, merge and change it later 😉

CagtayFabry (approved):
maybe take another look @marscher but LGTM 🚀

PR #456, opened 2021-07-30
[asdf] Do not copy array data

### LINKED ISSUE #454: xarray.Variable deserialization casts NDArrayType (asdf) to np.ndarray
As a consequence, memory mapping of stored arrays does not work, e.g. copy_arrays flag has no effect.

marscher:
## Changes

Because asdf does not handle time datatypes (correctly), there was a line of code within the serialization of xarray Variable to convert all ndarray like types from the asdf ndarray wrapper. This effectively prevents the use of the "copy_arrays" flag, since the
data got always copied.

## Related Issues

Closes #454

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/

### DIFF
--- a/weldx/asdf/tags/weldx/core/common_types.py
+++ b/weldx/asdf/tags/weldx/core/common_types.py
@@ -134,7 +134,12 @@ class VariableTypeASDF(WeldxType):
 
         """
         dtype = np.dtype(tree["dtype"])
-        data = np.array(tree["data"], dtype=dtype)
+        # TODO: it would be ideal, if asdf would handle time types natively.
+        if dtype.char in ("M", "m"):  # handle np.timedelta64 and np.datetime64
+            data = np.array(tree["data"], dtype=dtype)
+            # assert data.base is tree["data"]
+        else:
+            data = tree["data"]  # let asdf handle np arrays with its own wrapper.
         if "unit" in tree:  # convert to pint.Quantity
             data = Q_(data, tree["unit"])
 
--- a/weldx/asdf/tags/weldx/core/data_array.py
+++ b/weldx/asdf/tags/weldx/core/data_array.py
@@ -1,3 +1,4 @@
+from asdf.tags.core import NDArrayType
 from xarray import DataArray
 
 import weldx.asdf.tags.weldx.core.common_types as ct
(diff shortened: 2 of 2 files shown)


weldx/asdf/tags/weldx/core/common_types.py
```diff
@@ -134,7 +134,12 @@ def from_tree(cls, tree, ctx):
 
         """
         dtype = np.dtype(tree["dtype"])
-        data = np.array(tree["data"], dtype=dtype)
+        # TODO: it would be ideal, if asdf would handle time types natively.
```
CagtayFabry, line 137: cref https://github.com/asdf-format/asdf-standard/issues/270

CagtayFabry:
nice, does this work now with `copy_arrays`?

we should probably dig into the (asdf related) code for more `np.array()`, `np.asarray()` etc. calls

weldx/asdf/tags/weldx/core/common_types.py
marscher: Thanks for the reference (and welcome back). Is there any roadmap for asdf standard 2.0? I guess this has to be designed super carefully in order not breaking every bodies files.

marscher:
Our unit tests switch the copy_arrays flag at least. But I doubt that the underlying functionality of mem mapping is really being tested. We can argue that this is already tested extensively upstream.

CagtayFabry:
> Our unit tests switch the copy_arrays flag at least. But I doubt that the underlying functionality of mem mapping is really being tested. We can argue that this is already tested extensively upstream.

that is certainly not tested, we just use these parameters to catch any errors that might flare up
we would probably have to look into a memory profiler to see the difference

weldx/asdf/tags/weldx/core/common_types.py
CagtayFabry: haven't looked into the state of the asdf 2.0 standard for a while tbh

CagtayFabry (approved):

marscher:
I got my memory profiler booted up anyways for another PR, so eventually I'll take a look.

PR #459, opened 2021-08-02
WelDXfile compression/show_header fixes

### LINKED ISSUE #417: compression behavior with WeldxFile
When using `WeldxFile` a few inconsistencies can arise regarding the array compression behavior.

Because our custom classes/tags that do not support the lazy loading of `NDarrayType` will always load the `ndarray` data the compression associated with that `ndarray` will get reset, which can increase the filesize unexpectedly when using `WeldxFile`

The following example shows this when running in jupyterlab:
```python
from pathlib import Path

import numpy as np
import xarray as xr

import asdf
import weldx

filename = "demo.asdf"

with asdf.AsdfFile({"data": xr.DataArray(np.ones((1000, 1000)))}) as af:
    af.write_to(filename, all_array_compression="zlib")
    af.close()


print(Path(filename).stat().st_size)
wx_file = weldx.WeldxFile(filename, "rw")
print(Path(filename).stat().st_size)
display(wx_file)
print(Path(filename).stat().st_size)
wx_file.close()

# 12685
# 12685
# 8001014
```
a few things to note:
- opening with `"r"` prevents this issue (but gives the warning)
- only the call to `display` (the interactive json view) recreates the file contents, maybe we can also prevent this here?
- passing a pure `ndarray` in the tree doesn't show this issue (`{"data": np.ones((1000, 1000))}`)

Should be kept in mind when working with files containing large data. @vhirtham @marscher

marscher:
## Changes

This PR contains two changes. First it adds two arguments to deal with compression and memory mapping of arrays. The flags are just being passed through to asdf.
Second the show_asdf_header method does not write any blocks during serialization. We achieve this by setting the block manager to a mock object. Some tests assure that this does not have side effects on the internal asdf tree.

## Related Issues

Closes #417, 

## Checks

- [ ] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/


CagtayFabry:
cool solution ! 🚀

CagtayFabry (approved):

marscher:
Thanks, I also gave the compression suggestion of yours a shot, but wasn't able to enable it properly. But with the compression solution all blocks will be created and then compressed. So not storing the blocks in the first place is the way to go.

CagtayFabry:
> Thanks, I also gave the compression suggestion of yours a shot, but wasn't able to enable it properly. But with the compression solution all blocks will be created and then compressed. So not storing the blocks in the first place is the way to go.

good to know :)

PR #460, opened 2021-08-03
Add abstract base class for time dependent types

vhirtham:
## Changes

Adds the abstract base class `TimeDependent`

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] ~~updated tests~~
- [x] updated doc/

### DIFF
--- a/weldx/time.py
+++ b/weldx/time.py
@@ -498,10 +498,10 @@ class TimeDependent(ABC):
 
     @property
     @abstractmethod
-    def time(self) -> Union[Time, None]:
-        """Get the classes time component"""
+    def time(self) -> Time:
+        """Get the classes time component."""
 
     @property
     @abstractmethod
     def reference_time(self) -> Union[Timestamp, None]:
-        """Returns the reference timestamp if the time data is absolute."""
+        """Return the reference timestamp if the time data is absolute."""
--- a/weldx/time.py
+++ b/weldx/time.py
@@ -1,6 +1,7 @@
 """Contains classes and functions related to time."""
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 from functools import reduce
 from typing import List, Union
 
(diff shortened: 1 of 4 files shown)


weldx/time.py
```diff
@@ -490,3 +491,17 @@ def union(times=List[Union[types_time_like, "Time"]]) -> Time:
...
+
+    @property
+    @abstractmethod
+    def time(self) -> Union[Time, None]:
+        """Get the classes time component"""
+
+    @property
+    @abstractmethod
+    def reference_time(self) -> Union[Timestamp, None]:
+        """Returns the reference timestamp if the time data is absolute."""
```
vhirtham, line 507: Now every class that inherits from `TimeDependent` must implement those two methods (or any other we want to enforce). Otherwise we get the error:

~~~
TypeError: Can't instantiate abstract class SomeClass with abstract method reference_time
~~~

Unfortunately, the specified return type is not enforced. However, as soon as we get mypy up and running, it will complain about using other return types than the one given by the ABC:

~~~
error: Return type "str" of "time" incompatible with return type "Time" in supertype "TimeDependent"
~~~

vhirtham:
I made the LCS and the `TimeSeries` inherit from this class. Should we add any further functions? I think those two are enough for the start.

vhirtham:
> > I made the LCS and the `TimeSeries` inherit from this class. Should we add any further functions? I think those two are enough for the start.
> 
> the CSM should also follow this interface at some point imo

I thaught so too, but what is its returned `time`? The time union?

CagtayFabry:
> 
> 
> > > I made the LCS and the `TimeSeries` inherit from this class. Should we add any further functions? I think those two are enough for the start.
> > 
> > 
> > the CSM should also follow this interface at some point imo
> 
> I thaught so too, but what is its returned `time`? The time union?

I think that would be convenient, yes :)

CagtayFabry (approved):

vhirtham:
> > > > I made the LCS and the `TimeSeries` inherit from this class. Should we add any further functions? I think those two are enough for the start.
> > > 
> > > 
> > > the CSM should also follow this interface at some point imo
> > 
> > 
> > I thaught so too, but what is its returned `time`? The time union?
> 
> I think that would be convenient, yes :)

Not entirely sure about this, but we can do that in a future PR after discussing it. It's a trivial change. 😉

PR #463, opened 2021-08-04
add asdf `time/time` schema

CagtayFabry:
## Changes
adds an asdf schema for the `Time` class

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests

### DIFF
--- a/weldx/asdf/schemas/weldx.bam.de/weldx/time/time-1.0.0.yaml
+++ b/weldx/asdf/schemas/weldx.bam.de/weldx/time/time-1.0.0.yaml
@@ -12,34 +12,34 @@ examples:
     - A single timedelta of 2 seconds
     - |
       !<tag:weldx.bam.de:weldx/time/time-1.0.0>
-        values: !weldx!time/timedelta-1.0.0 {value: P0DT0H0M2S}
+        values: !<tag:weldx.bam.de:weldx/time/timedelta-1.0.0> {value: P0DT0H0M2S}
   -
     - A timedelta series from 0 s to 4 s with a 2 s frequency
     - |
       !<tag:weldx.bam.de:weldx/time/time-1.0.0>
-        values: !weldx!time/timedeltaindex-1.0.0
-            start: !weldx!time/timedelta-1.0.0 {value: P0DT0H0M0S}
-            end: !weldx!time/timedelta-1.0.0 {value: P0DT0H0M4S}
+        values: !<tag:weldx.bam.de:weldx/time/timedeltaindex-1.0.0>
+            start: !<tag:weldx.bam.de:weldx/time/timedelta-1.0.0> {value: P0DT0H0M0S}
+            end: !<tag:weldx.bam.de:weldx/time/timedelta-1.0.0> {value: P0DT0H0M4S}
             freq: 2S
-            min: !weldx!time/timedelta-1.0.0 {value: P0DT0H0M0S}
-            max: !weldx!time/timedelta-1.0.0 {value: P0DT0H0M4S}
+            min: !<tag:weldx.bam.de:weldx/time/timedelta-1.0.0> {value: P0DT0H0M0S}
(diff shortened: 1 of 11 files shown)


weldx/asdf/schemas/weldx.bam.de/weldx/time/time-1.0.0.yaml
```diff
@@ -0,0 +1,59 @@
...
+    type: object
+    oneOf:
+      - tag: "tag:weldx.bam.de:weldx/time/timedelta-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/timedeltaindex-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/timestamp-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/datetimeindex-1.*"
+
+  reference_time:
+    type: object
+    tag: "tag:weldx.bam.de:weldx/time/timestamp-1.*"
```
vhirtham, line 55: Should we add descriptions for the schema doc?

weldx/asdf/tags/weldx/time/time.py
```diff
@@ -0,0 +1,27 @@
...
+    name = "time/time"
+    version = "1.0.0"
+    types = [Time]
+
+    @classmethod
+    def to_tree(cls, node: Time, ctx):
+        """Serialize timedelta to tree."""
+        tree = dict()
+        tree["values"] = node.as_pandas()
+        tree["reference_time"] = node._time_ref
```
vhirtham, line 19:
```suggestion
        tree["values"] = node.as_timedelta()
        tree["reference_time"] = node.reference_time
```
I think it is cleaner this way. We used this structure before with our xarrays and there is no need to store absolute arrays if we have a reference time.

weldx/asdf/schemas/weldx.bam.de/weldx/time/time-1.0.0.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+type: object
+properties:
+  values:
+    type: object
+    oneOf:
+      - tag: "tag:weldx.bam.de:weldx/time/timedelta-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/timedeltaindex-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/timestamp-1.*"
+      - tag: "tag:weldx.bam.de:weldx/time/datetimeindex-1.*"
```
vhirtham, line 51:
```suggestion
      - tag: "tag:weldx.bam.de:weldx/time/timedelta-1.*"
      - tag: "tag:weldx.bam.de:weldx/time/timedeltaindex-1.*"
```
See my comment at the tag. If we change it, reevaluate the examples

weldx/asdf/tags/weldx/time/time.py
```diff
@@ -0,0 +1,27 @@
...
+        return tree
+
+    @classmethod
+    def from_tree(cls, tree, ctx):
+        """Construct Time from tree."""
+        values = tree["values"]
+        reference_time = tree.get("reference_time", None)
+        return Time(values, reference_time)
```
vhirtham, line 27:
```suggestion
        return Time(tree["values"], tree.get("reference_time"))
```

vhirtham (changes requested):

weldx/asdf/tags/weldx/time/time.py
CagtayFabry: true, however storing the datetime-values would enable direct comparison of the min/max values from the schema (for validation or others) which I kind of like even though we don't have these kind of validators implemented yet

One more problem I see with the above implementation (or the `as_timedelta` function in general) is that it would not reconstruct a `DatetimeIndex` into the same time object because of setting an explicit reference time.

```python
t1 = Time(["2020", "2021"])
t2 = Time(t1.as_timedelta(), t1.reference_time)

assert t1._time_ref is None
assert t2._time_ref is not None

t1.equals(t2) # False
```

Issue #297, opened 2021-03-23
Measurement.Data.data.data.data


marscher:
The data class data has the field data and name, but the since xarrays support the name attribute as well, I wonder why we have this kind of duplication? It makes code working with Measurements look strange, since we need to access data twice to obtain the actual xarray. Further if I want to access the raw data it gets even worse :)
We could just define a measurement to directly contain a "plain", although pretty high-level, xarray, right?

vhirtham:
I am currently rewriting the `MeasurementChain` related stuff in the context of #63. Removing this "construct" is also on the agenda... even though there is no such thing as too much data 😄

CagtayFabry:
yes it can get confusing quickly, I think we should try to stick to existing naming conventions if possible (so `.data` should only be an ndarray-like the way it is for xarray for example). We could find a more fitting attribute name for xarray objects as well

We should discuss a rework of the measurment module soon, maybe you can share your Ideas here or in #63 @vhirtham ?

vhirtham:
This should basically be solved with PR #326 . However, there are some occasions where we get a `data.data`, for example in the `TimeSeries`. This always occurs if a `data` function returns an xarray type and you want to have the raw data. Not sure if we want to go any further, but I think it should be okay.

vhirtham:
> maybe rename instances that access xarray objects to something else besides data like `dataarray`, `variable` or similar? and try to keep it consistent for different classes

Seems reasonable

CagtayFabry:
please review if still relevant after latest PRs @vhirtham

PR #468, opened 2021-08-05
Complete Time todos

vhirtham:
## Changes

-  [x] Use a [descriptor](https://docs.python.org/3/howto/descriptor.html) for the `union` method to use it as class and instance method. See [this link](https://stackoverflow.com/questions/28237955/same-name-for-classmethod-and-instancemethod/28238212#28238212)
- [x] Check docstrings if something is missing
- [x] Add test for `+` operator if `Time` instance represents an absolute time --- basically copy-paste the already existing test for a time delta `Time`, remove absolute types for the `other` object and tweak the expected results.
- [x] move time related utility functions to `time.py`
- [x] Test `-` operator
- [x] Test `union`
- [x] Reevaluate comments in PR #440
- [x] Reevaluate comments in PR #448

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] ~~updated CHANGELOG.md~~ Only changes that affect the ``Time`` class that is not yet released
- [x] updated tests
- [x] updated doc/
- [x] ~~update example/tutorial notebook~~

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Union
 from warnings import warn
 
-import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import pint
@@ -15,6 +14,7 @@ from weldx.constants import Q_
 from weldx.time import Time, TimeDependent, types_time_like
 
 if TYPE_CHECKING:
+    import matplotlib.pyplot
     import sympy
 
 __all__ = ["MathematicalExpression", "TimeSeries"]
@@ -290,8 +290,8 @@ class TimeSeries(TimeDependent):
             'step', 'linear'.
 
         """
-        self._data = None
-        self._time_var_name = None
+        self._data: Union[MathematicalExpression, xr.DataArray] = None
+        self._time_var_name = None  # type: str
         self._shape = None
         self._units = None
         self._interp_counter = 0
@@ -326,7 +326,7 @@ class TimeSeries(TimeDependent):
         if not isinstance(self.data, MathematicalExpression):
             if not isinstance(other.data, pint.Quantity):
                 return False
-            return self._data.identical(other.data_array)
+            return self._data.identical(other.data_array)  # type: ignore
 
         return self._data == other.data
 
@@ -616,11 +616,11 @@ class TimeSeries(TimeDependent):
     def plot(
         self,
         time: Union[pd.TimedeltaIndex, pint.Quantity] = None,
-        axes: plt.Axes = None,
+        axes: "matplotlib.pyplot.Axes" = None,
         data_name: str = "values",
         time_unit: Union[str, pint.Unit] = None,
         **mpl_kwargs,
-    ) -> plt.Axes:
+    ) -> "matplotlib.pyplot.Axes":
         """Plot the `TimeSeries`.
 
         Parameters
@@ -644,6 +644,8 @@ class TimeSeries(TimeDependent):
             The matplotlib axes object that was used for the plot
 
         """
+        import matplotlib.pyplot as plt
+
         if axes is None:
             _, axes = plt.subplots()
         if self.is_expression or time is not None:
@@ -655,7 +657,7 @@ class TimeSeries(TimeDependent):
         if time_unit is not None:
             time = time.to(time_unit)
 
-        axes.plot(time.m, self._data.data.m, **mpl_kwargs)
+        axes.plot(time.m, self._data.data.m, **mpl_kwargs)  # type: ignore
         axes.set_xlabel(f"t in {time.u:~}")
         y_unit_label = ""
         if self.units not in ["", "dimensionless"]:
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -12,8 +12,7 @@ import xarray as xr
 
 import weldx.util as ut
 from weldx.constants import Q_
-from weldx.time import Time, TimeDependent
-from weldx.types import types_time_like
+from weldx.time import Time, TimeDependent, types_time_like
 
 if TYPE_CHECKING:
     import sympy
@@ -270,7 +269,7 @@ class TimeSeries(TimeDependent):
     def __init__(
         self,
         data: Union[pint.Quantity, MathematicalExpression],
-        time: Union[types_time_like, Time] = None,
+        time: types_time_like = None,
         interpolation: str = None,
         reference_time: pd.Timestamp = None,
     ):
@@ -368,7 +367,7 @@ class TimeSeries(TimeDependent):
     def _initialize_discrete(
         self,
         data: Union[pint.Quantity, xr.DataArray],
-        time: Union[types_time_like, Time] = None,
+        time: types_time_like = None,
         interpolation: str = None,
     ):
         """Initialize the internal data with discrete values."""
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -364,6 +364,16 @@ class TimeSeries(TimeDependent):
         if not isinstance(data_array.data, pint.Quantity):
             raise TypeError("The data of the 'DataArray' must be a 'pint.Quantity'.")
 
+    @staticmethod
+    def _create_data_array(
+        data: Union[pint.Quantity, xr.DataArray], time: Time
+    ) -> xr.DataArray:
+        return (
+            xr.DataArray(data=data)
+            .rename({"dim_0": "time"})
+            .assign_coords({"time": time.as_timedelta_index()})
+        )
+
     def _initialize_discrete(
         self,
         data: Union[pint.Quantity, xr.DataArray],
@@ -379,6 +389,7 @@ class TimeSeries(TimeDependent):
             self._check_data_array(data)
             data = data.transpose("time", ...)
             self._data = data
+            # todo: set _reference_time?
         else:
             # expand dim for scalar input
             data = Q_(data)
@@ -388,22 +399,10 @@ class TimeSeries(TimeDependent):
             # constant value case
             if time is None:
                 time = pd.Timedelta(0)
-
             time = Time(time)
 
-            if time.is_absolute:
-                self._reference_time = time.reference_time
-                time = time - time.reference_time
-
-            time = time.as_pandas_index()
-
-            if not isinstance(time, pd.TimedeltaIndex):
-                raise ValueError(
-                    '"time" must be a time quantity or a "pandas.TimedeltaIndex".'
-                )
-
-            dax = xr.DataArray(data=data)
-            self._data = dax.rename({"dim_0": "time"}).assign_coords({"time": time})
+            self._reference_time = time.reference_time
+            self._data = self._create_data_array(data, time)
         self.interpolation = interpolation
 
     def _init_expression(self, data):
@@ -473,10 +472,7 @@ class TimeSeries(TimeDependent):
         if not np.iterable(data):  # make sure quantity is not scalar value
             data = np.expand_dims(data, 0)
 
-        dax = xr.DataArray(data=data)  # don't know exact dimensions so far
-        return dax.rename({"dim_0": "time"}).assign_coords(
-            {"time": time.as_pandas_index()}
-        )
+        return self._create_data_array(data, time)
 
     @property
     def data(self) -> Union[pint.Quantity, MathematicalExpression]:
--- a/weldx/tests/test_time.py
+++ b/weldx/tests/test_time.py
@@ -695,7 +695,8 @@ class TestTime:
             ([TDI([1, 5]), TDI([2, 6, 7]), TDI([1, 3, 7])], TDI([1, 2, 3, 5, 6, 7])),
         ],
     )
-    def test_union(list_of_objects, time_exp):
+    @pytest.mark.parametrize("test_instance", [True, False])
+    def test_union(test_instance, list_of_objects, time_exp):
         """Test input types for Time.union function.
 
         Parameters
@@ -706,4 +707,8 @@ class TestTime:
             Expected result time
 
         """
-        assert np.all(Time.union(list_of_objects) == time_exp)
+        if test_instance:
+            instance = Time(list_of_objects[0])
+            assert np.all(instance.union(list_of_objects[1:]) == time_exp)
+        else:
+            assert np.all(Time.union(list_of_objects) == time_exp)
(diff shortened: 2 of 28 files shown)


weldx/transformations/types.py
```diff
@@ -1,25 +1,15 @@
...
-types_timeindex = Union[pd.DatetimeIndex, pd.TimedeltaIndex, pint.Quantity]
-types_time_and_lcs = Union[
-    types_timeindex, "weldx.transformations.LocalCoordinateSystem"
-]
-
 
+# remove ???
 @runtime_checkable
 class SupportsTime(Protocol):
     """An ABC with one abstract method time."""
```
vhirtham, line 15: Can this be removed? I found no usages and I guess it was supposed to do the same thing as `TimeDependent`

CagtayFabry: I guess that is not needed any longer

vhirtham:
Skimmed over `weldx.util` and all remaining functions that involve time are related to xarray (if I have not missed something) and should stay in `weldx.util`

weldx/core.py
```diff
@@ -365,10 +364,20 @@ def _check_data_array(data_array: xr.DataArray):
...
             raise TypeError("The data of the 'DataArray' must be a 'pint.Quantity'.")
 
+    @staticmethod
+    def _create_data_array(
+        data: Union[pint.Quantity, xr.DataArray], time: Time
+    ) -> xr.DataArray:
+        return (
+            xr.DataArray(data=data)
+            .rename({"dim_0": "time"})
+            .assign_coords({"time": time.as_timedelta_index()})
```
CagtayFabry, line 374: In cases like these I would really like if we could attach the reference time to the xarray coordinates still

```suggestion
            xr.DataArray(data=data)
            .rename({"dim_0": "time"})
            .assign_coords({"time": time.as_data_array().weldx.time_ref_restore()})
```
(thinking about it the `.weldx.time_ref_restore()` call should probably be part of `as_data_array()`)

For the `TimeSeries` `reference_time` could then also point to the stored attribute instead of `self._reference_time`

vhirtham: Can we do this in a separate PR? I already have a list of todos for the `TimeSeries` because of its new support for absolute times. I will open a new issue related to this soon and we can add it as a to-do there.

CagtayFabry: sure

weldx/time.py
```diff
@@ -492,29 +498,67 @@ def _convert_other(time) -> Union[pd.TimedeltaIndex, pd.DatetimeIndex]:
...
+    >>>
+    >>> all(Time.union([t1, t2]) == Time(["1s", "2s", "3s", "4s", "5s"]))
+    True
+
+    Using the instance method:
+
+    >>> all(t1.union([t2]) == Time(["1s", "2s", "3s", "4s", "5s"]))
+    True
+
+    """
```
vhirtham, line 547: I tested the examples in another docstring and they run. Docs are also rendering well, but we should keep this in mind.

weldx/core.py
CagtayFabry: see https://github.com/BAMWelDX/weldx/issues/471

weldx/tests/test_time.py
```diff
@@ -375,6 +374,229 @@ def test_add_timedelta(
...
+            otherwise
+        time_class_is_array :
+            If `True`, the `Time` instance contains 3 time values and 1 otherwise
+        other_is_array :
+            If `True`, the other time object contains 3 time values and 1 otherwise
+
+        """
+        # skip array cases where the type does not support arrays
+        if other_type in [Timedelta, Timestamp] and other_is_array:
+            pytest.skip()
```
CagtayFabry, line 417: the `pytest.skip()` command immediately jumps out of the function here, right?
In that case I prefer to simply `return` and reserve `skip` for cases where the whole test case did not start for a cleaner pytest output

weldx/time.py
```diff
@@ -155,6 +163,30 @@ class 'TimeDependent' are supported too.
     >>> t_res = "2000-01-01" + Time(["1d", "2d"])
     >>> t_res = ["3s", "4s"] + Time(["1s", "2s"])
 
+    Subtraction is possible too, but there are some restrictions. It is not possible
+    to subtract an absolute time from a time delta. Additionally, since the values of
+    a ``Time`` instance must be monotonically increasing, any subtraction that
+    produces a result that doesn't fulfill this requirement will fail. This is always
+    the case when subtracting arrays from scalars because either the array that is
+    subtracted (it is intrenally cast to a ``Time`` instance) or the resulting array
```
CagtayFabry, line 171:
```suggestion
    subtracted (it is internally cast to a `Time` instance) or the resulting array
```
just wondering, does the linking work here?

weldx/time.py
```diff
@@ -280,7 +342,7 @@ def __iter__(self):
         """Use generator to iterate over index values."""
         return (t for t in self.as_pandas_index())
 
-    def __getitem__(self, item):
+    def __getitem__(self, item) -> Union[Timedelta, Timestamp]:
```
CagtayFabry, line 345: I think slicing like `[2:2:-1]` should also work and result in an `Index` type?

CagtayFabry (changes requested):
A few thoughts on the naming of the `as_*` methods I think we should consider:

Personally I would prefer to reserve these methods to cases where the exact object can be restored again, so that `t.equals(Time(t.as_*))` always passes.

For some of these functions I think it would be more fitting to be properties, namely `Time.timestamp` (instead of `Time.as_timestamp()`) and `Time.timedeltaindex` (instead of `Time.as_timedeltaindex()`)
also `Time.quantity` could return the quantity as default in `"s"`

maybe we should also introduce a `.time` method here as an alternative to the lengthy `as_pandas_index()`?

weldx/time.py
vhirtham: Have not tested it. Is that the case?

vhirtham:
> Personally I would prefer to reserve these methods to cases where the exact object can be restored again, so that `t.equals(Time(t.as_*))` always passes.

Not sure if I understand your intention. Do you mean cases where we have a single return type and not a `Union`?
 
> For some of these functions I think it would be more fitting to be properties, namely `Time.timestamp` (instead of `Time.as_timestamp()`) and `Time.timedeltaindex` (instead of `Time.as_timedeltaindex()`)

We can do this. I have no real opinion on what I like more. 

> also `Time.quantity` could return the quantity as default in `"s"`

The unit of the quantity doesn't really matter since it is just a predefined representation of the stored duration that can be cast to any other representation afterward. Okay, there might occur some floating-point issues on very large and very small numbers, but I guess seconds is a safe choice for our scope. 😉 

> maybe we should also introduce a `.time` method here as an alternative to the lengthy `as_pandas_index()`?

Not so fond of that one. `Time.as_pandas_index()` is rather self explanatory. But what is `Time.time()` supposed to do (if you do not know already). Also, I am having a [deja vu](https://github.com/BAMWelDX/weldx/issues/297). I know that it would never be called on the class itself, but maybe on an instance called `time`. Anyhow, I prefer medium-sized, more distinct names over short names if the latter ones do not provide enough information.

CagtayFabry:
> > Personally I would prefer to reserve these methods to cases where the exact object can be restored again, so that `t.equals(Time(t.as_*))` always passes.
> 
> Not sure if I understand your intention. Do you mean cases where we have a single return type and not a `Union`?
> 

My intention is that I would prefer not to loose any information for the output of a method named `as_*` which is currently the case since we loose the reference timestamp when using `as_timedeltaindex()` (hence the above code returns false)
I think this should pass for these kind of methods: `assert t.equals(Time(t.as_*))`

> > For some of these functions I think it would be more fitting to be properties, namely `Time.timestamp` (instead of `Time.as_timestamp()`) and `Time.timedeltaindex` (instead of `Time.as_timedeltaindex()`)
> 
> We can do this. I have no real opinion on what I like more.
> 
> > also `Time.quantity` could return the quantity as default in `"s"`
> 
> The unit of the quantity doesn't really matter since it is just a predefined representation of the stored duration that can be cast to any other representation afterward. Okay, there might occur some floating-point issues on very large and very small numbers, but I guess seconds is a safe choice for our scope. 😉
> 

It's just we cant have parameters for properties so 

> > maybe we should also introduce a `.time` ~~method~~ **property** here as an alternative to the lengthy `as_pandas_index()`?
> 
> Not so fond of that one. `Time.as_pandas_index()` is rather self explanatory. But what is `Time.time()` supposed to do (if you do not know already). Also, I am having a [deja vu](https://github.com/BAMWelDX/weldx/issues/297). I know that it would never be called on the class itself, but maybe on an instance called `time`. Anyhow, I prefer medium-sized, more distinct names over short names if the latter ones do not provide enough information.

I meant to write property there
We can call it `.index` if that is more fitting

weldx/tests/test_time.py
CagtayFabry: test it please and update accordingly

weldx/time.py
CagtayFabry: test it please and update accordingly

vhirtham:
~~~ python
from weldx import Time

t_delta = Time(["1s", "3s", "4s", "5s"])
t_abs = Time(["2000", "2001", "2002", "2003"])

print(type(t_delta[2:2:-1]))
print(type(t_abs[2:2:-1]))
~~~

output

~~~
<class 'pandas.core.indexes.timedeltas.TimedeltaIndex'>
<class 'pandas.core.indexes.datetimes.DatetimeIndex'>
~~~

good finding 👍

vhirtham:
> My intention is that I would prefer not to loose any information for the output of a method named `as_*` which is currently the case since we loose the reference timestamp when using `as_timedeltaindex()` (hence the above code returns false)
> I think this should pass for these kind of methods: `assert t.equals(Time(t.as_*))`

After having a day to think about it, I think we should leave it as is because the current implementation is more user-friendly. If I want to know all the possible types, I can cast to, I just type `.as_` and look what the autocompletion offers. Additionally, all casting methods are nicely grouped together in the docs. If we apply the proposed changes, this wouldn't be the case anymore.
Differentiating between "cast properties" and methods might also be a bit confusing. Sometimes I have to add `()` and sometimes not. In the docs they are also put into 2 different categories. 

I think that interface consistency/accessibility is more important here. If I cast to a time delta, I should be aware that I might lose information like if I cast a `float` to an `int`. We can also put a hint into the docstring that warns about possible data loss.

> It's just we cant have parameters for properties so pick a default one for that but keep the regular method with a parameter to choose the scaling unit

I am not so sure if it makes sense to have a property that is a specialization of an existing method and just saves some letters and parenthesis when calling it.

CagtayFabry:
> 
> 
> > My intention is that I would prefer not to loose any information for the output of a method named `as_*` which is currently the case since we loose the reference timestamp when using `as_timedeltaindex()` (hence the above code returns false)
> > I think this should pass for these kind of methods: `assert t.equals(Time(t.as_*))`
> 
> After having a day to think about it, I think we should leave it as is because the current implementation is more user-friendly. If I want to know all the possible types, I can cast to, I just type `.as_` and look what the autocompletion offers. Additionally, all casting methods are nicely grouped together in the docs. If we apply the proposed changes, this wouldn't be the case anymore.
> Differentiating between "cast properties" and methods might also be a bit confusing. Sometimes I have to add `()` and sometimes not. In the docs they are also put into 2 different categories.
> 
> I think that interface consistency/accessibility is more important here. If I cast to a time delta, I should be aware that I might lose information like if I cast a `float` to an `int`. We can also put a hint into the docstring that warns about possible data loss.
> 
> > It's just we cant have parameters for properties so pick a default one for that but keep the regular method with a parameter to choose the scaling unit
> 
> I am not so sure if it makes sense to have a property that is a specialization of an existing method and just saves some letters and parenthesis when calling it.

sure, we don't have to remove anything. Maybe we can have both? :)
Do you mind if I add some of these as properties as well?

vhirtham:
> > > My intention is that I would prefer not to loose any information for the output of a method named `as_*` which is currently the case since we loose the reference timestamp when using `as_timedeltaindex()` (hence the above code returns false)
> > > I think this should pass for these kind of methods: `assert t.equals(Time(t.as_*))`
> > 
> > 
> > After having a day to think about it, I think we should leave it as is because the current implementation is more user-friendly. If I want to know all the possible types, I can cast to, I just type `.as_` and look what the autocompletion offers. Additionally, all casting methods are nicely grouped together in the docs. If we apply the proposed changes, this wouldn't be the case anymore.
> > Differentiating between "cast properties" and methods might also be a bit confusing. Sometimes I have to add `()` and sometimes not. In the docs they are also put into 2 different categories.
> > I think that interface consistency/accessibility is more important here. If I cast to a time delta, I should be aware that I might lose information like if I cast a `float` to an `int`. We can also put a hint into the docstring that warns about possible data loss.
> > > It's just we cant have parameters for properties so pick a default one for that but keep the regular method with a parameter to choose the scaling unit
> > 
> > 
> > I am not so sure if it makes sense to have a property that is a specialization of an existing method and just saves some letters and parenthesis when calling it.
> 
> sure, we don't have to remove anything. Maybe we can have both? :)
> Do you mind if I add some of these as properties as well?

Feel free to add anything you think we need. 😉

CagtayFabry:
I'm pretty sure the conda build failures are not related to this PR @vhirtham , see #474 
So I think we can merge this?

CagtayFabry (approved):

weldx/transformations/local_cs.py
```diff
@@ -450,9 +450,9 @@ def from_axis_vectors(
         y: types_coordinates = None,
         z: types_coordinates = None,
         coordinates: types_coordinates = None,
-        time: types_time_like = None,
-        time_ref: types_timestamp_like = None,
-    ):
+        time: Union[types_time_like, Time] = None,
+        time_ref: Union[types_timestamp_like, Time] = None,
+    ) -> LocalCoordinateSystem:
```
vhirtham, line 455: `Time` should be a part of those definitions. Looked into the definitions and saw, that I somehow missed putting it into `types_time_like` but it is already in types_timestamp_like. See here: 
https://github.com/BAMWelDX/weldx/blob/8a40edb2b3e7469c79e083636e25fcd45df5ac9f/weldx/time.py#L675

vhirtham: Just realized, that `Time` should already be included in `types_time_like` since it is composed of other type collections

CagtayFabry: oh I see, thanks
this somehow looked weird after merging master into this branch

PR #469, opened 2021-08-06
Dummy inline arrays in asdf.util and weldxfile

marscher:
## Changes
* The dummy block manager impl was replaced with a much simpler patching of the block creating function inside asdf. The replaced function does not allocate a block, but replaces the array with an empty list. With the additional asdf write option ``all_array_storage="internal"`` we achieve zero additional output memory for array serialization. This is being checked with a test.
* The display_header function takes a (deep) copy of the Asdf tree, but without copying numpy arrays. This approach avoids side effects due manipulating the tree internally.
* Added pytest options to use ``WeldxFile`` inside the read/write_buffer functions used in the testsuite. This increases the testing coverage of the class. Added a GH actions job to check this new configuration on Linux, py3.8 soley.

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests


marscher:
osx is still fucked up (most likely because it is a fucked up os). Linux and Windows do no need more than 10% memory now during serialization. I've patched the copy behavior of ndarrays, because asdf always deepcopies the tree.

CagtayFabry:
nice, really liking this solution
can we skip this specific test only for osx?

vhirtham:
> osx is still fucked up (most likely because it is a fucked up os). Linux and Windows do no need more than 10% memory now during serialization. I've patched the copy behavior of ndarrays, because asdf always deepcopies the tree.

~~~ python
        before = get_mem_info()
        fn = tempfile.mktemp(suffix=".wx", dir=tmpdir)
        with WeldxFile(mode=mode) as fh:
            fh["x"] = large_array
            fh.show_asdf_header(use_widgets=False, _interactive=False)
            fh.write_to(fn)

        after = get_mem_info()
~~~

Didn't really try to understand the whole test, but can't we move the second line out of the memcheck block? If we only want to check the memory footprint of `show_asdf_header`, we should remove as many possibly allocating commands as possible. Don't know what `tempfile` does under the hood, but maybe the fail on osx is not caused by `show_asdf_header` but one of the other involved lines. Just roughly followed this one, so sorry if I missed something.

> can we skip this specific test only for osx?

I would want to understand why it is failing before disabling it. Disabling a failing test just lets me ask: why do we have the test in the first place if it isn't that important.

marscher:
@vhirtham tempfile checks for an unused filename in $TEMPDIR and returns it. But the writeto method could involve larger memory footprints (e.g. buffers are kept in the background for whatever reason). I followed your advice and measure only the memory during show_asdf_header. And you are right about the test philosophy :) Still, osx sucks!

vhirtham:
> @vhirtham tempfile checks for an unused filename in $TEMPDIR and returns it. But the writeto method could involve larger memory footprints (e.g. buffers are kept in the background for whatever reason). I followed your advice and measure only the memory during show_asdf_header. And you are right about the test philosophy :) Still, osx sucks!

Glad to hear that it worked. 😉

CagtayFabry:
In case there are still issues with osx I prefer to mark the test as xfail on that platform for now, see https://docs.pytest.org/en/latest/how-to/skipping.html#condition-parameter

CagtayFabry:
I have changed the non interactive header output to just show the yaml string again for testing @marscher e29052f763cc64712b276ebb8385f0e70a631f9e

marscher:
> I have changed the non interactive header output to just show the yaml string again for testing @marscher [e29052f](https://github.com/BAMWelDX/weldx/commit/e29052f763cc64712b276ebb8385f0e70a631f9e)

I do not understand why pytest stdout capturing does not work for the print statement in `asdf.info()`. I also thought that the cluttering of the test output was quite annoying. But we should not change the implementation/desired output just because the test tool has a hick up, right?

CagtayFabry:
> > I have changed the non interactive header output to just show the yaml string again for testing @marscher [e29052f](https://github.com/BAMWelDX/weldx/commit/e29052f763cc64712b276ebb8385f0e70a631f9e)
> 
> I do not understand why pytest stdout capturing does not work for the print statement in `asdf.info()`. I also thought that the cluttering of the test output was quite annoying. But we should not change the implementation/desired output just because the test tool has a hick up, right?

there were some (probably unrelated issues)
however looking into it I actually like the pure yaml output also in a console because it shows the tag definitions etc and all nested elements. In addition `asdf.info` seems to do some kind of tag resolving to classes which is not really needed and imo can complicate things
However since the change is very simple we can revert back anytime?

CagtayFabry:
On another note can you explain the idea behind the `_USE_WELDX_FILE` flag?

marscher:
> 
> 
> On another note can you explain the idea behind the `_USE_WELDX_FILE` flag?

The debug utility functions read_buffer and write_buffer will use WeldxFile internally, if the flag is set. Since these util functions are used extensively in the test suite, it increases the testing coverage of WeldxFile. There are two pytest flags to toggle this flag. I've also added a single pytest job to the matrix (just on Linux) to run this configuration.

.github/workflows/pytest.yml
```diff
@@ -77,6 +77,11 @@ jobs:
           pytest -n 2 --dist loadfile --nbval --current-env --cov-append tutorials/
           echo "Exited with '$?'"
 
+      - name: test with WeldxFile
+        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
+        run: |
+          pytest -n2 --dist loadfile --weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header
```
CagtayFabry, line 83:
```suggestion
          pytest ./weldx/tests/asdf_tests -n2 --dist loadfile --weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header
```

I think it is enough to run in this subfolder since that is basically where everything asdf related happens
preivious coverage was 89% in this run let's see
https://github.com/BAMWelDX/weldx/pull/469/checks?check_run_id=3375408820#step:11:68

.github/workflows/pytest.yml
```diff
@@ -77,6 +77,11 @@ jobs:
           pytest -n 2 --dist loadfile --nbval --current-env --cov-append tutorials/
           echo "Exited with '$?'"
 
+      - name: test with WeldxFile
+        if: (matrix.os == 'ubuntu-latest') && (matrix.py == '3.8')
+        run: |
+          pytest ./weldx/tests/asdf_tests -n2 --dist loadfile --weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header
```
CagtayFabry, line 83:
```suggestion
          pytest -n2 --dist loadfile --weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header --cov-append ./weldx/tests/asdf_tests
```
probably append coverage here ...

CagtayFabry (approved):

.github/workflows/pytest.yml
marscher: Good hint! Thank you

PR #470, opened 2021-08-09
replaced attrdict with custom impl.

marscher:
## Changes

We do not depend on the unmaintained attrdict lib any more. Replaced the impl with a simple recursive internal dict patching solution. As the interface did not change, I kindly refuse to write a changelog entry :)


CagtayFabry (approved):
thanks, that looks short and elegant

please list the change in dependencies in the changelog anyway so it is easier to spot for the conde-forge release 😉

marscher:
sure, will do.

PR #473, opened 2021-08-14
Fix warnings

CagtayFabry:
## Changes
- Fixes array conversion warnings from tests
- small cleanup
- don't allow k3d==2.10.0 because of conda dependencies
- use plain yaml str output for `_show_non_interactive` instead of `asdf.info`

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -7,7 +7,6 @@ test_patterns = [
 
 exclude_patterns = [
   "doc/**",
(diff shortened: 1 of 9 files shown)


CagtayFabry:
for reference:
the following error was encountered on conda builds due to a missing `msgpack-python` dependency in `k3d==2.10` on conda-forge

https://github.com/BAMWelDX/weldx/runs/3328719961#step:7:1729

PR #476, opened 2021-08-14
csm time interpolation range

### LINKED ISSUE #289: CSM interpolation outside of time range
When interpolating between systems with any time dependency along the path, the result will always be time dependent even if the transformation is known to be static because no "real" time dependencies exist (or can exists) for the interpolation timeline.

Here is an example:
```python
import weldx
from weldx import Q_, CoordinateSystemManager, LocalCoordinateSystem

csm = CoordinateSystemManager("base")
# add A as time dependent in base
csm.add_cs("A","base",lcs=LocalCoordinateSystem(coordinates=[[0,0,0],[1,1,1]],time=Q_([0,1],"s")))
# add B as static in A
csm.add_cs("B","A",lcs=LocalCoordinateSystem())

# this is always static because no time dependencies are between A and B
print(csm.get_cs("B","A",time=Q_([2,3,4],"s")).time)
>>> None

# this is time dependent even though the result is known to be static
print(csm.get_cs("B","base",time=Q_([2,3,4],"s")).time)
>>> array([2000000000, 3000000000, 4000000000], dtype='timedelta64[ns]')

print(csm.get_cs("B","base",time=Q_([2,3,4],"s")).coordinates.data)
>>> array([[1., 1., 1.],
>>>        [1., 1., 1.],
>>>        [1., 1., 1.]])
```

This really only is an issue when considering large time interpolations.
I guess we can check in a form similar to (not actual code)
```python
if np.all(time<=csm.time_union.min()) or np.all(time>=csm.time_union.max()):
    # revert to single timestamp
```

CagtayFabry:
## Changes

_Describe changes in this PR_

## Related Issues

Closes #289

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


CagtayFabry:
it probably makes sense to implement this in the `get_cs_on_edge` method

weldx/tests/test_transformations.py
```diff
@@ -851,17 +851,17 @@ def test_reset_reference_time_exceptions(
...
             ),
             (  # broadcast right
                 TS("2020-02-10"),
-                TDI([29, 30], "D"),
+                TDI([22, 29, 30], "D"),
                 TS("2020-02-10"),
-                r_mat_z([0.5, 0.5]),
-                np.array([[3, 1, 2], [3, 1, 2]]),
+                r_mat_z([0.5, 0.5, 0.5]),
+                np.array([[3, 1, 2], [3, 1, 2], [3, 1, 2]]),
```
vhirtham, line 864: I guess, that it does make sense to collapse edge cases (the only time intersection is at the boundaries) to a static system too since we still have always the same value

vhirtham: Implemented this behavior now.

vhirtham:
Need to do some cleanup and update the docstrings of the CSM and LCS methods to reflect the new behavior, but apart from that this should be ready. You can already have a look. I am not sure if there is a way to merge some of the conditions in `LCS.interp_time` but I think there isn't much room for optimizations because coordinates and orientations are pretty independent of each other. Maybe we can put their individual interpolations into small internal functions, but that would be just code cosmetics.

CagtayFabry:
I took a quick look and I think the implementation looks good (and pretty readable now with the `Time` class), no need to micro manage there I guess 👍 

however for some of these larger `if` branches and longer functions I sometimes find it easier to have a few inline comments about the code (just for readability), I'll add an example

weldx/transformations/local_cs.py
```diff
@@ -687,18 +692,36 @@ def interp_time(
...
-        orientation = ut.xr_interp_orientation_in_time(self.orientation, time)
+        # calculate orientations
+        if self.orientation.ndim == 2:
+            orientation = self.orientation
+        elif time.max() <= self.time.min():
+            orientation = self.orientation.values[0]
+        elif time.min() >= self.time.max():
+            orientation = self.orientation.values[-1]
+        else:
+            orientation = ut.xr_interp_orientation_in_time(self.orientation, time)
```
CagtayFabry, line 703:
```suggestion
        # consider interpolation time with regard of LCS time for interpolation
        if self.orientation.ndim == 2:  # don't interpolate static
            orientation = self.orientation
        elif time.max() <= self.time.min():  # only use edge timestamp
            orientation = self.orientation.values[0]
        elif time.min() >= self.time.max():  # only use edge timestamp
            orientation = self.orientation.values[-1]
        else:  # full interpolation with overlapping times
            orientation = ut.xr_interp_orientation_in_time(self.orientation, time)
```

what do you think @vhirtham ?
Comments like this sometimes help me get back into the code

CagtayFabry:
eventually we could also implement stuff like 'partially' overlapping times, so `LCS.time = [0,1,2,3,4,5]` and `interp_time=[3,4,5,6,7,8,9]` would give back `LCS.time=[3,4,5]` (I hope that makes sens 😉 )

weldx/transformations/local_cs.py
vhirtham: Added the comments.  I gave the coordinate and orientation interpolations their own private methods to get the code a bit less crowded

vhirtham:
I streamlined the behavior of `interp_time` so that it will also return a static system if only a single time value is passed. Before we had a time-dependent system with a single value. Now the time and the extra array dimension are stripped in that case.

> eventually we could also implement stuff like 'partially' overlapping times, so `LCS.time = [0,1,2,3,4,5]` and `interp_time=[3,4,5,6,7,8,9]` would give back `LCS.time=[3,4,5]` (I hope that makes sens 😉 )

I think this will make stuff too complicated. It might be easy to implement in the LCS, but it will be a nightmare in the CSM. We would have to check for this case separately and broadcast the missing values for the multiplication. But this would also require figuring out how many values were stripped on either side of the returned array. I would leave it as is.

CagtayFabry:
> I streamlined the behavior of `interp_time` so that it will also return a static system if only a single time value is passed. Before we had a time-dependent system with a single value. Now the time and the extra array dimension are stripped in that case.
> 
Maybe we can keep the time as a dimension but not coordinate of the dataarray, I'll take a look

> > eventually we could also implement stuff like 'partially' overlapping times, so `LCS.time = [0,1,2,3,4,5]` and `interp_time=[3,4,5,6,7,8,9]` would give back `LCS.time=[3,4,5]` (I hope that makes sens 😉 )
> 
> I think this will make stuff too complicated. It might be easy to implement in the LCS, but it will be a nightmare in the CSM. We would have to check for this case separately and broadcast the missing values for the multiplication. But this would also require figuring out how many values were stripped on either side of the returned array. I would leave it as is.
agreed

vhirtham:
> Maybe we can keep the time as a dimension but not coordinate of the dataarray, I'll take a look

I don't think that this makes sense.

CagtayFabry:
> > Maybe we can keep the time as a dimension but not coordinate of the dataarray, I'll take a look
> 
> I don't think that this makes sense.

Thought it could be nice to keep some time information but also not really needed right now I guess

weldx/transformations/local_cs.py
```diff
@@ -547,6 +547,11 @@ def is_time_dependent(self) -> bool:
         """
         return self.time is not None or self._coord_ts is not None
 
+    @property
+    def has_timeseries(self) -> bool:
+        """Return `True` if the coordinate system has a `TimeSeries` component."""
+        return self._coord_ts is not None
```
CagtayFabry, line 553:
```suggestion
        return isinstance(self.coordinates, TimeSeries) or isinstance(self.orientation, TimeSeries)
```
I guess this would make sense?

vhirtham:
@CagtayFabry Funny thing: I can't require your approval because you created the PR, but I can approve myself 😉 So are we good to go?

vhirtham (approved):

CagtayFabry:
> 
> 
> > > Maybe we can keep the time as a dimension but not coordinate of the dataarray, I'll take a look
> > 
> > 
> > I don't think that this makes sense.
> 
> Thought it could be nice to keep some time information but also not really needed right now I guess

looked into it a little and I think it would be good to differentiate between xarray `coords` and `dimensions` more in our codebase (currently we mostly use coords which is probably not what we actually want)

I'll see about it in a separate PR

PR #484, opened 2021-08-18
Update gh actions

### LINKED ISSUE #481: filter action runs
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

CagtayFabry:
## Changes
My idea is do get the following:
- skip `pytest`, `pytest_asdf` & `static_analysis` workflows if no changes in `weldx/**` (or the workflow file itself)
- ONLY run the full pytest matrix on a 'ready for review PR'
  The full matrix will be started in other circumstances as well but all slow steps (pip install, pytest) on win & osx will simply be skipped. So the tests report running but finish after 10s or so.
  this also skips these matrix elements on fork branches (there, only the main pytest on ubuntu will run)
- skip the conda/pypi build CI for changes only in `CHANGELOG.md` or `doc/**`

One thing that is not so great is that the checks will get triggered again if you switch between draft and open PRs (which we usually don't do)

I also removed the os/python matrix where we only had a single entry.

## Related Issues

Closes #481

### DIFF
--- a/.github/workflows/pytest_asdf.yml
+++ b/.github/workflows/pytest_asdf.yml
@@ -1,14 +1,10 @@
 name: pytest asdf
 on:
   push:
-    paths:
-      - 'weldx/**'
-      - '.github/workflows/pytest_asdf.yml'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
-    paths:
-      - 'weldx/**'
-      - '.github/workflows/pytest_asdf.yml'
+  # Run every Monday at 6am UTC
+  schedule:
+    - cron: '0 6 * * 1'
 
 jobs:
   pytest:
--- a/.github/workflows/pytest_asdf.yml
+++ b/.github/workflows/pytest_asdf.yml
@@ -1,5 +1,13 @@
 name: pytest asdf
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - 'weldx/**'
+      - '.github/workflows/pytest_asdf.yaml'
+  pull_request:
+    paths:
+      - 'weldx/**'
+      - '.github/workflows/pytest_asdf.yaml'
 
 jobs:
   pytest:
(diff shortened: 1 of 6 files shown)


.github/workflows/pytest_asdf.yml
```diff
@@ -1,19 +1,24 @@
...
+on:
+  push:
+    paths:
+      - 'weldx/**'
+      - '.github/workflows/pytest_asdf.yml'
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'weldx/**'
+      - '.github/workflows/pytest_asdf.yml'
```
vhirtham, line 11:
```suggestion
on:
  push:
    paths:
      - 'weldx/**'
      - 'tutorials/**'
      - '.github/workflows/pytest_asdf.yml'
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
    paths:
      - 'weldx/**'
      - 'tutorials/**'
      - '.github/workflows/pytest_asdf.yml'
```

Since our tutorials are also tested on Ubuntu (3.8?) we should include that directory. Maybe it is possible to include it only on that specific run

vhirtham (changes requested):
I think all actions should always run on the master branch, regardless if something changed or not.

.github/workflows/pytest_asdf.yml
CagtayFabry: added that for pytest

CagtayFabry:
> I think all actions should always run on the master branch, regardless if something changed or not.

I don't think this configuration exists, but in that case it is probably overkill to micro manager most of the jobs so I will just let everything run on push besides pytest (manually filtering steps there)

CagtayFabry:
also added a weekly schedule run for the main jobs, let's see if that works

vhirtham (approved):

PR #485, opened 2021-08-18
Fix warnings

vhirtham:
## Changes

Fixes some uncaptured warnings


setup.cfg
```diff
@@ -107,6 +107,8 @@ asdf_schema_root = weldx/asdf/schemas
 asdf_schema_skip_tests =
     weldx.bam.de/weldx/datamodels/single_pass_weld-1.0.0.schema.yaml
 norecursedirs = doc
+filterwarnings =
+    ignore:Sentinel is not a public part of the traitlets API.:DeprecationWarning
```
CagtayFabry, line 111: doesn't seem to work here? :(

setup.cfg
```diff
@@ -107,6 +107,8 @@ asdf_schema_root = weldx/asdf/schemas
 asdf_schema_skip_tests =
     weldx.bam.de/weldx/datamodels/single_pass_weld-1.0.0.schema.yaml
 norecursedirs = doc
+filterwarnings =
+    ignore::DeprecationWarning
```
vhirtham, line 111: It's annoying, I tried several solutions found in the docs and Stack Overflow. Nothing, worked to further limit the filter. Maybe I ask a question on SO tomorrow for this specific warning. Maybe they have an idea how to solve it.

CagtayFabry: If it is not something we can fix in our own code or that we can pinpoint down to a specific warning then we should just leave it in there, right now it's pretty clean

setup.cfg
```diff
@@ -107,6 +107,8 @@ asdf_schema_root = weldx/schemas
 #asdf_schema_skip_tests =
 #    weldx.bam.de/weldx/datamodels/single_pass_weld-1.0.0.schema.yaml
 norecursedirs = doc
+filterwarnings =
+    ignore::DeprecationWarning:traittypes.*:
```
CagtayFabry, line 111: This syntax only filters all `DeprecationWarning`s from the `traittypes` module/namespace @vhirtham

CagtayFabry (approved):

setup.cfg
vhirtham: Good that you got it working. I tried this syntax too, but with the wrong module (`traitlets`)

setup.cfg
vhirtham: Since you already solved it, this isn't important anymore, but the good thing about having all warnings gone is that you will notice new ones much faster and some of them should probably be fixed fast. We now get green result messages on GitHub and also locally. If that changes, we have a simple visual hint to check for new warnings 😃

PR #486, opened 2021-08-18
Refactor interpolation

CagtayFabry:
## Changes
I have reworked the xarray handling a little to now work with xarray `dimensions` instead of `coordinates` which is probably what we actually want. Mentioned here https://github.com/BAMWelDX/weldx/pull/476#issuecomment-900312142

It is now possible to keep e.g. a single timestamp as an xarray coordinate in a constant `orientations` with `shape=[3,3]`

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/


weldx/transformations/local_cs.py
```diff
@@ -683,18 +677,15 @@ def _interp_time_coordinates(self, time: Time) -> Union[xr.DataArray, np.ndarray
...
-        elif self.coordinates.ndim == 1:  # don't interpolate static
-            coordinates = self.coordinates
-        elif time.max() <= self.time.min():  # only use edge timestamp
-            coordinates = self.coordinates.values[0]
-        elif time.min() >= self.time.max():  # only use edge timestamp
-            coordinates = self.coordinates.values[-1]
-        else:  # full interpolation with overlapping times
-            coordinates = ut.xr_interp_coordinates_in_time(self.coordinates, time)
-
-        if len(coordinates) == 1:  # remove "time dimension" for single value cases
```
CagtayFabry, line 695: The `len(coordinates) == 1` check can be pretty unstable imo @vhirtham 
at first I was surprised this works without even knowing what it does. Using `len(coordinates.time) == 1` here would probably be more explicit.

weldx/util.py
```diff
@@ -919,41 +919,43 @@ def xr_interp_orientation_in_time(
...
     # resync and reset to correct format
     if time_ref:
-        dsx_out.weldx.time_ref = time_ref
-    dsx_out = dsx_out.weldx.time_ref_restore()
+        da.weldx.time_ref = time_ref
+    da = da.weldx.time_ref_restore().transpose(..., "time", "c", "v")
 
-    return dsx_out.transpose(..., "c", "v")
+    if len(da.time) == 1:  # remove "time dimension" for static case
+        return da.isel({"time": 0})
```
CagtayFabry, line 956: This is the replacement we should use instead of `return coordinates[0].values` or even `return coordinates[0]`, explicitly naming the dimension

weldx/transformations/local_cs.py
```diff
@@ -664,22 +662,18 @@ def as_rotation(self) -> Rot:  # pragma: no cover
...
-            orientation = self.orientation.values[-1]
-        else:  # full interpolation with overlapping times
-            orientation = ut.xr_interp_orientation_in_time(self.orientation, time)
-
-        if len(orientation) == 1:  # remove "time dimension" for single value cases
-            return orientation[0].values
-        return orientation
-
-    def _interp_time_coordinates(self, time: Time) -> Union[xr.DataArray, np.ndarray]:
+        if "time" not in self.orientation.dims:  # don't interpolate static
```
CagtayFabry, line 667: this is probably also a safer way to check for 'static' systems

weldx/transformations/local_cs.py
vhirtham: Well, this is why you find this specific line below the processing of the inputs. `coordinates` and `orientations` should already be formatted correctly to match the specific internal structure I was abusing here. So it is pretty stable I guess unless you change the order of instructions in the `__init__`. But in this case, we have a ton of tests that will fail.

Not sure about `len(coordinates.time) == 1`. You might need to check the existence of `coordinates.time` first since static `coordinates` shouldn't have this attribute. Checking the passed `time` classes length shouldn't help much either since it gives no indication if `coordinates` actually have this extra dimension. I guess we can get it more self-explaining but at the price of additional checks. Not sure if it is worth it for class internal code that already, has a comment to state what it does.

weldx/util.py
vhirtham: I agreed with you in the case of this generalized function. However, as the internal structure in the LCS was well defined, accessing per index was safe and probably faster (no lookups).

weldx/transformations/local_cs.py
vhirtham: True, but also more expensive. As already laid out in my other comments, we are dealing with a predefined internal structure that will most likely never change. So basically, we are just trading a bit of performance for a bit of syntactic sugar. For safety, we have our tests. I see no need for generalizations on a specialized internal data structure.

vhirtham:
Hope my comments do not sound/read too harshly. It is not meant to be harsh. Just not sure if the internal changes on the LCS are really that necessary (except for the renamings). We can apply them, but I am not convinced that they give us an advantage.

CagtayFabry:
> Hope my comments do not sound/read too harshly. It is not meant to be harsh. Just not sure if the internal changes on the LCS are really that necessary (except for the renamings). We can apply them, but I am not convinced that they are advantageous here.

The thing is that without using explicit `xarray` syntax we basically prevent ourselves from ever having any dimensions besides `time`, `c`, `v` in our dataarrays (like a length dimension/coordinate along the path)

as for performance, I would really like to see a benchmark first for these tiny changes to consider this 😉

weldx/transformations/local_cs.py
CagtayFabry: > Well, this is why you find this specific line below the processing of the inputs. `coordinates` and `orientations` should already be formatted correctly to match the specific internal structure I was abusing here. So it is pretty stable I guess unless you change the order of instructions in the `__init__`. But in this case, we have a ton of tests that will fail.
> 

I don't think the current `__init__` prevents passing additional dimensions in coordinate dimensions for example?

> Not sure about `len(coordinates.time) == 1`. You might need to check the existence of `coordinates.time` first since static `coordinates` shouldn't have this attribute. Checking the passed `time` classes length shouldn't help much either since it gives no indication if `coordinates` actually have this extra dimension. I guess we can get it more self-explaining but at the price of additional checks. Not sure if it is worth it for class internal code that already, has a comment to state what it does.

wouldn't `len(coordinates) == 1` also fail here if there was no time dimension?
`len(LocalCoordinateSystem().coordinates) == 3`

vhirtham:
> > Hope my comments do not sound/read too harshly. It is not meant to be harsh. Just not sure if the internal changes on the LCS are really that necessary (except for the renamings). We can apply them, but I am not convinced that they are advantageous here.
> 
> The thing is that without using explicit `xarray` syntax we basically prevent ourselves from ever having any dimensions besides `time`, `c`, `v` in our dataarrays (like a length dimension/coordinate along the path)

I am just talking about the LCS internals. I don't think it will ever need more than those 3 dimensions. But nevermind. As I said, we can do it.

> as for performance, I would really like to see a benchmark first for these tiny changes to consider this 😉

No need for a benchmark. Direct access is always the fastest data access method you can have, but it won't matter much since its contribution to the total execution time of our scripts, it is probably irrelevant. 😉

weldx/transformations/local_cs.py
vhirtham: > I don't think the current `__init__` prevents passing additional dimensions in coordinate dimensions for example?

Just checked `_build_coordinates` and it seems that you are correct. Currently, a `DataArray` will be just passed through if I didn't miss something. But I consider this as a bug that should be fixed. The LCS shouldn't store any data that is not necessary to do its job

> wouldn't `len(coordinates) == 1` also fail here if there was no time dimension?
> `len(LocalCoordinateSystem().coordinates) == 3`

No, because the trick here is, that the outer dimension of static coordinates is always of length 3. So if it differs from 3, we have a time dimension and we only want to remove it, if it is of length 1.

vhirtham (approved):

PR #490, opened 2021-08-20
Refactor utility module

### LINKED ISSUE #480: Clean up utility package
- remove all obsolete functions (functions that are sorely used by the `geometry` package can be moved)
- split into multiple sub packages (`util.util`, `util.xarray`)

vhirtham:
## Changes

Separates xarray related utility functions from other utility functions and removes no longer needed functions.

## Related Issues

Closes #480

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


vhirtham:
Looking at the remaining functions in the `util` module, I think it might make sense to mark it as private. Apart from some xarray functions, most of the stuff is not of any interest to the user. Even the xarray functions probably shouldn't be used outside of the package. What do you think @CagtayFabry @marscher 

In case we want to keep it public, we need to discuss how the doc pages should look like. Currently, you get to an intermediate site that lists both submodules. I think this is okay this way to further categorize the functions and classes.

CagtayFabry (approved):
> Looking at the remaining functions in the `util` module, I think it might make sense to mark it as private. Apart from some xarray functions, most of the stuff is not of any interest to the user. Even the xarray functions probably shouldn't be used outside of the package. What do you think @CagtayFabry @marscher
> 
> In case we want to keep it public, we need to discuss how the doc pages should look like. Currently, you get to an intermediate site that lists both submodules. I think this is okay this way to further categorize the functions and classes.

I actually like it the way it is now. 👍 
Since the utility functions are in their own module  now and don't clutter anything I like having them in the documentation :)
Especially for the xarray functions since they probably become much more handy when working more with spatial data or the new geometry stuff later. (we should probably flesh out the docstrings some more though)

vhirtham:
> > Looking at the remaining functions in the `util` module, I think it might make sense to mark it as private. Apart from some xarray functions, most of the stuff is not of any interest to the user. Even the xarray functions probably shouldn't be used outside of the package. What do you think @CagtayFabry @marscher
> > In case we want to keep it public, we need to discuss how the doc pages should look like. Currently, you get to an intermediate site that lists both submodules. I think this is okay this way to further categorize the functions and classes.
> 
> I actually like it the way it is now. 👍
> Since the utility functions are in their own module now and don't clutter anything I like having them in the documentation :)
> Especially for the xarray functions since they probably become much more handy when working more with spatial data or the new geometry stuff later. (we should probably flesh out the docstrings some more though)

I think that all the functionality they provide should be wrapped by `SpatialData` and the corresponding `geometry` functions and classes. But we can leave them public for now and reevaluate later.

PR #497, opened 2021-08-26
CSM serialization

### LINKED ISSUE #354: CSM graph serialization

vhirtham:
## Changes

The asdf schema of the CSM now uses the `DiGraph` schema. This change is incompatible with older WelDX files.

## Related Issues

Closes #354

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/

## ToDo

- [x] consider spatial data
- [x] consider sub systems
- [x] add legacy tags


weldx/tags/core/transformations/coordinate_system_hierarchy.py
```diff
@@ -92,277 +9,27 @@ class CoordinateSystemManagerConverter(WeldxConverter):
...
-                        break
-                if not lcs_added:
-                    lcs_data_list_next += [lcs_data]
-
-            leaf_nodes = leaf_nodes_next
-            lcs_data_list = lcs_data_list_next
+    # todo
+    #  - check possible bug: WeldxFile does not clear old yaml tree when writing. It
+    #    just overwrites the bytes it needs. Write large tree first, then remove stuff
+    #    and rewrite
```
vhirtham, line 15: Reminder create issue

weldx/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
```diff
@@ -13,59 +13,59 @@ examples:
   -
     - An abstract example of a coordinate system hierarchy with nested subsystems and without data
     - |
-      !weldx!core/transformations/coordinate_system_hierarchy-1.0.0
+      !<asdf://weldx.bam.de/weldx/tags/core/transformations/coordinate_system_hierarchy-1.0.0>
         name: CSM
```
vhirtham: This is somewhat strange. The removed tags were stored this way in the file generated with the script I posted earlier. Had to replace them for the asdf tests to work. However, the normal reading of a file is not affected.

weldx/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
```diff
@@ -10,114 +10,59 @@ description: |
...
-    description: |
-      Name of the root coordinate system
-    type: string
-
   reference_time:
-    description: |
-      A timestamp that serves as reference point for all attached coordinate systems.
     oneOf:
       - tag: "asdf://weldx.bam.de/weldx/tags/time/timestamp-1.*"
-      - tag: "tag:weldx.bam.de:weldx/time/timestamp-1.*" # legacy_tag
```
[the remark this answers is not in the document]
vhirtham: So I left them away, if I got you righi here.

weldx/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
CagtayFabry: This is because the weldx extension sets the custom YAML tag handle in the asdf file header when creating files, see this PR: https://github.com/asdf-format/asdf/pull/963:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
%TAG !weldx! asdf://weldx.bam.de/weldx/tags/
--- !core/asdf-1.1.0

```
Our handle is defined here https://github.com/BAMWelDX/weldx/blob/b15007fef82d7b750d8ecb14e61b7579bddb4e9b/weldx/asdf/extension.py#L58

However, the YAML tag handle context is not present when `sphinx-asdf` evaluates the schema examples and cannot be set either (I am not sure how the asdf default `!` handle gets 'injected' but we would probably have to make a change to `sphinx-asdf`)

I have listed a regex to replace the `!weldx!` tags here: https://github.com/BAMWelDX/weldx/discussions/479

vhirtham: good to know.

weldx/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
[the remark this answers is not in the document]
CagtayFabry: since the layout is now completely different and there are (cant be) any legacy files with this structure we can stick to the new tags
It would be nice however to still be able to read the old weldx files (and convert them afterwards)

CagtayFabry:
I am still thinking about the best way to support the old 0.4 weldx files

My best idea is to include the legacy code in separate files and add a new converter for the old tag that can read old files and create CSM instances
Since we only need to read old files using the legacy code we can omit the `to_yaml_tree` section of the old code

CagtayFabry:
> I am still thinking about the best way to support the old 0.4 weldx files
> 
> My best idea is to include the legacy code in separate files and add a new converter for the old tag that can read old files and create CSM instances
> Since we only need to read old files using the legacy code we can omit the `to_yaml_tree` section of the old code

I have implemented this approach here https://github.com/vhirtham/weldx/pull/3 and it seems to work for the IIW example 🎉 

you can load the old file and create a new version file upon save

vhirtham:
> > I am still thinking about the best way to support the old 0.4 weldx files
> > My best idea is to include the legacy code in separate files and add a new converter for the old tag that can read old files and create CSM instances
> > Since we only need to read old files using the legacy code we can omit the `to_yaml_tree` section of the old code
> 
> I have implemented this approach here [vhirtham#3](https://github.com/vhirtham/weldx/pull/3) and it seems to work for the IIW example 🎉
> 
> you can load the old file and create a new version file upon save

If it works we can just write a small script that reads all WelDX files in a directory and stores the updated WelDX files in a new one. We should then convert all important files and get rid of the legacy code ASAP.

CagtayFabry:
> > > I am still thinking about the best way to support the old 0.4 weldx files
> > > My best idea is to include the legacy code in separate files and add a new converter for the old tag that can read old files and create CSM instances
> > > Since we only need to read old files using the legacy code we can omit the `to_yaml_tree` section of the old code
> > 
> > 
> > I have implemented this approach here [vhirtham#3](https://github.com/vhirtham/weldx/pull/3) and it seems to work for the IIW example 🎉
> > you can load the old file and create a new version file upon save
> 
> If it works we can just write a small script that reads all WelDX files in a directory and stores the updated WelDX files in a new one. We should then convert all important files and get rid of the legacy code ASAP.

My idea would be to keep full legacy support for the `0.5` release and drop the legacy file/tag release in `0.6`

This way we can still read files with the old tags and provide 'proper' backwards compatibility with the new API from `0.6` on

weldx/tags/core/transformations/coordinate_system_hierarchy.py
vhirtham: done: Issue #516

vhirtham:
> My idea would be to keep full legacy support for the `0.5` release and drop the legacy file/tag support in `0.6`
> 
> This way we can still read (and convert) files with the old tags and provide 'proper' backwards compatibility with the new API from `0.6` on

Seems plausible. So just merge your updates into this branch. Are we good to go then (apart from this weird new k3d error?) or is there something missing?

CagtayFabry:
> Seems plausible. So just merge your updates into this branch. Are we good to go then (apart from this weird new k3d error?) or is there something missing?

I think everything should be ready
we should also prepare a PR on the IIW AA example repo that works with the current version imo (and update the example file to the new graph schema)

weldx/schemas/weldx.bam.de/weldx/core/transformations/coordinate_system_hierarchy-1.0.0.yaml
```diff
@@ -9,115 +9,191 @@ description: |
...
+            edges:
+            - !<asdf://weldx.bam.de/weldx/tags/core/graph/di_edge-1.0.0>
+              direction: bwd
+              attributes:
+                defined: true
+                transformation: !<asdf://weldx.bam.de/weldx/tags/core/transformations/local_coordinate_system-1.0.0> {}
+              target_node: !<asdf://weldx.bam.de/weldx/tags/core/graph/di_node-1.0.0>
+                name: A
+                attributes:
+                  data: {}
```
CagtayFabry, line 32: I guess the example is created using the debug code:
```python

csm = CoordinateSystemManager("R", "CSM")
csm.create_cs("A", "R")
csm.create_cs("B", "R")
csm.create_cs("C", "B")
```
looking at the csm plot output in a notebook the arrows point `R->A`, `R->B` etc.
In the schema should that not result in a `fwd` direction of the edges or am I mixing things up here?

CagtayFabry (approved):

PR #501, opened 2021-08-30
compare sets in compare_nested

CagtayFabry:
## Changes
compare sets in nested data structure using `__eq__`
This does not traverse (nested) elements in sets

## Checks

- [x] updated tests


vhirtham (approved):

marscher:
Why does it not traverse children in the set? Ain't this a deal breaker of the function name?

IMO this should pass as well:
``` python
    def test_nested_sets(self):
        a = dict(x={("bar", (1, 2)), 1, 2, 3})
        b = dict(x={("baz", (1, 2, 3)), 1, 2, 3})
        assert not compare_nested(a, b)
```

weldx/util/util.py
```diff
@@ -291,23 +297,13 @@ def compare_nested(
...
-            raise TypeError("either a or b are not a nested data structure.")
-
         # we bind the input structures a, b to the visit function.
         visit = functools.partial(_EqCompareNested._visit, a=a, b=b)
 
         try:
-            iterutils.remap(a, visit=visit, reraise_visit=True)
+            iterutils.remap(
+                a, visit=visit, reraise_visit=True, enter=_EqCompareNested._enter
+            )
```
CagtayFabry, line 306: the `_enter` function should probably have been used here before, which confused me a little

marscher: I guess you're right. I haven't looked into the code yet.

marscher (approved):

CagtayFabry:
> Why does it not traverse children in the set? Ain't this a deal breaker of the function name?
> 
> IMO this should pass as well:
> 
> ```python
>     def test_nested_sets(self):
>         a = dict(x={("bar", (1, 2)), 1, 2, 3})
>         b = dict(x={("baz", (1, 2, 3)), 1, 2, 3})
>         assert not compare_nested(a, b)
> ```

The problem is how to correctly 'traverse' two nested sets in parallel (which maybe cannot even be sorted, see the updated example)
The implementation now simply calls `==` when encountering a set, which is fine by me. We can list that behavior in the docstring.

Currently there is no example of nested sets in any of our datastructures afaik, since this is simply an internal utility function that is enough imo. We can adapt the function if we need to in the future

weldx/util/util.py
CagtayFabry: the behavior should be more consistent overall now
`set` comparison behavior is now listed under `compare_funcs` which prevents entering/traversing the items

weldx/tests/test_utility.py
```diff
@@ -400,13 +400,8 @@ def test_compare_nested_raise(a, b):  # noqa: D102
         argvalues=[
             ((1, 2, 3), [1, 2, 3], True),
             ((1, 2, 3), [1, 2, 0], False),
-            ({1, 2, 3}, {3, 2, 1}, True),
-            ({1, 2, 3}, [3, 2, 1], False),
-            ({1, 2, 3}, {1, 2, 0}, False),
-            ({1, 2, 3}, [1, 2, 3], True),
-            ([1, 2, 3], {1, 2, 3}, True),
```
CagtayFabry, line 407: Since sets are no longer considered a "nested" datastructure here `compare_nested` cannot be used on sets directly

marscher (approved):

weldx/util/util.py
```diff
@@ -271,7 +277,9 @@ def compare_nested(
         """Deeply compares [nested] data structures combined of tuples, lists, dicts...
 
         Also compares non-nested data-structures.
-        Arrays are compared using np.all and xr.DataArray.identical
+        Arrays are compared using np.all and xr.DataArray.identical.
+        Sets are compared by calling ``==`` and not traversed, hence the order of all
+        set items is important.
```
vhirtham, line 282: How is the order important or which order do you mean? `{"a", "b"} == {"b", "a"}` is `True`.

weldx/util/util.py
```diff
@@ -291,28 +299,18 @@ def compare_nested(
...
-            iterutils.remap(a, visit=visit, reraise_visit=True)
+            iterutils.remap(
+                a, visit=visit, reraise_visit=True, enter=_EqCompareNested._enter
+            )
         # Key not found in b, values not equal, more elements in a than in b
         except (KeyError, RuntimeError, IndexError):
             return False
         except TypeError:
-            raise TypeError("either a or b are not a nested data structure.")
+            raise TypeError("One of a or b is not a nested data structure (or a set).")
```
vhirtham, line 313: If we use sets as `a` and `b`, shouldn't the function just return `a==b` instead of raising an exception?

weldx/util/util.py
CagtayFabry: The order only really becomes important if the set contains some complex/nested elements (basically when you cannot apply sorting to the set items), see the `test_nested_sets` test case example

weldx/util/util.py
CagtayFabry: it would be nice but the way it is implemented using `remap` now `set` is not considered a nested structure (by the `enter` function) so you cannot apply it here

In the end I guess we have basic functionality for sets that appear in nested structures which hopefully covers most of our use cases

PR #503, opened 2021-08-31
delegate Time function calls to pandas.Index

CagtayFabry:
## Changes
This PR delegates function calls on `Time` objects to the underlying 'pandas.Index' if they are not implemented on the `Time` class itself.
This conveniently exposes the core pandas index API to the `Time` class without the need for custom implementations, e.g. one can simply call `Time(["1s","2s"]).max()`, `Time(["1s","2s"]).shape` instead of `Time(["1s","2s"]).as_pandas_index().max()`, `Time(["1s","2s"]).as_pandas_index().shape`

I have excluded propagation of 'dunder' methods because it could break some basic implementations (like `copy.copy()`)

- try delegating `Time` method calls to underlying `pandas.Index`
- support `Time` in compare nested
- remove custom implementation of `Time.max()`, `Time.min()`
- remove `Time.length` (use `len()` instead)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/ *(need to update the docstring to mention the available pandas API)*


CagtayFabry:
what do you think about this idea @vhirtham  ?

One thing that also came to my attention:
for `__getitem__` we are currently returning the pandas 'slice', maybe it would be more in line here to return at `Time` object like so

```python
    def __getitem__(self, item) -> types_pandas_times:
        """Access pandas index."""
        return Time(self.as_pandas_index()[item], self.reference__time)
```

vhirtham:
> what do you think about this idea @vhirtham ?
> 
> One thing that also came to my attention:
> for `__getitem__` we are currently returning the pandas 'slice', maybe it would be more in line here to return at `Time` object like so
> 
> ```python
>     def __getitem__(self, item) -> types_pandas_times:
>         """Access pandas index."""
>         return Time(self.as_pandas_index()[item], self.reference__time)
> ```

In my opinion, reducing interfaces to always return the same type is a really good thing. So I'd say go for it.

weldx/time.py
```diff
@@ -228,24 +228,20 @@ class 'TimeDependent' are supported too.
...
     >>> t = Time(["2000", "2001", "2002"])
-    >>> t[1] == Timestamp("2001")
-    True
-
-    >>> isinstance(t[1], Timestamp)
-    True
+    >>> t[1]
+    Time:
+    2001-01-01 00:00:00
+    reference time: 2000-01-01 00:00:00
```
vhirtham, line 241: Just curious: Have you created those examples with the reprexlite?

vhirtham (approved):

weldx/time.py
CagtayFabry: no I just copied the 'expected' form the doctest error 😉 

but reprexlite makes it super easy, just get rid of `#>` and add the `>>>`, here is the original output
```python
from weldx import Time

t = Time(["2000", "2001", "2002"])
t[1]
#> Time:
#> 2001-01-01 00:00:00
#> reference time: 2000-01-01 00:00:00
```

vhirtham: > no I just copied the 'expected' form the doctest error 😉

Ah... a "cheater", I see 😄

PR #504, opened 2021-09-06
WeldxFile copy interface

### LINKED ISSUE #493: WeldxFile copy has ambigious interface
Currently the we derive the class from UserDict to give a dictionary like usage pattern. The copy method just returns a copy of the internally mapped dictionary, but not the file itself. I think a user of this class would expect the latter. So we should discuss some cases here.

1. Copy the whole file (copy data and file)
2. Copy only the internal data (leave things as they are)
3. Break the interface of copy by passing an optional file argument

For 1. we have the additional case of an existing filename. So we can automatically put a prefix like "copy of $fn" to it or raise an Exception.

Option 2 has the previously discussed downside of having unexpected behavior.

Option 3 sounds most reasonable to me at the moment.

marscher:
## Changes

copy creates a `WeldxFile` instance optionally with a new filename. If the optional filename already exists, it depends on the overwrite flag, if the `FileExistsError` is raised or not.

## Related Issues

Closes #493

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests

### DIFF
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -17,7 +17,6 @@ from jsonschema import ValidationError
 
 from weldx.types import SupportsFileReadWrite, types_file_like, types_path_and_file_like
 from weldx.util import inherit_docstrings
-
 from .util import get_schema_path, get_yaml_header, view_tree
 
 __all__ = [
@@ -340,7 +339,7 @@ class WeldxFile(UserDict):
             Optional software used to make the change.
 
         Notes
-        --------
+        -----
         The software entry will be inferred from the constructor or, if not defined,
         from ``software_history_entry``.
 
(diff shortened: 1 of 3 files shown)


marscher:
I think we have agreed to add optional argument to the `copy()` method. However deepsource does not like. I guess we can safely ignore this one. Should I research for the ignore code? :)

CagtayFabry:
> I think we have agreed to add optional argument to the `copy()` method. However deepsource does not like. I guess we can safely ignore this one. Should I research for the ignore code? :)

I think codacy only looks for changed code in PRs right? We can leave it as is imo

weldx/asdf/file.py
```diff
@@ -429,18 +473,16 @@ def write_to(
...
-            created = True
 
         # if no args are given, we use the specifications given in the constructor.
         if not write_args:
             write_args = self._write_kwargs
 
         self._asdf_handle.write_to(fd, **write_args)
 
-        if created:
+        if isinstance(fd, types_file_like.__args__):
```
CagtayFabry, line 485: interesting!
(I suppose the `.__args__` is needed here?)

marscher: yes, it is kind of ugly, but works as expected. I mean what is the use of all of this type annotation stuff, if we could not use it for runtime type checking in the first place :)

PR #506, opened 2021-09-06
pin k3d

### LINKED ISSUE #505: [conda-build] k3d 2.11.0 requires jupyterlab, which is not installed.
The conda package of k3d now requires jupyterlab. I think we should optionally depend on k3d and raise an informative exception in the CSM plot method, if k3d is demanded as visualization method, but the package is not in place.
This way we can reduce the number of installed dependencies.

The approach of fixing the conda build by pinning to !=2.10 does not seem to work anymore. https://github.com/BAMWelDX/weldx/pull/473

marscher:
## Changes

This pins k3d to versions smaller 2.10.

## Related Issues

Closes #505

### DIFF
--- a/devtools/conda.recipe/meta.yaml
+++ b/devtools/conda.recipe/meta.yaml
@@ -31,6 +31,7 @@ test:
   requires:
     - pytest
     - pytest-xdist
+    - jupyterlab
   files:
     - conftest.py
   imports:
--- a/devtools/conda.recipe/meta.yaml
+++ b/devtools/conda.recipe/meta.yaml
@@ -31,7 +31,6 @@ test:
   requires:
     - pytest
     - pytest-xdist
-    - jupyterlab
   files:
     - conftest.py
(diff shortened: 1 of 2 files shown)


CagtayFabry:
I suggest setting `k3d <=2.9` for now since 2.11 brought other issues
https://github.com/K3D-tools/K3D-jupyter/blob/102bdbfde6058285439b589caef6a2f936e90986/setup.py#L83

marscher:
> I suggest setting `k3d <=2.9` for now since 2.11 brought other issues
> https://github.com/K3D-tools/K3D-jupyter/blob/102bdbfde6058285439b589caef6a2f936e90986/setup.py#L83

That's unfortunate. So many dependency issues. I have pinned it like you suggested.

CagtayFabry (approved):

CagtayFabry:
thanks, let's wait and see :)

PR #507, opened 2021-09-06
Use micromamba in Github workflows

marscher:
## Changes

the miniconda action takes almost 3 minutes to setup mamba itself. So we switch to a preinstalled version of mamba.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
           # Increase this value to reset cache if ./devtools/conda.recipe/build_env.yml has not changed
           CACHE_NUMBER: 1
         with:
-          path: ~/conda_pkgs_dir
+          path: ~/micromamba/pkgs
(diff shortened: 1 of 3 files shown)


CagtayFabry:
since the setup with readthedocs and conda-forge is now quite stable I think we can also remove the final 3 steps of the build process (`set package file env`, `actions/upload-artifact@v2`, `upload releases to dev label on Anaconda Cloud`) for the conda build workflow @marscher 👍

marscher:
nice, that would simplify things a lot.

marscher:
The rtd_environment.yaml installs weldx in development mode via a relative path. However this does not seem to work within the Github action, because the working directory is outside of the source/working copy (but in the $HOME/micromamba/envs dir). I've opened up an issue upstream.

CagtayFabry (approved):

CagtayFabry:
setup seems pretty fast now 🚀

marscher:
The only thing missing for a fast doc building is to use multiple sphinx jobs #508

CagtayFabry:
wondering why the pytest run is missing here 🤔

PR #514, opened 2021-09-13
support legacy tag in process schema

CagtayFabry:
## Changes
add missing old tag support

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,4 +11,5 @@ Closes # (add issue numbers)
 - [ ] updated CHANGELOG.md
 - [ ] updated tests
 - [ ] updated doc/
-- [ ] update example/tutorial notebooks 
+- [ ] update example/tutorial notebooks
+- [ ] update manifest file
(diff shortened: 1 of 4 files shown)


marscher:
May I ask why this pattern is only needed for the timestamp tag?

CagtayFabry:
> May I ask why this pattern is only needed for the timestamp tag?

sure, do you mean the `wx_property_tag` here? Or the `oneOf` construct?

The `oneOf` I simply forgot here during the change to the new ASDF API because I was only looking for `tag` validation and not `wx_property_tag`

The `wx_property_tag` is not really needed specifically for `timestamp` but kind of a little custom extension of the ASDF `tag` validator.
With the `tag` validator you validate that a specific property (that you have named in the schema) is 'tagged' with a specific tag
The `wx_property_tag` validator get's attached to a yaml/ASDF `object` (instead of a `property` for `tag`) and makes sure that all properties of that object validate against the given tag. The property names are not restricted in this case.
(In Python terms you could think 'make sure that all `values` in this `dict` (that I am serializing) are instances of a specific class)

PR #518, opened 2021-09-14
Pint xarray

CagtayFabry:
## Changes
- use `pint-xarray` in for interpolation in `xr_interp_like`, see https://github.com/xarray-contrib/pint-xarray

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests


CagtayFabry:
I will reopen this later from the current master

weldx/constants.py
```diff
@@ -63,6 +64,9 @@
 """
 
 
+# set default unit registry for pint-xarray
+pint.set_application_registry(WELDX_UNIT_REGISTRY)
```
CagtayFabry, line 68: I got an error saying the unit registries are not the same when previously using 
```python
pint_xarray.unit_registry = WELDX_UNIT_REGISTRY
```

I will have to look more into the documentation here but this approach now seems to work as intended @vhirtham

CagtayFabry:
I still need to add some test for different combinations of coordinates (units/time-types) and expand some docstrings

however in general this should be ready as an initial implementation that should work with units in `xr_interp_like`

vhirtham (approved):

PR #519, opened 2021-09-14
set pint default unit format to short notation

CagtayFabry:
## Changes
use `pint` short notation for units as default

I encountered some problems with serialization using short notation (not sure why), so explicitly keeping the long notation there

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/constants.py
+++ b/weldx/constants.py
@@ -14,6 +14,8 @@ WELDX_UNIT_REGISTRY = _ureg(
 WELDX_UNIT_REGISTRY.define("percent = 0.01*count = %")
 # swap plank constant for hour definition
 WELDX_UNIT_REGISTRY.define("hour = 60*minute = h = hr")
+# set default string format to short notation
+WELDX_UNIT_REGISTRY.default_format = "~"
 
 WELDX_QUANTITY = WELDX_UNIT_REGISTRY.Quantity
 Q_ = WELDX_QUANTITY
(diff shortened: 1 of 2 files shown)


weldx/tests/test_measurement.py
```diff
@@ -454,7 +455,7 @@ def test_get_signal_data(self):
         mc.add_transformation(self._default_transformation(), data=data)
         mc.create_transformation("transformation_2", None, output_signal_unit="A")
 
-        assert mc.get_signal_data("transformation").identical(data)
```
CagtayFabry, line 457: I am not really sure why this worked in the first place since `get_signal_data` is supposed to return a `TimeSeries` (which doesn't have `.identical`) 🤔

vhirtham: Is that what the type hint says or the actual code. Maybe the type hints or the code need an adjustment too.

vhirtham (approved):

weldx/tests/test_measurement.py
CagtayFabry: the type hint says `TimeSeries` but I didn't check any further (the code looks more like `xarray`)

CagtayFabry:
updated to use short notation for serialization, see above

weldx/tests/test_measurement.py
vhirtham: Hmmm... strange. Will have to check this on my own. Usually, this should be caught by `mypy`

PR #522, opened 2021-09-14
support tagged string, add unit schema

CagtayFabry:
A very neat asdf feature that wasn't really feasible using the old API imo:
Not only `dict`s can be tagged by asdf but also `list` and `str`. This makes some concise and very simple schemas possible. (just return the type that you want in `to_yaml_tree`)

Take a look at the reworked `UUID` schema: it is no longer necessary to have an `object` with extra properties but the string itself gets tagged (which is basically what we want, a simple string that just represents a UUID)

here is an example asdf file snippet:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
%TAG !weldx! asdf://weldx.bam.de/weldx/tags/
--- !core/asdf-1.1.0
# ...
uuid: !weldx!uuid-1.0.0 2f946734-9074-41a2-b7b6-40861f1ffb97
unit: !weldx!unit/unit-1.0.0 meter
```

I have also added a very simple schema for representing units.

~~Using this, I think we should implement our own custom `Quantity` schema as well (currently using the default `asdf` schema)~~
Done

We should also serialize units directly wherever we use them (instead of working with strings which can be error prone)

## Changes
- add `constants.U_` for easy unit access
- changes to `WeldxConverter` to support not only `dict` but also `str`, `list` for tagging
- simplify UUID schema/code as tagged string
- add `asdf://weldx.bam.de/weldx/schemas/unit/unit-1.0.0` schema to represent `pint.Unit`
- add `asdf://weldx.bam.de/weldx/schemas/unit/quantity-1.0.0` schema to represent `pint.Quantity`

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [ ] updated doc/
- [x] update manifest file


weldx/schemas/weldx.bam.de/weldx/core/graph/di_node-1.0.0.yaml
```diff
@@ -32,7 +32,7 @@ examples:
 type: object
 properties:
   name:
-    oneOf:
+    anyOf:
       - type: string
       - tag: "asdf://weldx.bam.de/weldx/tags/uuid-1.0.0"
```
CagtayFabry, line 37:
```suggestion
    type: string
```
Only using `type: string` instead of the combiner structure would also work, but I find this more explicit in what types are supported

CagtayFabry:
As with the CSM PR, we can remove the 'compatibility' `oneOf` constructs in `0.6` if we want to

vhirtham:
Going to review this tomorrow morning

vhirtham (approved):
LGTM

weldx/schemas/weldx.bam.de/weldx/core/graph/di_node-1.0.0.yaml
vhirtham: Explicit is better than implicit - Would leave it as is :wink:

PR #525, opened 2021-09-15
Remove test case interdependency

### LINKED ISSUE #334: Test architecture internal dependencies
Since we started to run our tests in parallel, we have to ensure, that the order of execution is independent. For instance a test class
has a static attribute, which is being modified by the individual test methods.
``` python
class TestFoo:
    complex_structure = {}

    def test_add_attr1(self):
        self.complex_structure['a'] = 1
        assert len(self.complex_structure) == 1

    def test_add_attr2(self):
        self.complex_structure['b'] = 1
        assert len(self.complex_structure) == 2
```
Here we see, that we test_add_attr2 would fail, if is executed prior test_add_attr1. If the execution order of the test methods is strict, we can rely on a pytest-xdist feature to distribute all test cases to the nodes on per file basis, which basically circumvents this problem. Generally we should write tests such, that they have no interdependence.
One example where this interdependence occurs is weldx.tests.test_transformation.TestCoordinateSystemManager.test_add_coordinate_system.
This test method is parameterized to add different coordinate systems to the manager and it is being checked if the previous additions are still valid. The methods act on a single (static) instance of the CSM, which grows for every (parameterized-)test executed.

In my opinion we should avoid having static data as much as possible, since it is easy to cause side-effects, which therefore affects the outcome of the tests. Only the side-effect of the actual property of interest should be triggered and tested. What would happen if another programmer adds a test acting on the same static instance and affects the other tests in a non intended way. These kind of problems are hard to debug and should be avoided.
When the same data is needed, we should consider factory functions like pytest.fixtures or unittest.TestCase.setUp to create a clean and defined environment for the actual test case execution.

vhirtham:
## Changes

Removes the interdependency of different test cases

## Related Issues

Closes #334 

## Checks

- [x] ~~updated CHANGELOG.md~~
- [x] updated tests
- [x] ~~updated doc/~~


weldx/tests/transformations/test_cs_manager.py
```diff
@@ -88,56 +88,50 @@ def test_init():
...
 
 
-# test_add_coordinate_system -------------------------------------------------------
+# test_add_cs --------------------------------------------------------------------------
 
-# todo
-#  add time dependent systems. The problem is, that currently something messes
-#  up the comparison. The commented version of lcs_2 somehow switches the order of
-#  how 2 coordinates are stored in the Dataset. This lets the coordinate comparison
-#  fail.
```
vhirtham, line 97: This seems to be resolved now due to the refactoring of the CSM

vhirtham:
For the `add_cs` test I simply moved everything inside the function and removed the parametrization, which I think is okay here. I searched for other interdependencies in the "big" test files, but couldn't find any more. So I guess there wasn't too much to fix at all.

CagtayFabry:
Does that mean we can remove the `--dist=load file` flag now?
would be cool to see if it gets us any speedups

CagtayFabry (approved):
merge after checking the pytest config

vhirtham:
> Does that mean we can remove the `--dist=loadfile` flag now?
> would be cool to see if it gets us any speedups

Theoretically. I will test it and fix the remaining problems if possible

weldx/tests/transformations/test_cs_manager.py
```diff
@@ -88,56 +88,50 @@ def test_init():
...
+        parent = d[1]
+        lcs = d[2]
+        child_in_parent = d[3]
+
+        if name not in csm.coordinate_system_names:
+            exp_num_cs += 1
+
+        csm.add_cs(name, parent, lcs, child_in_parent)
+
+        assert csm.number_of_coordinate_systems == exp_num_cs
```
marscher, line 126: Thanks for fixing this. I suggest that you put a common assert message indicating the former input parameters. This is of course only interesting in case of a failing test, but eventually one does not need to fire up a debugger to know which case is giving trouble.

vhirtham: Good point. I will add a message.

vhirtham:
> Does that mean we can remove the `--dist=loadfile` flag now?
> would be cool to see if it gets us any speedups

This seems to work - the tests pass. I have removed `--dist=loadfile` from the WeldxFile job too. Let's see how this works out. However, I think it doesn't make sense to remove it for the notebooks I guess (does it even have an effect there?)

weldx/tests/transformations/test_cs_manager.py
vhirtham: Added a short message to all assertions. Not beautiful, but should do the trick.

PR #530, opened 2021-09-16
Add pre commit

CagtayFabry:
In a final attempt to end all linting whoes I have added a (what I think) sensible `pre-commit` config for our project.
The config is used for checking only in a CI run. Of course, it can also be used for a local install (have set it up and tested)

resources:
- https://pre-commit.com/
- https://github.com/pre-commit/pre-commit-hooks (basic hooks and configuration)

 The GitHub action seems to do the job nicely for now, we can also look into https://pre-commit.ci/ 
It looks like https://pre-commit.ci/ allows to edit the code in PR using the pre-commit config, which could make linting a lot easier in general (certainly for outside contributors)
If that saves us time we could give it a try but I will leave that up to you @vhirtham @marscher 

## Changes
- add `.pre-commit-config.yaml`
- add ci job for pre-commit lint check


weldx/schemas/weldx.bam.de/weldx/core/geometry/spatial_data-1.0.0.yaml
```diff
@@ -38,7 +38,7 @@ properties:
       datatype:
         type: string
         enum: [float32, float64]
-    wx_shape: [(:), 3]
+    wx_shape: [(~), 3]
```
CagtayFabry, line 41: I have changes this because `:` is not a valid yaml syntax in this context (it got picked up by the linter)

I have already removed the mention of `:` in [the documentation](https://weldx.readthedocs.io/en/latest/standard/shape-validation.html) but I think we might also just remove the support in the code as well? @vhirtham @mark141

marscher (approved):
This looks super useful! So this also can be run locally by just installing the precommit python package?

CagtayFabry:
yes, basically there are two ways to use `pre-commit` that make sense for us (you need to have `pre-commit` installed in your environment for any of these: `mamba install pre-commit -c conda-forge`)

- trigger a 'manual' run of the pre-commit configuration
  `pre-commit run` (only run on staged/changed files - default)
  `pre-commit run --all-files` (run on all files, I used this to format the whole repo in f4da8f00080994d59c8fdb3c4f3118018d5b6b36)
- setup `pre-commit` as a git-hook, this will run the pre-commit configuration on all staged/changed files in your current workdir (that would be included in the commit) whenever you git commit.
(I think this will apply the changes in your commit 'in place', so in principle you could also deactivate auto formatting with black on save for example since this will run on any commit files anyway)

CagtayFabry:
If needed I will redo this after #533

vhirtham:
I will check this tomorrow

Issue #529, opened 2021-09-16
remove legacy and compatibility code and schemas


CagtayFabry:
look for the following throughout the project

- ~~`legacy_tag` (in schemas)~~  #533
- `legacy_code` (in code)
- `tag:weldx.bam.de:` everywhere
- `tag:stsci.edu:asdf/unit/quantity-1.1.0` or `!unit/quantity`
  (we can keep read support)
- remove `get_legacy_resource_mapping()`
- remove adding legacy tags in `WeldxConverterMeta`
- ~~update manifest script to not add legacy tags~~ #533

any other things that should get removed?

PR #533, opened 2021-09-16
Add legacy files

CagtayFabry:
I think I have (finally) figured out a 'nice' way to keep the support for our old `weldx` tags and files without cluttering the new schemas. The new `asdf` API makes this pleasantly easy.

In the end this should allow the following:
- you can open and validate all `0.4` version files, saving a copy will automatically convert to the new version and tags
- obviously it should not affect purely "new" files that are generated form master
- you **cannot** open files that have a mixture of `0.4` and `0.5` version tags in them, but I don't think any of these exists in the wild anyway, you would have to try to update an old file with the new code to create those

## Changes
- include the full `0.4.1` schema files in `weldx/schema/weldx.bam.de/legacy`
- setup resource mapping for legacy schemas
- clean up all new schema files (remove `oneOf` etc. for legacy files)
- update the manifest file and script

## Related Issues

#529
#467 

## Checks

- [x] updated CHANGELOG.md
- [x] update manifest file


vhirtham:
Maybe you can run me through the important bits if you got the time. I guess there are only a few relevant lines, but finding all of them in 4000+ additions and 98 files would be a bit time consuming.

CagtayFabry:
> 
> 
> Maybe you can run me through the important bits if you got the time. I guess there are only a few relevant lines, but finding all of them in 4000+ additions and 98 files would be a bit time consuming.

It's not that much actually, basically we have two complete sets of schemas in the repository:
- the 'new' schemas following the `asdf://` syntax for tag and schema URI
  those are stored under `weldx/schema/weldx.bam.de/weldx` (our default schema directory)
- the 'old' schemas with the `tag:weldx.bam.de/` and `httsp://weldx.bam.de` tag/schema uris
  those are kept under `weldx/schema/weldx.bam.de/legacy`

the `WeldxExtension` knows both schema sets via their `ResourceMappings` (in `extension.py`)
By default the `*Converter` classes only know the new tag format but the 'WeldxConverterMeta' links the old tags formats to the same converters

When reading an *old* file only the *old* tags/uris are encountered and validated against the correct (*old*) files.
When saving a file everything defaults to the new styles.

The good thing is we don't have to care about the old schemas because they stay as is and we can style the new schemas however we want. Just in cases where we break the old implementation for serialization we have to provide some backwards compatibility for reading. (see `UuidConverter` or the `CSMConverter`)

PR #535, opened 2021-09-16
change weldx schema versions to `0.1.0`

### LINKED ISSUE #531: default tag version
with `0.5.0` and the switch to the new ASDF extension API we also get a 'fresh start' with the version numbering

I would prefer to reserve the `1.0.0` version for schema files for the initial (final) `1.0.0` release of weldx
But since there are now use cases based on the format we should also keep compatibility in mind from `0.6.0` forward.

So I am thinking, maybe we should reset the schema versions to `0.1.0` and apply all the versioning mechanisms of asdf going forward. We don't have to keep backwards compatibility code in the codebase for too long imo but should avoid breaking files between releases (like we are doing now with the old tag format)

~~I might remember trying to set a `0.1.0` version somewhere before and running into some kind of issue but not sure so I will try it again just to see if that is viable~~
everything seems to work fine with `0.1.0` #535

please offer your opinions @vhirtham @marscher @AndreasPittner

CagtayFabry:
## Changes
change the weldx schema version to `0.1.0`

## Related Issues

Closes #531

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/
- [x] update manifest file


vhirtham (approved):

weldx/tests/transformations/test_cs_manager.py
```diff
@@ -1824,6 +1824,7 @@ def test_remove_subsystems(list_of_csm_and_lcs_instances, nested):
     ],
 )
 @pytest.mark.slow
+@pytest.mark.filterwarnings("ignore:The following coordinate systems.*:UserWarning")
```
CagtayFabry, line 1821: fyi @vhirtham I have disable the specific CSM warning for this test to clean up the output

vhirtham: Noticed recently, that there are new warnings, just didn't realize so far where they were introduced since we at all of them gone at some point.

PR #536, opened 2021-09-16
Export geometry as CAD file

vhirtham:
## Changes

Adds a method to the `Geometry` class that allows to save it as any point- and triangle-based CAD file that is supported by `meshio`.

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


vhirtham:
Here is a little test script: 

~~~python
from weldx import Q_, Geometry, LinearHorizontalTraceSegment, Trace, get_groove

groove = get_groove(
    groove_type="VGroove",
    workpiece_thickness=Q_("5mm"),
    groove_angle=Q_("50deg"),
    root_face=Q_("1mm"),
    root_gap=Q_("1mm"),
)

trace = Trace(LinearHorizontalTraceSegment(Q_("300mm")))

geometry = Geometry(groove.to_profile(width_default=Q_("50mm")), trace)

geometry.write_to_file(
    "test.stl", profile_raster_width=Q_("10mm"), trace_raster_width=Q_("50mm")
)
~~~

vhirtham:
I loaded it into FreeCAD and everything looks fine so far. I will test it with different groove types. 

I separated the rasterization/triangulation for each of the `Profile`s `Shape`s. The idea is to separate all the data per part if `meshio` and the data format support it. I guess this is more convenient for our workshops.

As far as I can tell, most things should work except for variable profiles. Radial traces will probably cause some trouble due to profile overlaps. As mentioned in an earlier PR, extremely convex profiles (we don't have them yet) have an increased probability to break the profile triangulation algorithm.

I will talk to one of our workers if the files are usable for them.

weldx/geometry.py
```diff
@@ -2377,6 +2410,102 @@ def spatial_data(
...
+        points = []
+        triangles = []
+        for i, shape_raster_data in enumerate(raster_data):
+            shape_points, shape_triangle = self._shape_points_and_triangles(
+                shape_raster_data,
+                len(points),
+                self.profile.shapes[i].is_polygon_winding_order_cw(),
+            )
+            points += shape_points
+            triangles += shape_triangle
```
CagtayFabry, line 2505: can we simply move this part to the end of `Geometry.spatial_data()` ?

The way I understand it now we get a different triangulation when calling `write_to_file()` on a `Geometry` directly or on the `SpatialData` (via `Geometry.rasterize().write_to_file()` )

I think getting a single (correct) triangulation when calling `Geometry.rasterize()` is more intuitive than different triangulations. In that case we would not need the `Geometry.write_to_file()` function but could use the existing `SpatialData.write_to_file()` function

I think the intu way would be

weldx/geometry.py
```diff
@@ -2377,6 +2410,102 @@ def spatial_data(
         )
         return SpatialData.from_geometry_raster(rasterization, closed_mesh)
 
+    def _shape_points_and_triangles(
+        self, shape_raster_data: np.array, offset: int, cw_ordering: bool
+    ):
+        if isinstance(self._profile, VariableProfile):
+            raise NotImplementedError
+
+        shape_raster_data = shape_raster_data.swapaxes(1, 2)
```
CagtayFabry, line 2419: I noticed in my 3D Viewer that the model orientation seems 'sideways', is this because of this line?

CagtayFabry:
when this is ready we can add it to the design dashboard @marscher

weldx/geometry.py
vhirtham: Have mostly implemented this now. Just have to replace the branch for single parts. Afterwards, I can remove the old triangulation function (maybe, see below). I just realized that the implementation should also work for variable profiles if the number of points for different profiles stays the same. However, I have to check the rasterization output for variable profiles. I am actually not sure what it returns because the different profiles can have a different number of points and as far as I know, this shouldn't play well with `numpy.array` which is the return type for constant profiles. I came across this when I removed the `VariableProfile` exception, which can't be used the same way in the `SpatialData` as in the `Geometry`.

Apart from that, I will come up with a triangulation version for `VariableProfiles` (already have an idea) or we keep the old one for them.

marscher:
> when this is ready we can add it to the design dashboard @marscher

Sure, will do.

vhirtham:
I cleaned everything up and tested some other file formats. As mentioned multiple times, the algorithm for the front and back faces delivers bad triangulations for concave surfaces. This can be seen when using a U-Groove. However, I would leave it for now until we need it. The corresponding triangulation is placed inside its own function so that we can easily swap it out. 

After looking into the current implementations of variable profiles I come to the conclusion that we might want to postpone it too since it would require more coding and maybe some rewrites. Basically, we just need to do a per shape segment triangulation but as far as I can see the current code needs to be changed for this and I don't know if that turns into a never-ending story, which isn't worth it for functionality we currently don't need and a module we might rewrite entirely.

vhirtham:
I also haven't reached the corresponding employee of the workshop that can tell me if the current file is useable with their software. I will retry it tomorrow.

weldx/geometry.py
```diff
@@ -2457,40 +2569,14 @@ def from_geometry_raster(
             New `SpatialData` instance
 
         """
-        # todo: this needs a test
-        # todo: workaround ... fix the real problem
-        # if not isinstance(geometry_raster, np.ndarray):
-        #    geometry_raster = np.array(geometry_raster)
-        if geometry_raster[0].ndim == 2:
-            return SpatialData(*_triangulate_geometry(geometry_raster))
```
vhirtham, line 2465: I removed this branch entirely. I am not sure if this is still an issue, however, notebooks and tests run without it. To my understanding, this branch should just catch cases where a single shape was passed to the profile. I tested a profile with a single shape and it passed. So I guess it is fine. If not, will find out.

CagtayFabry:
> I cleaned everything up and tested some other file formats. As mentioned multiple times, the algorithm for the front and back faces delivers bad triangulations for concave surfaces. This can be seen when using a U-Groove. However, I would leave it for now until we need it. The corresponding triangulation is placed inside its own function so that we can easily swap it out.
> 
> After looking into the current implementations of variable profiles I come to the conclusion that we might want to postpone it too since it would require more coding and maybe some rewrites. Basically, we just need to do a per shape segment triangulation but as far as I can see the current code needs to be changed for this and I don't know if that turns into a never-ending story, which isn't worth it for functionality we currently don't need and a module we might rewrite entirely.
> 
> EDIT:
> I initially thaught that we need the variable profile triangulation for the plots, but I realized that we skipped it there too 😉

agree, we should just skip the variable profiles for now
In any case users can implement their own triangulation but it is nice to have for the basic stuff that we can use. 👍

weldx/geometry.py
```diff
@@ -2439,8 +2452,107 @@ def from_file(file_name: Union[str, Path]) -> SpatialData:
...
+
+        return [
+            *cls._shape_profile_triangles(n_prf, n_prf_pts, offset, cw_ord),
+            *cls._shape_front_back_triangles(n_prf, n_prf_pts, offset, cw_ord),
+        ]
+
+    @classmethod
     def from_geometry_raster(
-        geometry_raster: np.ndarray, closed_mesh: bool = True
+        cls, geometry_raster: np.ndarray, closed_mesh: bool = True
```
CagtayFabry, line 2555: the `closed_mesh` argument is not used in the code

CagtayFabry (approved):
looking at the code I assume I now get the same output from `Geometry.write_to_file()` and `Geometry.spatial_data().write_to_file()` ?

What do you think about simply naming the methods `to_file()`?

weldx/geometry.py
vhirtham: Right. The question is: Do we want to keep this option (easy to add to the new triangulation) or do we want to remove it?

vhirtham:
> looking at the code I assume I now get the same output from `Geometry.write_to_file()` and `Geometry.spatial_data().write_to_file()` ?
> 
> What do you think about simply naming the methods `to_file()`?

sounds good

weldx/geometry.py
CagtayFabry: I thought the triangles would get set, it looks like that in the code? (line 2576)
Either way, would make sense to have the option

vhirtham: Not sure if we are on the same boat here: The purpose of ´closed_mesh´ was to generate the additional side surfaces that were not covered by the initial algorithm. Now they are always generated but since it happens in its own method, we can just add a branch there. We just need to pass the option down to the method. I think it would be nice for the matplotlib plots since the additional triangles sometimes clutter the picture.

CagtayFabry: ok I see, agree that it can be hand in some cases to not have those triangulations

weldx/geometry.py
```diff
@@ -2534,6 +2534,8 @@ def _shape_triangles(
             Raster data of the shape
         offset :
             An offset that will be added to all indices.
+        closed_mesh :
+            If `True`, the side faces of the geometry will also be triangulated.
```
CagtayFabry, line 2538: From what you showed me before, do you think "front and back faces" is fitting here?

vhirtham: It is probably more specific. I will adjust it and "smuggle it in" with another PR

PR #540, opened 2021-09-17
Doc fixes2

### LINKED ISSUE #526: documentation fixes
some minor points regarding the documentation

- the `Q_` docstring seems to have some errors in the examples formatting: https://weldx.readthedocs.io/en/latest/_autosummary/weldx.Q_.html
- `Time` is missing at the user API section (should be next to `Q_` I guess)
- can we get the objects listed in the `constants` module? https://weldx.readthedocs.io/en/latest/_autosummary/weldx.constants.html (not sure how this works, do we add docstrings?)
- move `GEOMETRY` section to the end of the user API documentation (imo)

marscher:
## Changes

minor doc fixes

## Related Issues

Closes #526

## Checks

- [x] updated doc/


doc/conf.py
```diff
@@ -273,7 +273,7 @@ def _prevent_sphinx_circular_imports_bug():
     "matplotlib": ("https://matplotlib.org", None),
     # "dask": ("https://docs.dask.org/en/latest", None),
     # "numba": ("https://numba.pydata.org/numba-doc/latest", None),
-    "pint": ("https://pint.readthedocs.io/en/stable", None),
+    "pint": ("https://pint.readthedocs.io/en/latest", None),
```
CagtayFabry, line 276: I had hoped this would include your fix in the `pint` docs but somehow it doesn't 😞  @marscher

CagtayFabry:
I'm fine with leaving `U_` out of the documentation for now since it doesn't get any real use so far in user facing code

doc/conf.py
marscher: They merged it quite some time ago, but it hasn't been released yet, right? I'd guess latest is just link to the latest release, not unreleased main branch stuff :)

marscher:
I guess that is fine. We never use `Unit` in an example or tutorial.

doc/conf.py
CagtayFabry: I think `stable` is the current release and `latest` is actually the main branch (same as for weldx or asdf etc.)

The fix is in there but still doesn't work somehow..
https://pint.readthedocs.io/en/latest/developers_reference.html?highlight=to_compact#pint.Quantity.to_compact

CagtayFabry:
> I really do not understand, why Sphinx cannot find the TimeDependent class in the Time docstring.

Don't know, in that case just double escape it to disable linking and we can look into it another time

CagtayFabry:
merci

PR #541, opened 2021-09-20
use tagged string schemas

CagtayFabry:
## Changes
update the following schemas/tags to be `type: string` instead of `type: object`
- `timedelta`
- `timestamp`

also removed some `type: object` schema requirements where only `tag` is sufficient

## Related Issues

Closes # (add issue numbers)

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] updated doc/


vhirtham:
> The only downside of using tagged strings instead of dicts is that we cannot add our `wx_user` and `wx_metadata` fields to these but I think that is ok for these singleton instances?

Shouldn't be an issue. Can't think of an example where this would make sense.

CagtayFabry:
the missing coverage on the legacy code is expected, I will test again with the reference dataset

vhirtham (approved):

PR #544, opened 2021-09-20
update pydocstyle match patterns

### LINKED ISSUE #543: Pydocstyle setup broken
While working on the `geometry` package I realized that pydocstyle does not report issues anymore. It passes no matter which documentation crime I commit. This seems to be related to the recent pre-commit PRs since removing the '"' here fixes the problem partially:

**`setup.cfg`** 
~~~ python
[pydocstyle]
# convention numpy is currently equivalent to ignoring 'D107', 'D203', 'D212', 'D213', 'D402', 'D413'
convention = numpy
# match options need to be in "" for pre-commit
match = "(?!__)(?!_version)(?!conftest).*\.py"
match_dir = "[^\.][^\docs][^\tags].*"
~~~

But the problem remains for example for the files of the `transformations` package. My first thought was that subfolders might be ignored generally, but it works normally for `util`. So it must get skipped for some other reason I haven't figured out yet.

CagtayFabry:
## Changes
update the pydocstyle config patterns

apparently the old patterns didn't really work as intended.
Note that pydocstyle must be run in `./weldx` for this to work correctly

~~I guess we need a different config for the pre-commit hook (which doesn't work recursively)~~ done

## Related Issues

Closes #543


CagtayFabry:
would be good if we could reproduce the same output for the `pre-commit` run to get a better idea for the syntax

CagtayFabry:
I think this is what we want @vhirtham  ?

CagtayFabry:
I actually like the failed output formatting from the pre-commit.ci tool
https://results.pre-commit.ci/repo/github/238160452

CagtayFabry:
both the [pydocstyle action](https://github.com/BAMWelDX/weldx/runs/3654399651) and [pre-commit action](https://github.com/BAMWelDX/weldx/runs/3654399862) file with the same errors so hopefully the configuration is identical now

vhirtham:
> I think this is what we want @vhirtham ?

if pydocstyle checks all (relevant) files again, then this is at least what I want 😉

CagtayFabry:
feel free to change as desired

vhirtham:
> feel free to change as desired

Don't know what to change as long as it works as intended

PR #545, opened 2021-09-20
add black and isort hooks for notebooks

CagtayFabry:
## Changes
add hooks to format notebooks


CagtayFabry:
easy notebook formatting with `pre-commit run --all-files` 🎉 @vhirtham @marscher

marscher:
does it clear output too?

CagtayFabry:
> does it clear output too?

pretty sure it doesn't, basically just a helper to apply any of the regular linters to notebooks.
We probably could create our own notebook linter as a pre-commit hook if we wanted too but it might be a little overkill

PR #551, opened 2021-09-21
units refactor

### LINKED ISSUE #550: nomenclature: `unit` vs `units`
### Discussed in https://github.com/BAMWelDX/weldx/discussions/528

<div type='discussions-op-text'>

<sup>Originally posted by **CagtayFabry** September 15, 2021</sup>
Since we implemented our own `quantity` schema in #519 we no longer have to follow the asdf schema which would allow us to rename the `unit` field of the quantity

In general two options come to mind: `unit` oder `units`
I think it would be very good to decide on one of these options and use that consistently throughout `weldx`

Maybe we should look into other fields what is most common. I know that `pint-xarray` stores unit information under the `units` key but not sure if that is actually defined in the CF-conventions

Any preferences? please discuss</div>

CagtayFabry:
## Changes
refactor `unit` -> `units` in schemas

also serialize as `unit/unit` instead of string

## Related Issues

Closes #550

## Checks

- [x] updated CHANGELOG.md
- [x] updated tests
- [x] update manifest file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,8 @@
   [[#494]](https://github.com/BAMWelDX/weldx/pull/494).
 - `compare_nested` can compare sets [[#496]](https://github.com/BAMWelDX/weldx/pull/496)
 - `WeldxFile` cleans old memory blocks during file updates [[#539]](https://github.com/BAMWelDX/weldx/pull/539).
-- `WeldxFile` respects `mode` argument also for BytesIO and file handles [[#539]](https://github.com/BAMWelDX/weldx/pull/539).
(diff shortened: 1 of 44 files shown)


CagtayFabry:
I have added where I think reading the old files needs to be considered but not yet tested

CagtayFabry:
I think this looks pretty good, did a quick run on the old dataset as well

CagtayFabry:
> I think this looks pretty good, did a quick run on the old dataset as well

looking good 🚀

CagtayFabry:
sorry I actually wanted to merge the other PR @vhirtham ... 🙄

PR #555, opened 2021-09-23
New tutorials

vhirtham:
## Changes

Adds a set  of new tutorials as discussed [here](https://github.com/BAMWelDX/weldx/discussions/455)

## Related Issues

Closes #400 

## Checks

- [ ] remove `single_pass_weld.wx`
- [ ] updated CHANGELOG.md
- [ ] updated tests
- [ ] updated doc/
- [ ] update example/tutorial notebooks
- [ ] update manifest file

PR #560, opened 2021-09-27
long format pint serialization

CagtayFabry:
## Changes
- use long pint format for serialization
- small fixes
- update the GH changelog action

### DIFF
--- a/weldx/geometry.py
+++ b/weldx/geometry.py
@@ -2535,7 +2535,8 @@ class SpatialData:
         offset :
             An offset that will be added to all indices.
         closed_mesh :
-            If `True`, the side faces of the geometry will also be triangulated.
+            If `True`, the front and back faces of the geometry will also be
+            triangulated.
 
         Returns
         -------
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -16,5 +16,5 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Grep for PR number in CHANGELOG.md
-        run: grep -P '\[[^\]]*#${{github.event.number}}[,\]]' CHANGELOG.md
+      - name: Grep for PR number in CHANGELOG.rst
+        run: grep -P '\[[^\]]*#${{github.event.number}}[,\]]' CHANGELOG.rst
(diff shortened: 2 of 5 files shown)


marscher:
Does this fix this "delta" unit issue which started to pop when @vhirtham started the new file tutorial in #555?

Are the long description in any sense "more" reliable?
I was also thinking that the registry used to restore the units could change in the meantime. Is this case covered? (sorry for being off-topic, it just came to my mind again).

CagtayFabry:
> Does this fix this "delta" unit issue which started to pop when @vhirtham started the new file tutorial in #555?
> 
> Are the long description in any sense "more" reliable?
> I was also thinking that the registry used to restore the units could change in the meantime. Is this case covered? (sorry for being off-topic, it just came to my mind again).


it should fix the issue, yes.
In that way the long description could be considered more reliable so we should keep it for now.

currently we simply apply the default [pint unit registry](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt) which I consider rather stable since it follows other well known conventions.
However we might also decide to distribute and load our own copy of the unit dictionary with each weldx version.

weldx/geometry.py
```diff
@@ -2535,7 +2535,8 @@ def _shape_triangles(
         offset :
             An offset that will be added to all indices.
         closed_mesh :
-            If `True`, the side faces of the geometry will also be triangulated.
+            If `True`, the front and back faces of the geometry will also be
+            triangulated.
```
vhirtham, line 2539: Also fixed that one on the doc branch 😄

vhirtham (approved):

marscher:
> currently we simply apply the default [pint unit registry](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt) which I consider rather stable since it follows other well known conventions.
> However we might also decide to distribute and load our own copy of the unit dictionary with each weldx version.

The "problem" is that we have "user-defined" units like percentage and so on. If the registry would change, e.g. we decide to drop a unit again, old files cannot be opened again. But I guess you are right, that shipping the unit-dict would solve that. At the moment it seems a bit overkill of course.

CagtayFabry:
> > currently we simply apply the default [pint unit registry](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt) which I consider rather stable since it follows other well known conventions.
> > However we might also decide to distribute and load our own copy of the unit dictionary with each weldx version.
> 
> The "problem" is that we have "user-defined" units like percentage and so on. If the registry would change, e.g. we decide to drop a unit again, old files cannot be opened again. But I guess you are right, that shipping the unit-dict would solve that. At the moment it seems a bit overkill of course.

exactly, currently we load the default pint registry (by using pint defaults) and change the few lines that we want to from our code in `constants.py`
of course we could also make these changes in the file and ship that but I think the current approach works fine for now

PR #561, opened 2021-09-27
update code to use units

### LINKED ISSUE #527: use `pint.Unit` explicitly instead of strings
after #519 using units (outside of `pint.Quantity`) should now be be easy, I have added a simple `U_` class to go with `Q_` for short access.

To make the code more usable and fully use the `pint` capabilities we should use the `pint.Unit` class wherever possible instead of working with 'unit strings' . Basically units should not be represented as strings (unless in the `xarray.DataArray.attrs.units` field, since that might not work with `pint-xarray` but we should check (would be awesome!).

For compatibility reasons we should ~still keep and allow the `type: string` in the schema~ and explicitly convert to units using `U_` (only where needed, not when it goes into `Q_` directly) after reading the tree.
*since we seperated old and new schemas it is no longer necessary to update the schema restrictions*. We should still convert to `U_` on read though to keep compatibility

CagtayFabry:
## Changes
- use pint units explicity

## Related Issues

Closes #527

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/geometry.py
+++ b/weldx/geometry.py
@@ -1169,7 +1169,7 @@ class Shape:
 class Profile:
     """Defines a 2d profile."""
 
-    def __init__(self, shapes, units=None):
+    def __init__(self, shapes, units: pint.Unit = None):
         """Construct profile class.
 
         Parameters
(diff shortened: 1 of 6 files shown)


CagtayFabry:
not sure if I caught every line that deals with units but I think the important stuff is covered @vhirtham

vhirtham (approved):
LGTM

PR #562, opened 2021-09-27
[WeldxFile] update signature matches parent

marscher:
## Changes

The iterable or mapping argument can be truly optional.
So update with only passing kwargs can work.

## Checks

- [x] updated tests


marscher:
I think there is no need for a changelog entry, because the update interface has just been overriden on master and was not yet released.

PR #565, opened 2021-09-28
add  `Δ°` pint preprocessor

CagtayFabry:
## Changes
- add text preprocessor for handling `Δ°` in unit notation

I have tested it for  `Δ°C`, `Δ°F` and `Δ°Re` but I think we should stick to the long serialization format for now @marscher @vhirtham

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Release Notes
 added
 ~~~~~
 
--  added “units” (exact) and “dimensionality” (dimensionality
+-  added "units" (exact) and “dimensionality” (dimensionality
    compatible) checking options to `util.xr_check_coords`
    `[#442] <https://github.com/BAMWelDX/weldx/pull/442>`__
 -  `Time` class that can be initialized from several other time types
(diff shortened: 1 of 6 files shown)


marscher:
I'm just curios why this preprocessor is mandatory, if we do not even request to export a "delta" unit in the first place.

CagtayFabry:
> I'm just curios why this preprocessor is mandatory, if we do not even request to export a "delta" unit in the first place.

just for convenience so we can use e.g. `Q_(150, "Δ°C")` since the default format everywhere outside of serialization is still 'short' so `Δ°` might still pop up in console outputs, plots etc.

marscher:
Shouldn't we add a test for preprocessors? I think at least the delta case is not covered yet.

CagtayFabry:
> Shouldn't we add a test for preprocessors? I think at least the delta case is not covered yet.

good idea, I will integrate it in one of the other testcases

CagtayFabry:
please note the changes I made to unit checking in `weldx/tests/test_core.py` @vhirtham 
using `.is_compatible_with` for unit checking looks the most straight forward to me

vhirtham (approved):

PR #571, opened 2021-09-29
update wx_shape validator optional syntax

### LINKED ISSUE #184: redefine "optional" syntax for wx_shape validation
thinking about #176:

the current implementation (from #176) allows shape definition for optional properties (good), however I see the following problem:

- the optional indication like `(optional_prop)` is redundant with the `required` field (and must be kept in sync). Every field that is not a required property has to have the shape attribute as optional for the schema to make sense.

### proposal:

1. only use ASDF default `required: ...` to indicate required fields
2. allow shape validation to pass **by default** if a property is missing (since this schema violation should be caught by `required`)
3. rework the `required_shape / (optional_shape)` syntax to do the following:
    - if the property exists and a shape is *required*, the property **must** report a shape **and** validate against the shape requirements
    - if the property exists and a shape is *optional*, the property **must** validate against the shape requirements **only if** it reports a shape
    - if the property exists and a shape is *optional*, the property passes validation if it does not report a shape

### Example 1:
```yaml
properties:
  required_prop:
  optional_prop:
wx_shape:
  required_prop: [n]
  optional_prop: [n]
required: [required_prop]
```
This is the default case for optional properties. If `optional_prop` exists, it must validate correctly. If it does not exist it is skipped during `wx_shape` validation.

### Example 2:
```yaml
properties:
  required_prop:
  optional_prop:
    oneOf:
      - tag: tag:stsci.edu:asdf/core/ndarray-1.0.0
      - tag: **/timedeltaindex-1.0.0
      - type: string
wx_shape:
  required_prop: [n]
  (optional_prop): [n]
required: [required_prop]
```
This is an example that shows the new behavior.
- If `optional_prop` does not exist, it is simply skipped (like example 1)
- If `optional_prop` does exist and is a `ndarray`, it must validate the `[n]` shape (because we can evaluate the shape of `ndarray`)
- If `optional_prop` does exist and is a `timedeltaindex`, it must validate the `[n]` shape (because we can evaluate the shape of `timedeltaindex` from our custom code)
- If `optional_prop` does exist and is a `string`, shape validation is skipped (because we don't define a shape property of `string`)

### Example 3:
```yaml
properties:
  required_prop:
  optional_prop:
wx_shape:
  required_prop: [n]
  (optional_prop): [n]
required: [required_prop]
```
This case shows the same behavior as Example 1, and is redundant since all possibilities of `optional_prop` define a shape. It should be avoided.

CagtayFabry:
## Changes
see #184

## Related Issues
Closes #184

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
+++ b/weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
@@ -103,14 +103,14 @@ properties:
     tag: "asdf://weldx.bam.de/weldx/tags/time/timedeltaindex-0.1.*"
   coordinates:
     description: |
-      A ndarray containing the coordinates.
+      An ndarray containing the coordinates.
 
     oneOf:
       - tag: "asdf://weldx.bam.de/weldx/tags/core/variable-0.1.*"
       - tag: "asdf://weldx.bam.de/weldx/tags/core/time_series-0.1.*"
   orientations:
     description: |
-      A ndarray containing the orientations.
+      An ndarray containing the orientations.
     tag: "asdf://weldx.bam.de/weldx/tags/core/variable-0.1.*"
 wx_shape:
   coordinates: [..., 3]
--- a/weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
+++ b/weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
@@ -103,20 +103,18 @@ properties:
     tag: "asdf://weldx.bam.de/weldx/tags/time/timedeltaindex-0.1.*"
   coordinates:
     description: |
-      An ndarray containing the coordinates.
+      A ndarray containing the coordinates.
 
     oneOf:
       - tag: "asdf://weldx.bam.de/weldx/tags/core/variable-0.1.*"
       - tag: "asdf://weldx.bam.de/weldx/tags/core/time_series-0.1.*"
   orientations:
     description: |
-      An ndarray containing the orientations.
+      A ndarray containing the orientations.
     tag: "asdf://weldx.bam.de/weldx/tags/core/variable-0.1.*"
-
 wx_shape:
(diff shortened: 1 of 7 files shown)


CagtayFabry:
I have added another way to indicate a optional shape requirement by enclosing the list in braces for instances where `wx_shape` is defined directly at the property
```yaml
wx_shape: ([1,2,a])
```

weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
```diff
@@ -103,20 +103,18 @@ properties:
     tag: "asdf://weldx.bam.de/weldx/tags/time/timedeltaindex-0.1.*"
   coordinates:
     description: |
-      An ndarray containing the coordinates.
+      A ndarray containing the coordinates.
```
vhirtham, line 106: I think the previous version was correct due to the way `ndarray` is pronounced. But not 100% sure here

weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
```diff
@@ -103,20 +103,18 @@ properties:
...
 
     oneOf:
       - tag: "asdf://weldx.bam.de/weldx/tags/core/variable-0.1.*"
       - tag: "asdf://weldx.bam.de/weldx/tags/core/time_series-0.1.*"
   orientations:
     description: |
-      An ndarray containing the orientations.
+      A ndarray containing the orientations.
```
vhirtham, line 113: Same as above

vhirtham (approved):

weldx/schemas/weldx.bam.de/weldx/core/transformations/local_coordinate_system-0.1.0.yaml
CagtayFabry: true! I just auto-accepted the PyCharm spelling correction

PR #576, opened 2021-10-01
Add `WeldxFile.info`

### LINKED ISSUE #558: `WeldxFile` content printer for pure Python
For jupyter notebooks, we can visualize the content of a WeldxFile rather easily in a non-verbose way with `show_asdf_header`.
In pure python however, I can only print the unwieldy asdf YAML header or start to write some custom code to navigate manually through the tree (unless I missed something in the tutorial). It shouldn't be too hard to write a function (or expand `show_asdf_header`) that traverses the tree recursively until it finds an object that is not a container and prints its type.
So an output like this would be nice (comments are just for understanding the principle and won't be in the true output):

~~~
# Object items
- coordinate_systems: <CoordinateSystemManager>
- measurements: <MeasurementChain> 
# List and set items
- data:
    - <SpatialData>
    - <SpatialData>
    - <SpatialData>
# dictionary items
- wxmeta:
    - welder: <str>
    - timestamps:
        - start: <Time>
        - end: <Time>  
    - error:
        - failurecode: <int>
        - message: <str>
~~~

This way I know exactly how to navigate through a tree to get the item I want.

For lists we might think of a different solution though, like just printing the different types in it and its length

What do you think @marscher @CagtayFabry

vhirtham:
## Changes

Add `WeldxFile.info` to print a quick content overview to the stdout.

## Related Issues

Closes #558

## Checks

- [x] updated CHANGELOG.rst
- [x] ~~updated tests~~
- [x] ~~updated doc/~~

### DIFF
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -812,6 +812,27 @@ class WeldxFile(UserDict):
 
         display(self.show_asdf_header(use_widgets=True, _interactive=True))
 
+    def info(
+        self, max_rows: int = None, max_length: int = None, show_values: bool = False
+    ):
+        """Print the content to the stdout.
+
+        Parameters
+        ----------
+        max_rows :
+            The maximum number of rows that will be printed. If rows are cut, a
+            corresponding message will be printed
+        max_length :
+            The maximum line length. Longer lines will be truncated
+        show_values :
+            Set to `True` if primitive values should be displayed
+
+        """
+        tree = self._asdf_handle.tree.copy()
+        tree.pop("asdf_library")
+        tree.pop("history")
+        asdf.info(tree, max_rows=max_rows, max_cols=max_length, show_values=show_values)
+
 
 class _HeaderVisualizer:
     def __init__(self, asdf_handle: AsdfFile):
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -828,7 +828,7 @@ class WeldxFile(UserDict):
             Set to `True` if primitive values should be displayed
 
         """
-        tree = self._asdf_handle.tree.copy()
+        tree = {key:value  for key, value in self._asdf_handle.tree if key not in ["asdf_library", "history"]}
         tree.pop("asdf_library")
         tree.pop("history")
         asdf.info(tree, max_rows=max_rows, max_cols=max_length, show_values=show_values)
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -827,6 +827,31 @@ class WeldxFile(UserDict):
         show_values :
             Set to `True` if primitive values should be displayed
 
+        Examples
+        --------
+
+        >>> from weldx import WeldxFile
+        >>> tree = dict(
+        ...     project="Weldx",
+        ...     developers=["Cagtay Fabry", "Volker Hirthammer", "Martin Scherer"],
+        ...     meta=dict(depends_on=["ASDF", "NumPy", "some more"], is_on_github=True),
+        ... )
+        >>> wxfile = WeldxFile(tree=tree, mode="rw")
+        >>>
+        >>> wxfile.info()
+        \x1b[1mroot\x1b[0m (dict)
+        \x1b[2m\u251c\u2500\x1b[0m\x1b[1mproject\x1b[0m (str)
+        \x1b[2m\u251c\u2500\x1b[0m\x1b[1mdevelopers\x1b[0m (list)
+        \x1b[2m\u2502 \u251c\u2500\x1b[0m[\x1b[1m0\x1b[0m] (str)
+        \x1b[2m\u2502 \u251c\u2500\x1b[0m[\x1b[1m1\x1b[0m] (str)
+        \x1b[2m\u2502 \u2514\u2500\x1b[0m[\x1b[1m2\x1b[0m] (str)
+        \x1b[2m\u2514\u2500\x1b[0m\x1b[1mmeta\x1b[0m (dict)
+        \x1b[2m  \u251c\u2500\x1b[0m\x1b[1mdepends_on\x1b[0m (list)
+        \x1b[2m  \u2502 \u251c\u2500\x1b[0m[\x1b[1m0\x1b[0m] (str)
+        \x1b[2m  \u2502 \u251c\u2500\x1b[0m[\x1b[1m1\x1b[0m] (str)
+        \x1b[2m  \u2502 \u2514\u2500\x1b[0m[\x1b[1m2\x1b[0m] (str)
+        \x1b[2m  \u2514\u2500\x1b[0m\x1b[1mis_on_github\x1b[0m (bool)
+
         """
         tree = {
             key: value
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -827,31 +827,6 @@ class WeldxFile(UserDict):
         show_values :
             Set to `True` if primitive values should be displayed
 
-        Examples
-        --------
-
-        >>> from weldx import WeldxFile
-        >>> tree = dict(
-        ...     project="Weldx",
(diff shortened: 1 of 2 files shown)


vhirtham:
Here is the example output you get when using it on the IW example file:

~~~ 
root (dict)
├─TCP (LocalCoordinateSystem)
├─coordinate_systems (CoordinateSystemManager)
├─equipment (list)
│ ├─[0] (MeasurementEquipment)
│ ├─[1] (MeasurementEquipment)
│ └─[2] (MeasurementEquipment)
├─measurements (list)
│ ├─[0] (Measurement)
│ ├─[1] (Measurement)
│ ├─[2] (Measurement)
│ └─[3] (Measurement)
├─process (dict)
│ ├─shielding_gas (ShieldingGasForProcedure)
│ ├─weld_speed (TimeSeries)
│ ├─welding_process (GmawProcess)
│ └─welding_wire (dict)
│   ├─class (str)
│   ├─diameter (Q_)
│   └─wx_user (dict)
│     ├─charge id (str)
│     └─manufacturer (str)
├─reference_timestamp (Timestamp)
├─welding_current (Signal)
├─welding_voltage (Signal)
├─workpiece (dict)
│ ├─base_metal (dict)
│ │ ├─common_name (str)
│ │ └─standard (str)
│ └─geometry (dict)
│   ├─groove_shape (VGroove)
│   └─seam_length (Q_)
└─wx_user (dict)
  ├─WID (int)
  ├─operator (str)
  └─project (str)
~~~

weldx/asdf/file.py
```diff
@@ -812,6 +812,27 @@ def _ipython_display_(self):
...
+        max_rows :
+            The maximum number of rows that will be printed. If rows are cut, a
+            corresponding message will be printed
+        max_length :
+            The maximum line length. Longer lines will be truncated
+        show_values :
+            Set to `True` if primitive values should be displayed
+
+        """
+        tree = self._asdf_handle.tree.copy()
```
marscher, line 831: taking a copy just to remove history and asdf version seems a bit overkill, as the tree will be deeply copied. This seems prohibitive for large datasets.

marscher: Does asdf.info also triggers serialization? If so, we should encapsulate the info call in the private Visualizer class, as this implementation does not serialize the arrays, saving resources. Eventually we could add another switch for `asdf.info` or replace the pure text representation with it.

vhirtham: Since it is a function provided by asdf itself, I assume that it is not deserializing the data, but I am not sure.

Regarding the copy. I would also favor to avoid it, if there is another way to remove the dropped stuff. However, are you sure this is a deep copy and not a shallow one?

CagtayFabry: since we are working on the 'tree' it is already de-serialized into python types (asdf itself probably doesn't *load* array data but our classes most likely will because of xarray etc.) and really just a prettier `dict` viewer imo. No asdf serialization involved imo.
For performance it would be nice to have this kind of overview generated purely form the yaml header on 'unopened' files. But that would mean working with tags instead of classes etc (which would also be nice)

regarding the copy, I guess something like the following should work by creating only a few more pointers?
```suggestion
        tree = {key:value  for key, value in self._asdf_handle.tree if key not in ["asdf_library", "history"]}
```

marscher:
Thanks for this fix @CagtayFabry. Do you mind adding a small doctest/example for the new info method @vhirtham?

weldx/asdf/file.py
vhirtham: So after quick research, I guess my assumption that `dict.copy` just makes a shallow copy seemed to be right. [See here](https://stackoverflow.com/questions/3975376/understanding-dict-copy-shallow-or-deep) for example. However, since @CagtayFabry suggestion does the same thing and is guaranteed to be a shallow copy, let's just take that.

vhirtham:
> Do you mind adding a small doctest/example for the new info method @vhirtham?

I will write one. Just wanted to wait for your responses before bothering with doctest 😉

vhirtham:
Due to the special formatting of `asdf.info`, getting the doctest output right was a bit of a pain. However, I managed it using this little script below to print me the raw output send to `stdout`:

~~~ python
import sys
from io import StringIO

from weldx import WeldxFile

tree = dict(
    project="Weldx",
    developers=["Cagtay Fabry", "Volker Hirthammer", "Martin Scherer"],
    meta=dict(depends_on=["ASDF", "NumPy", "some more"], is_on_github=True),
)

# save old stdout handle and redirect output 
old_stdout = sys.stdout
sys.stdout = mystdout = StringIO()

# print some stuff
WeldxFile(tree=tree, mode="rw").info()

# restore old stdout and print raw string
sys.stdout = old_stdout
print(mystdout.getvalue().encode("unicode_escape"))
~~~

All you need to do now is to turn all `\\n` into an actual newline and replace all `\\` with `\`. Maybe I should move this in a special discussion regarding doctest in case someone faces this or a similar problem in the future.

vhirtham:
Looks pretty shitty [in the documentation](https://weldx--576.org.readthedocs.build/en/576/_autosummary/weldx.WeldxFile.info.html#weldx.WeldxFile.info) though...

vhirtham:
Haven't read it entirely yet, but maybe [this SO question](https://stackoverflow.com/questions/46465422/how-can-i-make-doctests-triggered-by-pytest-ignore-unicode-prefix-u-of-st) and its answer might help here

marscher:
If I would have known, that this would cause such an hassle, I would not have requested a doctest... Sorry for that. 
I guess we need to convince sphinx to properly handle the unicode characters there. Actually the string class of Python3 should handle unicode transparently without any manual intervention, but eventually there is a manual setting of the encoding breaking the proper handling.

marscher:
to redirect the stdout, I'd recommend the contextlib.redirect_output manager. If the test fails before resetting sys.stdout, all tests executed afterwards will still use a broken stdout reference.

CagtayFabry:
my thoughts:
- I don't think we need a doctest (or any test for that matter) here. It is simply a utility function that cannot be used for more than what it is. If we can get a "nice" sphinx output that looks like the actual console output but doesn't work with doctest then just use that.
- since this is basically a convenient wrapper around `asdf.info` it would be nice to have the same options (adding the `node_or_path` argument do look into specific parts of the file)
- we don't have to think too much about docstrings or documentation here, we can just link to `asdf.info` in the docstring

vhirtham:
> to redirect the stdout, I'd recommend the contextlib.redirect_output manager. If the test fails before resetting sys.stdout, all tests executed afterwards will still use a broken stdout reference.

it's not a test, just a script to print the unicode output. So no danger here 😉

vhirtham:
> * I don't think we need a doctest (or any test for that matter) here. It is simply a utility function that cannot be used for more than what it is. If we can get a "nice" sphinx output that looks like the actual console output but doesn't work with doctest then just use that.

Removed it entirely.

> * since this is basically a convenient wrapper around `asdf.info` it would be nice to have the same options (adding the `node_or_path` argument do look into specific parts of the file)

Don't think adding the `node_or_path` makes much sense here unless we want to reimplement the features on our own. `asdf.info` is a free function and `WeldxFile.info` a class method tied to a specific file. Therefore the path-related stuff is not really fitting here. So there only remains the node related stuff... not sure here either

> * we don't have to think too much about docstrings or documentation here, we can just link to `asdf.info` in the docstring

Doc is already there, so no need for any furthe actions.

marscher:
> Don't think adding the `node_or_path` makes much sense here unless we want to reimplement the features on our own. `asdf.info` is a free function and `WeldxFile.info` is a class method tied to a specific file. Therefore the path-related stuff is not really fitting here. 

Sure, if you only want to get an overview it is sufficient for now. But imagine a very large tree, it could make total sense to restrict the output to a given path. It does not hurt anybody to have this optional argument.

vhirtham:
The coverage will go up again once the tutorials are merged 😉

PR #578, opened 2021-10-04
Update single-pass weld schema

vhirtham:
## Changes

Update the single-pass weld schema to use a `TimeSeries` instead of a `Signal` for `welding_current` and `welding_voltage`

## Related Issues

See #559 

## Checks

- [x] updated CHANGELOG.rst
- [x] ~~updated tests~~
- [x] updated doc/

### DIFF
--- a/weldx/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-0.1.0.yaml
+++ b/weldx/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-0.1.0.yaml
@@ -167,17 +167,6 @@ examples:
     - A simple welding application
     - |
       !<tag:stsci.edu:asdf/core/asdf-1.1.0>
-        asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
-          name: asdf, version: 2.8.1}
-        history:
-          extensions:
-          - !core/extension_metadata-1.0.0
-            extension_class: asdf.extension.BuiltinExtension
-            software: !core/software-1.0.0 {name: asdf, version: 2.8.1}
-          - !core/extension_metadata-1.0.0
-            extension_class: weldx.asdf.extension.WeldxExtension
-            extension_uri: asdf://weldx.bam.de/weldx/extensions/weldx-0.1.0
-            software: !core/software-1.0.0 {name: weldx, version: 0.4.2.dev34+gf6e3c49.d20210820}
         TCP: !<asdf://weldx.bam.de/weldx/tags/core/transformations/local_coordinate_system-0.1.0>
           time: !<asdf://weldx.bam.de/weldx/tags/time/timedeltaindex-0.1.0>
             values: !core/ndarray-1.0.0
--- a/weldx/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-0.1.0.yaml
+++ b/weldx/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-0.1.0.yaml
@@ -82,12 +82,12 @@ properties:
   welding_current:
     description: |
       The signal representing the welding current measurement.
-    tag: "asdf://weldx.bam.de/weldx/tags/measurement/signal-0.1.*"
+    tag: "asdf://weldx.bam.de/weldx/tags/core/time_series-0.1.*"
     wx_unit: "A"
   welding_voltage:
     description: |
       The signal representing the welding voltage measurement.
-    tag: "asdf://weldx.bam.de/weldx/tags/measurement/signal-0.1.*"
+    tag: "asdf://weldx.bam.de/weldx/tags/core/time_series-0.1.*"
     wx_unit: "V"
   TCP:
     description: |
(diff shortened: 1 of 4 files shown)


vhirtham:
@CagtayFabry  Do you still have the script to update the example in the schema?

We should probably add these changes to the `0.5.0` release.

CagtayFabry:
> @CagtayFabry Do you still have the script to update the example in the schema?
> 
> We should probably add these changes to the `0.5.0` release.

that was all regex 🚀 
the example is the file generated from `asdf/cli/welding_schema.py` so I think just rebuild and use that

I was wondering why the asdf schema tests pass here but it makes sense because the schema tests don't know about any custom schema file to validate against etc.

vhirtham:
> that was all regex 🚀 the example is the file generated from `asdf/cli/welding_schema.py` so I think just rebuild and use that

okay, I will use this then
 
> I was wondering why the asdf schema tests pass here but it makes sense because the schema tests don't know about any custom schema file to validate against etc.

Was wondering about this too until I realized that custom schema examples are not validated 😄

CagtayFabry:
> > I was wondering why the asdf schema tests pass here but it makes sense because the schema tests don't know about any custom schema file to validate against etc.
> 
> Was wondering about this too until I realized that custom schema examples are not validated 😄

I think the examples do get validated ( the second dot of the test output [here](https://github.com/BAMWelDX/weldx/pull/578/checks?check_run_id=3800524592#step:5:49)) but it will only validate the tagged objects in the tree and not the overall schema layout (because the test doesn't know we want to use the file as a  `custom_schema`)

vhirtham:
> > > I was wondering why the asdf schema tests pass here but it makes sense because the schema tests don't know about any custom schema file to validate against etc.
> > 
> > 
> > Was wondering about this too until I realized that custom schema examples are not validated 😄
> 
> I think the examples do get validated ( the second dot of the test output [here](https://github.com/BAMWelDX/weldx/pull/578/checks?check_run_id=3800524592#step:5:49)) but it will only validate the tagged objects in the tree and not the overall schema layout (because the test doesn't know we want to use the file as a `custom_schema`)

That's what I meant.

weldx/schemas/weldx.bam.de/weldx/datamodels/single_pass_weld-0.1.0.yaml
```diff
@@ -327,13 +329,10 @@ examples:
...
-              start: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M0S
-              end: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M10S
-              freq: 2S
-              min: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M0S
-              max: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M10S
-            units: !<asdf://weldx.bam.de/weldx/tags/unit/unit-0.1.0> A
+            time: !<asdf://weldx.bam.de/weldx/tags/time/timedeltaindex-0.1.0> {start: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M0S,
+              end: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M10S, freq: 2S, min: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M0S,
+              max: !<asdf://weldx.bam.de/weldx/tags/time/timedelta-0.1.0> P0DT0H0M10S}
+            units: !<asdf://weldx.bam.de/weldx/tags/unit/unit-0.1.0> ampere
```
CagtayFabry, line 335: I don't know why the formatting changed here but it doesn't really matter

CagtayFabry: we can probably force it with `flowStyle: block`, I'll add that back in some other time

CagtayFabry:
> Just a reminder, If we change a schema, shouldn't we begin to increase the version number of the schema? Now files created with the old version will just fail to validate against the same version, which is kind of confusing. Just a quick question, is there an ASDF way of upgrading trees according to a schema (like a Converter)?

yes, if we make schemas we should change the schema name (=increase version numbers)
For 0.5 the version change is included in changing the asdf tags to `asdf://`

I only consider the released weldx version as 'stable' in the sense that they should work, I expect changes on master between weldx to break things rather often (until the release)

For the file-schema here, if you want to load an `0.4.x` file and convert it to `0.5.0` the following should work (load legacy file, change tree, save as new file)
```python
from pathlib import Path

from weldx import WeldxFile
from weldx.asdf.util import get_schema_path

file_schema = get_schema_path("single_pass_weld-1.0.0.schema.yaml") # legacy schema to validate

compat_file = Path("old.weldx") # path to version 0.4 file
new_file = Path("new.weldx") # path to save new 0.5 file
with WeldxFile(
    compat_file,
    mode="r",
    custom_schema=file_schema,
    sync=False,
    asdffile_kwargs={"copy_arrays": True},
) as file_04:
    file_04["welding_current"] = file_04["welding_current"].data # need to change this for new schema
    file_04["welding_voltage"] = file_04["welding_voltage"].data # need to change this for new schema
    file_05 = WeldxFile(tree=file_04, mode="rw", custom_schema="single_pass_weld-0.1.0") # validate against new schema
    if new_file.exists():
        new_file.unlink()
    file_05.write_to(new_file)
```

if you want to convert a file that was contructed with the current master, the following should work:
- load the file **without** validating against the new schema
- make changes to the tree (s.a.)
  ```python
    tree["welding_current"] = tree["welding_current"].data # need to change this for new schema
    tree["welding_voltage"] = tree["welding_voltage"].data # need to change this for new schema
  ```
- save to new file (or overwrite old) and use schema validation with current schema again


let me know if it works

PR #583, opened 2021-10-06
Geometry enhancements

vhirtham:
## Changes

Apply some small enhancements to the `Geometry` class. Most of them help to improve the readability/understandability of the new tutorials.

- [ ] create `Geometry` from a profile and extrusion length
- [ ] `__init__` also accepts groove types
- [ ] add `k3d` as backend option to `plot`
- [ ] support unit strings in interfaces
- [ ] OPTIONAL: Set pint wrapper to strict
- [ ] Add new triangulation algorithm (found one that works for all polygon types)

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/geometry.py
+++ b/weldx/geometry.py
@@ -23,6 +23,8 @@ _DEFAULT_ANG_UNIT = UREG.rad
 if TYPE_CHECKING:  # pragma: no cover
     import matplotlib.axes
 
+    import weldx.welding.groove.iso_9692_1 as iso
+
 # helper -------------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


weldx/geometry.py
```diff
@@ -2023,6 +2025,31 @@ def __init__(self, profile, trace):
...
+            The length of the extrusion
+
+        Returns
+        -------
+        weldx.geometry.Geometry :
+            The new geometry
+
+        """
+        trace = Trace(LinearHorizontalTraceSegment(length))
+        return cls(profile, trace)
```
CagtayFabry, line 2051: seeing this I had the following idea:
what about we make `__init__` also accept a `pint.Quantity[length]` for trace and in that case it auto generates the `Trace(LinearHorizontalTraceSegment(length))`

no need to remember an extra function name in that case :)

CagtayFabry:
> 
> 
> > did you do anything specific to fix this error? https://github.com/BAMWelDX/weldx/pull/583/checks?check_run_id=3827087537#step:10:178 not sure why it sometimes keep popping up, should be unrelated here and my guess is some kind of wrong intersphinx loading
> 
> Didn't try it so far, but busting those warnings "the right way" is extremely time-consuming. Try a fix and wait 5 minutes to see if it worked. If we want to get a warning-free build, I can just put it into the nitpick ignore. Is that okay?



> 
> 
> > did you do anything specific to fix this error? https://github.com/BAMWelDX/weldx/pull/583/checks?check_run_id=3827087537#step:10:178 not sure why it sometimes keep popping up, should be unrelated here and my guess is some kind of wrong intersphinx loading
> 
> Didn't try it so far, but busting those warnings "the right way" is extremely time-consuming. Try a fix and wait 5 minutes to see if it worked. If we want to get a warning-free build, I can just put it into the nitpick ignore. Is that okay?

it's probably a good use for the nitpick ignore 👍

PR #588, opened 2021-10-07
Update unit handling of geometry package

vhirtham:
## Changes

Turn all unit decorators to strict mode. This enforces the usage of units and allows us to pass unit strings to the function without any further modifications

**What that means:**

- All public interfaces only accept quantities or strings that can be converted to a quantity as inputs. Pure numbers are rejected
- All public interfaces return a `pint.Quantity` (a.k.a. `Q_`) if the returned value represents a quantity
- `Groove` types are based on the `geometry` types and therefore require indirectly the usage of units too. However, I have not encountered them being used without units anywhere in our code, so they should be fine.
- Internally, the classes work without units, since the public interfaces enforce a consistent magnitude

## ToDo
- [x] ~~`SpatialData` should check unit of the data array~~ postponed
- [x] Purge new warnings
- [x] Check correct behavior of rasterize functions

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/


CagtayFabry:
just a small hint if `Q_(value, units)` sometimes looks ugly in the code, we should get the same result using `value * U_(units)`

vhirtham:
> > Tests and notebooks are finally working again. Tomorrow, I will get rid of those 300 Warnings, finish the remaining todos from the list above and then I will rip the `Q`, `_`, `"` and `M` keys from my keyboard 🤣
> 
> looking at all those warnings that seems like **a lot** of stuff to rewrite let's get in touch tomorrow morning

Its almost always the same warning emitted by the `__init__ ` method. So it shouldn't be too much

vhirtham:
The failing deepsource test is not related to the unit update but something introduced earlier. I think we can ignore it for now.

PR #590, opened 2021-10-11
Update pytest actions

CagtayFabry:
## Changes

Main goal is to make the different workflows easier to read/edit and manage in general

- split jobs between ubuntu (+ python matrix) and mac/win (only run default py 3.8 for now)
  - add `detect-ci-trigger` job prior to running on win/macos
- add publish step for unit test results
- cancel old workflow runs with new commits
- move notebook check to static-analysis workflow
- remove black, flake8 and pydocstyle jobs from static-analysis workflow

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,7 +12,7 @@ on:
     - cron: '0 6 * * 1'
 
 jobs:
-  detect-ci-trigger: # only run win/max pytest job if review PR or on master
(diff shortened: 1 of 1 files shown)


CagtayFabry:
fyi: some of the "new features" like canceling old workflows and adding test results to PR comments are only enabled once the workflow here is on master @vhirtham @marscher (at least I think this is how it works 😉 )

PR #603, opened 2021-10-19
Time improvements

vhirtham:
## Changes

- Adds `duration` property to get the covered time duration
- Adds `resample` method to get a resampled time object with the same time limits as the original object
- Fix the `all_close` method

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/


marscher:
Overall this is a nice improvement! Thank you! I'm just a bit curious where the additional overhead arises from.

vhirtham:
From the article I linked: "differences of as much as 50% are common"

PR #618, opened 2021-10-27
Add `GenericSeries`

### LINKED ISSUE #106: create GenericSeries class
create a more basic implementation of what the `TimeSeries` currently does:

- support discrete xarray or MathExpression style
- choose input dimensionality (`TimeSeries` would be "s")
- optionally fix output shape/coordinates (`SpatialSeries` would be `[t,3]` / `"x","y","z"`)

I think we should keep the `TimeSeries` as is for now and implement a pint-only generic series

vhirtham:
## Changes

Add new `GenericSeries` class

## Related Issues

Closes #106

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/
- [x] update manifest file


weldx/core.py
```diff
@@ -1006,7 +1009,8 @@ def __call__(self, **kwargs) -> GenericSeries:
                     v, dims=self._symbol_dims[k], coords={self._symbol_dims[k][0]: v.m}
                 )
             else:
-                v = v.m
+                ref_unit = self._data.coords[k].attrs["units"]
```
CagtayFabry, line 1012:
```suggestion
                ref_unit = self._data.coords[k].pint.units
```
can you do a check check if this works @vhirtham ?

vhirtham: Well, let's see what happens if I commit this 😉

vhirtham: Doesn't seem to work :(
The tutorial notebook crashes in code cell 10.
If i print the outputs as follows: 

~~~ python
ref_unit = self._data.coords[k].pint.units
print(ref_unit)
print(self._data.coords[k].attrs.get("units", ""))
v = xr.DataArray(v.to(ref_unit), dims=[k]).pint.dequantify()
~~~

I get:

~~~
None
s
~~~

So unfortunately, I guess the `units` property doesn't work if the unit is stored in the attributes.

CagtayFabry: yes I realized now too, that's a bummer

vhirtham:
The unit handling seems to work pretty well now. Some tests failed (correctly) due to the wrong units being used 😉

weldx/core.py
```diff
@@ -151,6 +153,7 @@ def set_parameter(self, name, value):
         """
         self.set_parameters({name: value})
 
+    # todo: Use kwargs here???
     def set_parameters(self, params: _me_parameter_types):
```
CagtayFabry, line 157: I think using kwargs would be easier to read here 👍

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+#  - use dimensionality errors
+
+
+class GenericSeries:
+    """Describes a quantity depending on one or more parameters."""
+
+    _allowed_variables: List[str] = None
+    """Allowed variable names"""
+    _required_variables: List[str] = None
+    """Required variable names"""
```
CagtayFabry, line 734: as discussed, the default values should be empty containers of their respective types

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+
+    # todo: other possible constraints:
+    # - allowed dimensions (not the same as allowed variables)
+
+    @classmethod
+    def _check_constraints(cls, dims, units_out: pint.Unit = None):
+        if cls._required_dimensions is not None:
+            for v in cls._required_dimensions:
+                if v not in dims:
+                    raise ValueError(f"{cls.__name__} requires dimension '{v}'.")
```
CagtayFabry, line 764: something like
```python
if not set(dims).issubset(_required_dimensions)
    raise
```

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+                raise ValueError(
+                    f"{cls.__name__} requires its output unit to be of dimensionality "
+                    f"'{cls._required_unit_dimensionality}' but it actually is "
+                    f"'{units_out.dimensionality}'."
+                )
+
+    @classmethod
+    def _check_constraints_discrete(cls, data_array: xr.DataArray):
+        if cls is GenericSeries:
+            return
```
CagtayFabry, line 776: Everything we need here should be added to `xr_check_coords` I think

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+
+        cls._check_constraints(data_array.dims, data_array.data.u)
+
+        if cls._required_dimension_units is not None:
+            for k, v in cls._required_dimension_units.items():
+                if k in data_array.dims and (
+                    k not in data_array.coords.keys()
+                    or U_(data_array.coords[k].attrs.get("units", "")).dimensionality
```
CagtayFabry, line 784:
```suggestion
                    or U_(data_array.coords[k].attrs.get(UNITS_KEY, "")).dimensionality
```
use `weldx.constants.UNITS_KEY` verywhere for attributes access

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+    def _check_constraints_discrete(cls, data_array: xr.DataArray):
+        if cls is GenericSeries:
+            return
+
+        cls._check_constraints(data_array.dims, data_array.data.u)
+
+        if cls._required_dimension_units is not None:
+            for k, v in cls._required_dimension_units.items():
+                if k in data_array.dims and (
+                    k not in data_array.coords.keys()
```
CagtayFabry, line 783:
```suggestion
                    k not in data_array.coords
```
`dict`syntax should work here

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+
+        if cls._required_dimension_coordinates is not None:
+            for k, v in cls._required_dimension_coordinates.items():
+                if k not in data_array.coords.keys() or not ut.compare_nested(
+                    data_array.coords[k].data.tolist(), v
+                ):
+                    raise ValueError(
+                        f"{cls.__name__} requires dimension {k} to have the "
+                        f"coordinates {v}"
+                    )
```
CagtayFabry, line 800: if we stick to the xarray syntax here the naming is slightly confusing to me.
Dimensions cannot have values, if they do they are already promited to coordinates.

So  `_required_coordinates_values` would be a better name over `_required_dimension_coordinates`

regardless, I think we have this functionality already in `xr_check_coords` (using `values`)

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+                )
+
+    @classmethod
+    def _check_constraints_discrete(cls, data_array: xr.DataArray):
+        if cls is GenericSeries:
+            return
+
+        cls._check_constraints(data_array.dims, data_array.data.u)
+
+        if cls._required_dimension_units is not None:
```
CagtayFabry, line 780: this check should be covered by `xr_check_coords` with  `"dimensionality"`

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+                units[v] = U_("")
+
+        return dims, units
+
+    def _init_expression(self, expr, dims, parameters, units):
+        """Initialize the internal data with a mathematical expression."""
+        # Check and update expression
+        if isinstance(expr, MathematicalExpression):
+            parameters = expr.parameters
+            expr = expr.expression.__repr__()
```
CagtayFabry, line 955: This should be `str(expr.expression)` instead of using `__repr__`

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+        -------
+        GenericSeries :
+            A new generic series containing discrete values for the interpolated
+            dimensions.
+
+        """
+        raise NotImplementedError
+
+    @property
+    def coordinates(self) -> Union[None, pint.Quantity, Dict[str, pint.Quantity]]:
```
CagtayFabry, line 1169: I suggest outputting a similar format to the xarray coords object (so `Dict[str, pint.Quantity]`

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+        return self._obj.dims
+
+    @property
+    def interpolation(self) -> str:
+        """Get the name of the used interpolation method."""
+        return self._interpolation
+
+    @interpolation.setter
+    def interpolation(self, val: str):
+        if val not in ["linear", "step"]:
```
CagtayFabry, line 1221: we should double check interpolation methods available with array (didn't we define them somewhere already?)

it probably makes sense to distinguish between 1d/nd cases

CagtayFabry (changes requested):

weldx/core.py
vhirtham: DeepSource agrees with you 😉

weldx/core.py
```diff
@@ -708,3 +711,557 @@ def units(self) -> pint.Unit:
...
+        #  - tuple should accept third element (coords)
+        for k, v in params.items():
+            if isinstance(v, tuple):
+                v = (Q_(v[0]), v[1])
+            elif isinstance(v, xr.DataArray):
+                if not isinstance(v.data, pint.Quantity):
+                    raise ValueError(f"Value for parameter {k} is not a quantity.")
+            else:
+                v = Q_(v)
+            params[k] = v
```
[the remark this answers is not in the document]
vhirtham: As mentioned, docstring update will be done after the clean-up. 

The function modifies the dict in place. Therefore it does not need a return value. Its purpose is mainly to assure that every possible input format to the `MathematicalExpression` contains quantities. For example, the first branch keeps the tuple, only the first entry is converted (This can/will be rewritten a bit more direct)

weldx/core.py
vhirtham: Well, most/all of the required fields that aren't already in `xr_check_coords` are not related to coordinates, so I think it doesn't make sense to add them there. However, I already added a new xarray utility function

weldx/core.py
vhirtham: I think `_required_coordinates` is even shorter and also fits the bill. What do you think?

CagtayFabry: also fine

weldx/util/xarray.py
```diff
@@ -529,13 +530,47 @@ def xr_check_coords(dax: xr.DataArray, ref: dict) -> bool:
...
+    Raises
+    ------
+    pint.DimensionalityError:
+        The error is raised if the check fails
+
+    """
+    if units_ref is None:
+        return
+    if not isinstance(units_ref, pint.Unit):
+        units_ref = U_(units_ref)
```
CagtayFabry, line 559: I think be default it's ok to simply cast `units_ref = U_(units_ref)` at the beginning without type checking checking

weldx/util/xarray.py
```diff
@@ -529,13 +530,47 @@ def xr_check_coords(dax: xr.DataArray, ref: dict) -> bool:
...
+
+    units = da.attrs.get(UNITS_KEY, None)
+    if units is None and isinstance(da.data, pint.Quantity):
+        units = da.data.u
```
CagtayFabry, line 563: Here I think it would be safer to first check the actual data for quantity and only check the attributes after (if we have a quantity that unit should take precedence over the attributes)
```suggestion
    units = da.pint.units
    if units is None:
        units = da.attrs.get(UNITS_KEY, None)
```

CagtayFabry:
> The way it works now, the serialized `GenericSeries` based on expressions would look like this:
> 
> ```yaml
> gs: !weldx!core/generic_series-0.1.0
>   expression: a*t + c*x
>   free_dimensions:
>     space: {symbol: x, units: !weldx!units/units-0.1.0 meter}
>     time: {symbol: t, units: !weldx!units/units-0.1.0 second}
>   parameters:
>     a: !weldx!core/data_array-0.1.0
>       attributes: {}
>       coordinates: []
>       data: !weldx!core/variable-0.1.0
>         name: data
>         dimensions: [velocity]
>         dtype: <i4
>         units: !weldx!units/units-0.1.0 meter / second
>         data: !core/ndarray-1.0.0
>           data: [1, 2]
>           datatype: int32
>           shape: [2]
>     c: !weldx!units/quantity-0.1.0
>       value: !core/ndarray-1.0.0
>         data: [2, 4]
>         datatype: int32
>         shape: [2]
>       units: !weldx!units/units-0.1.0 dimensionless
>   units: !weldx!units/units-0.1.0 meter
> ```
> 
> The `units` field is just informal and not really necessary for deserialization. I renamed the `dimensions` field to `free_dimensions` to highlight that there might be other dimensions contained in parameters like the acceleration in the example above. I also "flipped" the relation between dimension symbol and name when compared to the internal data of the `GenericSeries`. I think this way it is much clearer than something like this:
> 
> ```yaml
>   free_dimensions:
>     x: {full_name: space, units: !weldx!units/units-0.1.0 meter}
>     t: {full_name: time, units: !weldx!units/units-0.1.0 second}
> ```
> 
> Let me knowe what you think.

That looks pretty epic to me 😁 🚀

vhirtham:
> That looks pretty epic to me 😁 🚀

Good, then I will write the schema accordingly to match this output 😉

CagtayFabry:
after giving it some thought I think we can skip implementing `.sel` for now @vhirtham 
it would only make sense for arrays and those we can access so no real benefit.
I think we simply remove `__get_item__` and leave it at that for now

weldx/schemas/weldx.bam.de/weldx/core/generic_series_free_dimension-0.1.0.yaml
```diff
@@ -0,0 +1,24 @@
...
+properties:
+  symbol:
+    description: |
+      The symbol of the dimension that is used in the expression. If the dimension name and symbol are identical, this
+      property can be omitted
+    type: string
+  units:
+    description: |
+      A unit that represents the dimensions unit dimensionality.
+    tag: "asdf://weldx.bam.de/weldx/tags/units/units-0.1.*"
```
CagtayFabry, line 21: Do I understand it correctly that we wouldn't need this schema/tag if we omit the mapping between name and symbol here?
And we could simply do a name -> units mapping with `wx_property_tag: "asdf://weldx.bam.de/weldx/tags/units/units-0.1.*"` ?

In that case maybe we implement this functionality using the `_alias_names` ?

CagtayFabry:
> Examples are still missing and I need to tweak the tutorial a bit, but the rest is ready for review.

maybe best to wait with finishing the examples and tutorial for after the code review

CagtayFabry:
the docs build errors seem unrelated as they also happen on [master](https://github.com/BAMWelDX/weldx/runs/4570280776?check_suite_focus=true)

CagtayFabry (approved):

CagtayFabry:
unfortunately the doctest output is slightly broken because of the `Parameters` keyword

https://weldx--618.org.readthedocs.build/en/618/_autosummary/weldx.GenericSeries.html#weldx.GenericSeries

PR #621, opened 2021-10-29
Add xarray support to `MathematicalExpression`

### LINKED ISSUE #619: `MathematicalExpression` should support `xarray.DataArrays`
I have not fully thought this through and there might be some issue that prevents my suggestion, but it would be pretty nice if the `MathematicalExpression` would support `DataArray`s as parameters and values (during evaluation). This would make the implementation of the `GenericSeries` much easier because we can let xarray handle all the dimension-related stuff. Additional benefit: The result of the evaluation should also be a  `xarray.DataArray`  which could be directly passed to `GenericSeries.__init__` (useful when interpolating a `GenericSeries`).

vhirtham:
## Changes

`MathematicalExpression` now uses xarray internally. This can simplify multi-dimensional expressions if one uses `DataArray`s with correctly named dimensions. If no `DataArray`s are used as parameters and variables, the behavior should remain more or less the same since xarray will automatically add dimension names.

In the current version `evaluate` will now return a `DataArray` instead of a quantity. This requires some changes to the internals of the `TimeSeries`. However, when those changes are fully applied, the `TimeSeries` implementation should be much simpler than before.

## Related Issues

Closes #619

## Checks

- [ ] update schema
- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -47,7 +47,8 @@ class MathematicalExpression:
         self.function = sympy.lambdify(
             tuple(self._expression.free_symbols), self._expression, "numpy"
         )
-        self._parameters = {}
+
+        self._parameters: Dict[str, Union[pint.Quantity, xr.DataArray]] = {}
         if parameters is not None:
             self.set_parameters(parameters)
 
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -410,8 +410,6 @@ class TimeSeries(TimeDependent):
             time = Time(time)
 
             self._reference_time = time.reference_time
-            print(data)
-            print(time)
             self._data = self._create_data_array(data, time)
         self.interpolation = interpolation
 
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -51,16 +51,23 @@ class MathematicalExpression:
         if parameters is not None:
             self.set_parameters(parameters)
 
-    def set_parameters(self, parameters):
-        if not isinstance(parameters, dict):
-            raise ValueError(f'"parameters" must be dictionary, got {type(parameters)}')
+    def set_parameters(self, params: Dict[str, Union[pint.Quantity, xr.DataArray]]):
+        """Set the expressions parameters.
+
+        Parameters
+        ----------
+        paras:
+            Dictionary that contains the values for the specified parameters.
+
+        """
+        if not isinstance(params, dict):
+            raise ValueError(f'"parameters" must be dictionary, got {type(params)}')
         variable_names = [str(v) for v in self._expression.free_symbols]
-        for k, v in parameters.items():
+        for k, v in params.items():
             if k not in variable_names:
                 raise ValueError(f'The expression does not have a parameter "{k}"')
             if not isinstance(v, xr.DataArray):
                 v = Q_(v)
-
             self._parameters[k] = v
 
     def __repr__(self):
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -57,7 +57,7 @@ class MathematicalExpression:
 
         Parameters
         ----------
-        paras:
+        params:
             Dictionary that contains the values for the specified parameters.
 
         """
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -467,7 +467,7 @@ class TimeSeries(TimeDependent):
         """Interpolate the time series if its data is composed of discrete values."""
         return ut.xr_interp_like(
             self._data,
-            {"time": time.as_timedelta_index()},
+            {"time": time.as_data_array()},
             method=self.interpolation,
             assume_sorted=False,
             broadcast_missing=False,
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -52,25 +52,6 @@ class MathematicalExpression:
         if parameters is not None:
             self.set_parameters(parameters)
 
-    def set_parameters(self, params: Dict[str, Union[pint.Quantity, xr.DataArray]]):
-        """Set the expressions parameters.
-
-        Parameters
-        ----------
-        params:
-            Dictionary that contains the values for the specified parameters.
-
-        """
-        if not isinstance(params, dict):
-            raise ValueError(f'"parameters" must be dictionary, got {type(params)}')
-        variable_names = [str(v) for v in self._expression.free_symbols]
-        for k, v in params.items():
-            if k not in variable_names:
-                raise ValueError(f'The expression does not have a parameter "{k}"')
-            if not isinstance(v, xr.DataArray):
-                v = Q_(v)
-            self._parameters[k] = v
-
     def __repr__(self):
         """Give __repr__ output."""
         representation = (
@@ -166,6 +147,27 @@ class MathematicalExpression:
         """
         self.set_parameters({name: value})
 
+    def set_parameters(self, params: Dict[str, Union[pint.Quantity, xr.DataArray]]):
+        """Set the expressions parameters.
+
+        Parameters
+        ----------
+        params:
+            Dictionary that contains the values for the specified parameters.
+
+        """
+        if not isinstance(params, dict):
+            raise ValueError(f'"parameters" must be dictionary, got {type(params)}')
+
+        variable_names = [str(v) for v in self._expression.free_symbols]
+
+        for k, v in params.items():
+            if k not in variable_names:
+                raise ValueError(f'The expression does not have a parameter "{k}"')
+            if not isinstance(v, xr.DataArray):
+                v = Q_(v)
+            self._parameters[k] = v
+
     @property
     def num_parameters(self):
         """Get the expressions number of parameters.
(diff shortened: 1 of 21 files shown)


weldx/core.py
```diff
@@ -458,21 +461,21 @@ def _interp_time_discrete(self, time: Time) -> xr.DataArray:
...
+        # data = data.astype(float).to_reduced_units()  # float conversion before reduce!
 
         # create data array
-        if not np.iterable(data):  # make sure quantity is not scalar value
-            data = np.expand_dims(data, 0)
-
-        return self._create_data_array(data, time)
+        # if not np.iterable(data):  # make sure quantity is not scalar value
+        #    data = np.expand_dims(data, 0)
+        data = data.assign_coords({"time": time.as_timedelta_index()})
```
CagtayFabry, line 477: do we loose the reference_timestamp information here?

I think
```suggestion
        data = data.assign_coords({"time": time.as_data_array()})
```
should also work (or manually setting the reference time)

vhirtham: Should work. However, there is this `()` shape problem again I have to fix. Apart from that I think we are good to go

weldx/core.py
```diff
@@ -459,20 +475,14 @@ def _interp_time_discrete(self, time: Time) -> xr.DataArray:
...
-        data = self._data.evaluate(**{self._time_var_name: time_q})
-        data = data.astype(float).to_reduced_units()  # float conversion before reduce!
-
-        # create data array
-        if not np.iterable(data):  # make sure quantity is not scalar value
-            data = np.expand_dims(data, 0)
-
-        return self._create_data_array(data, time)
+        data = self._data.evaluate(**{self._time_var_name: time_xr})
+        return data.assign_coords({"time": time.as_timedelta_index()})
```
CagtayFabry, line 485: I guess we don't really support reference times with `TimeSeries` using ME?
Still I think we should keep that information whenever possible to make future implementations easier

weldx/tests/test_core.py
```diff
@@ -220,7 +220,7 @@ class TestTimeSeries:
     me_expr_str = "a*t + b"
     me_params = {"a": Q_(2, "m/s"), "b": Q_(-2, "m")}
 
-    me_params_vec = {"a": Q_([[2, 0, 1]], "m/s"), "b": Q_([[-2, 3, 0]], "m")}
+    me_params_vec = {"a": Q_([2, 0, 1], "m/s"), "b": Q_([-2, 3, 0], "m")}
```
CagtayFabry, line 223: this is so much easier 😎  👍

CagtayFabry (approved):

weldx/core.py
vhirtham: Agreed. I just tried to change that, but unfortunately, some tests are failing because of "non-monotonic" indices and I couldn't figure out why by just looking at it. Maybe we should make an issue and do this in another PR. At least I got it working for the ME at the position you marked above.

weldx/tests/test_core.py
vhirtham: Indeed, it is. And the code in the ME is also cleaner 😄

weldx/core.py
vhirtham: Updated it. Just needed to adjust the implementation of `Time.as_data_array` because single time values caused failures. I replaced the usage of `self._time` with `self.as_pandas_index` when constructing the `DataArray`

weldx/core.py
CagtayFabry: the non-monotonic error surprises me, do we have a CI run with that popping up?

vhirtham: no, but I just realized, that I made an error. So it might not be an issue.

vhirtham:
I updated the schema and did some small refactoring. @CagtayFabry You might want to have a look at the changes. If there are no further concerns, I will merge this PR after lunch.

weldx/schemas/weldx.bam.de/weldx/core/mathematical_expression-0.1.0.yaml
```diff
@@ -32,7 +32,9 @@ properties:
     description: |
       List of constant parameters that to be set in the mathematical expression.
     type: object
-    wx_property_tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.*"
+    anyOf:
+      - wx_property_tag: "asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"
+      - wx_property_tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.*"
```
CagtayFabry, line 37: I would prefer if we don't have data_array here as it is very verbose, that should still be possible right? @vhirtham

CagtayFabry (changes requested):

CagtayFabry:
> I updated the schema and did some small refactoring. @CagtayFabry You might want to have a look at the changes. If there are no further concerns, I will merge this PR after lunch.

if we want to allow data_array tags in the schema we should bump the schema version (which I would like to avoid for the 0.5.x releases)

weldx/schemas/weldx.bam.de/weldx/core/mathematical_expression-0.1.0.yaml
vhirtham: Well, the solution would be to not use data arrays as parameters if you don't need to. But if I have a multi-dimensional equation that uses data arrays as parameters to make my life easier, I can't store it in asdf unless I sort out the dimensions myself and restore the parameters as quantities.

vhirtham:
> > I updated the schema and did some small refactoring. @CagtayFabry You might want to have a look at the changes. If there are no further concerns, I will merge this PR after lunch.
> 
> if we want to allow data_array tags in the schema we should bump the schema version (which I would like to avoid for the 0.5.x releases)

Since the schema is now less restrictive, no existing file will break. Furthermore, the current userbase is just us. So I think we can get away with not bumping the version.

CagtayFabry:
> > > I updated the schema and did some small refactoring. @CagtayFabry You might want to have a look at the changes. If there are no further concerns, I will merge this PR after lunch.
> > 
> > 
> > if we want to allow data_array tags in the schema we should bump the schema version (which I would like to avoid for the 0.5.x releases)
> 
> Since the schema is now less restrictive, no existing file will break. Furthermore, the current userbase is just us. So I think we can get away with not bumping the version.

true, but I find the xarray schema for simple things like this borderline unreadable in the schema
Also I am not sure how far the implementation for `wx_unit` validation is with xarray which I figure would be very important here

can some parameters be passed as xarray and others as quantities in the current python implementation?

weldx/core.py
```diff
@@ -450,7 +467,7 @@ def _interp_time_discrete(self, time: Time) -> xr.DataArray:
         """Interpolate the time series if its data is composed of discrete values."""
         return ut.xr_interp_like(
             self._data,
-            {"time": time.as_timedelta()},
+            {"time": time.as_timedelta_index()},
```
CagtayFabry, line 470: we should pass the reference time here as well

vhirtham: well, if you use `as_data_array` here, you get the "non-monotonic" error I mentioned above 😉

vhirtham:
> > > > I updated the schema and did some small refactoring. @CagtayFabry You might want to have a look at the changes. If there are no further concerns, I will merge this PR after lunch.
> > > 
> > > 
> > > if we want to allow data_array tags in the schema we should bump the schema version (which I would like to avoid for the 0.5.x releases)
> > 
> > 
> > Since the schema is now less restrictive, no existing file will break. Furthermore, the current userbase is just us. So I think we can get away with not bumping the version.
> 
> true, but I find the xarray schema for simple things like this borderline unreadable in the schema 

Why in the schema? Did you mean the written file?
My opinion: The goal should always be to get full functionality with good readability. However, if we can't have both, then functionality is way more important to me.

>Also I am not sure how far the implementation for `wx_unit` validation is with xarray which I figure would be very important here

If this is a problem it needs to be addressed in a follow-up issue/PR. 


> can some parameters be passed as xarray and others as quantities in the current python implementation?

Yes. Every parameter is stored internally (and also written to asdf) as a quantity unless it is a `DataArray`, which will be stored as `DataArray`. Only during evaluation, everything is promoted to a `DataArray` to perform the calculation. But that doesn't affect the internal variables

CagtayFabry:
> Yes. Every parameter is stored internally (and also written to asdf) as a quantity unless it is a `DataArray`, which will be stored as `DataArray`. Only during evaluation, everything is promoted to a `DataArray` to perform the calculation. But that doesn't affect the internal variables

In that case I guess the schema wouldn't work for 'mixed' parameters anyway since `wx_property_tag` doesn't support this

I'll catch up with you later so we can discuss in detail :)

vhirtham:
> > Yes. Every parameter is stored internally (and also written to asdf) as a quantity unless it is a `DataArray`, which will be stored as `DataArray`. Only during evaluation, everything is promoted to a `DataArray` to perform the calculation. But that doesn't affect the internal variables
> 
> In that case I guess the schema wouldn't work for 'mixed' parameters anyway since `wx_property_tag` doesn't support this
> 
> I'll catch up with you later so we can discuss in detail :)

All right, I will take care of the tutorials in the meantime.

CagtayFabry (approved):
looks really clean now 👍

PR #625, opened 2021-11-02
protect keys from writing/deleting/manipulating

### LINKED ISSUE #557: weldxfile: consider removing asdf metadata from user dictionary like operations.
During writing examples (doctests) for WeldxFile, I noticed that it would be bad, if users are able to delete basic asdf meta data like history, software versions etc.
So I'd suggest to remove these data from the users view, restricting it to the data he or she has actually wanted to be in the file (dictionary). Of course the data is still there and will be written as such, but just became immutable. For `history` we already have an interface. So we'd need to another getter for software.

marscher:
## Changes

Protect some keys from reading/writing/deleting/manipulating them. There is a list of attributes in file.py, which are being passed to the protected key dictionary. The WeldxFile now derives from it.
We need to deprecate the `data` attribute, as it allowed raw access.

## Related Issues

Closes #557

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests


marscher:
does this "arguments differ" warnings raised by codacy origin from pylint? I cannot figure out which tool they use. The argument has been renamed from the top-level typing interface named "__m" to "mapping" for better user-experience. I can also add an ignore for this.
Another possibility which might have triggered this is the different (more sane) type annotation. I would also refuse to change that, because it makes more sense to require a key to be hashable in contrast to the super broad TypeVar("_K").

CagtayFabry:
it looks like the pylint run but I am fine with this kind of error (or adding a simple ignore 👍 )

vhirtham:
If you expand the error box, the second line is: 

~~~
Reported by Pylint (Python 3) Time to fix: 5 minutes
~~~

So as @CagtayFabry  said, its pylint. Just add `# pylint: disable=W0221` if you think we can ignore the issue

CagtayFabry (approved):
great, thanks! 🚀

weldx/asdf/file.py
```diff
@@ -59,9 +70,15 @@ def reset_file_position(fh: SupportsFileReadWrite):
 DEFAULT_ARRAY_COPYING = True
 """Stored Arrays will be copied to memory, or not. If False, use memory mapping."""
 
+_PROTECTED_KEYS = (
+    "history",
+    "asdf_library",
+)
+"""These keys are not not seen, nor can they be manipulated."""
```
vhirtham, line 77: > nor can they be manipulated

Would you bet on this? 😄

weldx/asdf/util.py
```diff
@@ -542,6 +553,80 @@ def _get_instance_shape(
...
+            }
+
+        self._data.update(_mapping)
+
+    def popitem(self) -> Tuple[Hashable, Any]:
+        for k in self.keys():
+            if k in self.protected_keys:
+                continue
+
+            return k, self.pop(k)
```
vhirtham, line 609:
```suggestion
        for k in self.keys():
            if k not in self.protected_keys:
                return k, self.pop(k)
```

vhirtham (approved):

weldx/asdf/file.py
marscher: Not if you suggest to make a bet :) What have you figured?

weldx/asdf/util.py
marscher: I thought I write it in a more readable manner (as you could quickly overlook the negation in the condition). I guess it is a matter of taste, not runtime.

vhirtham: Jep, shouldn't have any effect on runtime. So your decision :wink

weldx/asdf/file.py
vhirtham: Nothing in particular, just that in Python nothing is really protected from being manipulated. It was just a fun- comment. 😉

marscher:
these "kernel died" errors during the notebook tests seem unrelated to the latest change.

weldx/asdf/file.py
marscher: True, you can also re-define integers (try swapping one for zero is total fun).

vhirtham:
> these "kernel died" errors during the notebook tests seem unrelated to the latest change.

Seems so, just restart the actions and see what happens

marscher:
Now the number of hacks involved to get or set protected keys has been increased further (just for you @vhirtham :D)

weldx/asdf/util.py
```diff
@@ -542,6 +553,82 @@ def _get_instance_shape(
     return None
 
 
+class _ProtectedViewDict(MutableMapping):
+    def __init__(self, protected_keys, data=None):
+        super(_ProtectedViewDict, self).__init__()
+        self.__data = data
```
CagtayFabry, line 559: nice try, but I'm pretty sure as long as something is called anything like `*data` @vhirtham will always find it 😉

PR #631, opened 2021-11-04
enable pint 0.18 and fix sphinx build

CagtayFabry:
## Changes

- fix the circular import in sphinx config with pint 0.18 type hints
- allow pint version 0.18

## Checks
- [x] updated doc/

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -28,6 +28,7 @@ def _workaround_imports_typechecking():
     """Load some packages needed for type annotations."""
     import ipywidgets  # noqa
     import pandas  # noqa
+    import pint  # noqa
     import xarray  # noqa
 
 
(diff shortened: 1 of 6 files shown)


doc/conf.py
```diff
@@ -28,6 +28,7 @@ def _workaround_imports_typechecking():
     """Load some packages needed for type annotations."""
     import ipywidgets  # noqa
     import pandas  # noqa
+    import pint  # noqa
```
CagtayFabry, line 31: this was needed to fix the import error in sphinx builds @marscher @vhirtham

vhirtham (approved):

doc/conf.py
marscher: This is so confusing.. At first I added this workaround for pandas only. Eventually we have a module name collision with the stdlib or something.

CagtayFabry: it only happens in sphinx though for whatever reason...

PR #634, opened 2021-11-04
remove ipykernel dependency

CagtayFabry:
## Changes
- remove ipykernel dependency

The [previous error](https://github.com/BAMWelDX/weldx/runs/3555541994?check_suite_focus=true) does not seem to pop up again

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -3,13 +3,9 @@ on:
   push:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    paths:
-      - 'weldx/**'
-      - 'tutorials/**'
-      - '.github/workflows/pytest.yml'
-  # Run every Monday at 6am UTC
   schedule:
     - cron: '0 6 * * 1'
+  workflow_dispatch:
 
 jobs:
   event_file:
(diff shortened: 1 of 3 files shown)


marscher (approved):

CagtayFabry:
I just realized that pytest didn't even run here...

.github/workflows/pytest.yml
```diff
@@ -3,13 +3,9 @@ on:
   push:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    paths:
-      - 'weldx/**'
-      - 'tutorials/**'
-      - '.github/workflows/pytest.yml'
-  # Run every Monday at 6am UTC
```
CagtayFabry, line 10: I have removed these conditions here since the current configuration should prevent most of the 'unnecessary' runs anyway and it doesn't pick up project config related changes otherwise @marscher @vhirtham

vhirtham: Okay, just need to remember it in case something unforeseen happens in the future 😉

vhirtham (approved):

.github/workflows/pytest.yml
CagtayFabry: it should just run the tests more often/consistently now (which I guess you welcome anyway 🎂 😃 )

vhirtham: true... 😏

PR #635, opened 2021-11-04
Check time coords

CagtayFabry:
## Changes

- check for `time` in xarray coordinates instead of dimensions in some cases

In cases where we access the `time` coordinate value I think it makes sense to check for time in the coordinates instead of the array dimensions. If time is a dimension without coordinates we would only get integer values which would be error prone.

In contrast if we actually want to check if an object is *time_dependent* (changing in time) we should look for time in the dimensions

However so far it is a bit inconsistent with single timestamps outside of the definition range:
```python
from weldx import LocalCoordinateSystem as LCS

lcs = LCS(coordinates=[[0,0,0],[1,0,0]],time=["0s","3s"])
lcs.interp_time("2s")
#> <LocalCoordinateSystem>
#> Dimensions:      (c: 3, v: 3)
#> Coordinates:
#>   * c            (c) <U1 'x' 'y' 'z'
#>   * v            (v) int32 0 1 2
#>     time         timedelta64[ns] 00:00:02
#> Data variables:
#>     orientation  (v, c) float64 1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0
#>     coordinates  (c) float64 0.6667 0.0 0.0

lcs.interp_time("4s")
#> <LocalCoordinateSystem>
#> Dimensions:      (c: 3, v: 3)
#> Coordinates:
#>   * c            (c) <U1 'x' 'y' 'z'
#>   * v            (v) int32 0 1 2
#> Data variables:
#>     orientation  (v, c) float64 1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0
#>     coordinates  (c) float64 1.0 0.0 0.0
```

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests

### DIFF
--- a/weldx/geometry.py
+++ b/weldx/geometry.py
@@ -2538,6 +2538,9 @@ class SpatialData:
                 add_dims=["n"],
             )
 
+        # make sure we have correct dimension order
+        self.coordinates = self.coordinates.transpose(..., "n", "c")
+
         if self.triangles is not None:
             if not isinstance(self.triangles, np.ndarray):
                 self.triangles = np.array(self.triangles, dtype="uint")
@@ -2801,7 +2804,7 @@ class SpatialData:
     @property
     def is_time_dependent(self) -> bool:
         """Return `True` if the coordinates are time dependent."""
-        return "time" in self.coordinates.coords
+        return "time" in self.coordinates.dims
 
     @property
     def additional_dims(self) -> List[str]:
--- a/weldx/geometry.py
+++ b/weldx/geometry.py
@@ -2801,7 +2801,7 @@ class SpatialData:
     @property
     def is_time_dependent(self) -> bool:
         """Return `True` if the coordinates are time dependent."""
-        return "time" in self.coordinates.dims
+        return "time" in self.coordinates.coords
 
     @property
     def additional_dims(self) -> List[str]:
--- a/weldx/tests/transformations/test_local_cs.py
+++ b/weldx/tests/transformations/test_local_cs.py
@@ -498,11 +498,7 @@ def test_issue_289_interp_outside_time_range(
     exp_coords = [0, 0, 0] if time_dep_coords and all_less else [1, 1, 1]
 
     assert lcs_interp.is_time_dependent is False
-    if time_dep_coords or time_dep_orient:
-        if all_less:
-            assert lcs_interp.time.equals(Time(time[0]))
-        else:
-            assert lcs_interp.time.equals(Time(time[-1]))
+    assert lcs_interp.time is None
     assert lcs_interp.coordinates.values.shape == (3,)
     assert lcs_interp.orientation.values.shape == (3, 3)
     assert np.all(lcs_interp.coordinates.data == exp_coords)
--- a/weldx/tests/transformations/test_local_cs.py
+++ b/weldx/tests/transformations/test_local_cs.py
@@ -497,7 +497,12 @@ def test_issue_289_interp_outside_time_range(
     exp_orient = WXRotation.from_euler("x", exp_angle, degrees=True).as_matrix()
     exp_coords = [0, 0, 0] if time_dep_coords and all_less else [1, 1, 1]
 
-    assert lcs_interp.time is None
+    assert lcs_interp.is_time_dependent is False
+    if time_dep_coords or time_dep_orient:
+        if all_less:
+            assert lcs_interp.time.equals(Time(time[0]))
+        else:
+            assert lcs_interp.time.equals(Time(time[-1]))
     assert lcs_interp.coordinates.values.shape == (3,)
     assert lcs_interp.orientation.values.shape == (3, 3)
     assert np.all(lcs_interp.coordinates.data == exp_coords)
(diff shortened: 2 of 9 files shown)


weldx/tests/transformations/test_local_cs.py
```diff
@@ -518,7 +519,8 @@ def test_interp_time_discrete_single_time():
     exp_orient = WXRotation.from_euler("x", 90, degrees=True).as_matrix()
 
     lcs_interp = lcs.interp_time("2s")
-    assert lcs_interp.time is None
+    assert lcs_interp.is_time_dependent is False
+    assert lcs_interp.time.equals(Time("2s"))
```
CagtayFabry, line 523: this is one case where the updated behavior differs slightly

`lcs_interp.is_time_dependent` returns `False` but we can still access the (singular) timestamp from the LCS

weldx/time.py
```diff
@@ -668,7 +668,10 @@ def _convert_xarray(
         if "time" in time.coords:
             time = time.time
         time_ref = time.weldx.time_ref
-        time_index = pd.Index(time.values)
+        if time.shape:
+            time_index = pd.Index(time.values)
+        else:
+            time_index = pd.Index([time.values])
```
CagtayFabry, line 674: singular values weren't really handled all that well before

weldx/transformations/local_cs.py
```diff
@@ -548,7 +549,7 @@ def is_time_dependent(self) -> bool:
             `True` if the coordinate system is time dependent, `False` otherwise.
 
         """
-        return self.time is not None or self._coord_ts is not None
+        return self._coord_ts is not None or ("time" in self._dataset.dims)
```
CagtayFabry, line 552: I think this is the most important change, a LCS can be time independent but still return a `.time` value

weldx/transformations/local_cs.py
```diff
@@ -686,9 +687,9 @@ def _interp_time_coordinates(self, time: Time) -> xr.DataArray:
         if "time" not in self.coordinates.dims:  # don't interpolate static
             return self.coordinates
         if time.max() <= self.time.min():  # only use edge timestamp
-            return self.coordinates[0]
+            return self.coordinates.isel(time=0).data
```
CagtayFabry, line 690: we could also skip the `.data` here but in that case we would have to decide which "edge" timestamp we want to apply, however this is now more consistent for coordinates and orientations

vhirtham (approved):

weldx/geometry.py
```diff
@@ -2538,6 +2538,9 @@ def __post_init__(self, time):
                 add_dims=["n"],
             )
 
+        # make sure we have correct dimension order
+        self.coordinates = self.coordinates.transpose(..., "n", "c")
+
```
CagtayFabry, line 2543: this was causing the troubles with scan data transformation as an incorrectly transposed xarray **(p,c,n)** was passed to `SpatialData` @vhirtham

CagtayFabry:
added some minor changes but I think this is good now @vhirtham

weldx/geometry.py
vhirtham: Nice finding. I recently also passed a "misformed" xarray to ´SpatialData´ that gave me some headaches. The time coordinate was a datetime instead of a timedelta with an optional reference time. The error message was a bit cryptic and only occurred if time-dependent transformations were performed. The question is if we want to catch this too. While we might remember the format requirements, our new 2 users from the lab might get totally confused by such errors. Maybe we can introduce an ´xr_format_data_array´ function that orders dimensions, correct wrong time inputs and possibly other stuff.

vhirtham (approved):

weldx/geometry.py
CagtayFabry: can you reproduce this bug?

the time values can be formatted here by using the `.weldx.time_ref_restore` xarray accessor, however all operations that rely on these time formats (e.g. interpolation) should also make sure to handle these internally so it would be good to know where this happens.

vhirtham: Ideally, one can grasp the expected format from the function inputs

vhirtham:
~~~ python
import xarray as xr

from weldx import CoordinateSystemManager, SpatialData, Time, WXRotation

rot = WXRotation.from_euler("z", [0, 45, 90], True).as_matrix()

csm.create_cs("workpiece", "root", coordinates=[0, 1, 2])
csm.create_cs(
    "TCP",
    "root",
    orientation=rot,
    coordinates=[[0, 0, 0], [1, 0, 0], [2, 0, 0]],
    time=Time(["2000", "2001", "2002"]),
)


da = xr.DataArray(
    [[[0, -1, 1], [0, 1, 1]], [[0, -1, 2], [0, 1, 2]], [[0, -1, -1], [0, 1, -1]]],
    dims=["time", "n", "c"],
    coords={"time": Time(["2000", "2001", "2002"]).as_pandas_index()},
)

sd = SpatialData(da)

csm.assign_data(sd, "data", "TCP", "workpiece")
~~~

This was the script I used to understand the error. I didn't create an issue cause I thought it was my own stupidity to violate the expected xarray format. However, it is probably better if those input violations are either handled or give an instant and clear response to the user as you said. The annoying thing here is: If you remover the instant transformation to the workpiece cs, the error happens later (for example during a `.plot` call or wherever the first transformation occurs) and getting the connection isn't that easy anymore.

PR #636, opened 2021-11-05
use weldx-widgets k3d implementation if available

CagtayFabry:
## Changes
- optionally use code from `weldx-widgets` if installed

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/visualization/__init__.py
+++ b/weldx/visualization/__init__.py
@@ -1,7 +1,5 @@
 """Visualization of spatial data, coordinate systems (cs), and cs manager."""
-import importlib.util
-
-if importlib.util.find_spec("weldx_widgets"):
+try:
     from warnings import warn
 
     from weldx_widgets.visualization.csm_k3d import (
@@ -10,7 +8,7 @@ if importlib.util.find_spec("weldx_widgets"):
     )
 
     warn("Using experimental 'weldx-widgets' implementation.", UserWarning)
-else:
+except ImportError:
     from .k3d_impl import CoordinateSystemManagerVisualizerK3D, SpatialDataVisualizer
 
 from .matplotlib_impl import (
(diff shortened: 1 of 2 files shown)


weldx/visualization/__init__.py
```diff
@@ -1,5 +1,18 @@
...
+if importlib.util.find_spec("weldx_widgets"):
+    from warnings import warn
+
+    from weldx_widgets.visualization.csm_k3d import (
+        CoordinateSystemManagerVisualizerK3D,
+        SpatialDataVisualizer,
+    )
+
+    warn("Using experimental 'weldx-widgets' implementation.", UserWarning)
+else:
```
marscher, line 13:
```suggestion
except ImportError:
```

weldx/visualization/__init__.py
```diff
@@ -1,5 +1,18 @@
 """Visualization of spatial data, coordinate systems (cs), and cs manager."""
-from .k3d_impl import CoordinateSystemManagerVisualizerK3D, SpatialDataVisualizer
+import importlib.util
+
+if importlib.util.find_spec("weldx_widgets"):
```
marscher, line 4:
```suggestion
try:
```

marscher:
I think we should also start to remove the matplotlib and k3d dependencies and add import-guards in the respective visualization modules. These guards should show an error message indicating that these dependencies have moved to weldx-widgets, or can be installed manually.

CagtayFabry:
> I think we should also start to remove the matplotlib and k3d dependencies and add import-guards in the respective visualization modules. These guards should show an error message indicating that these dependencies have moved to weldx-widgets, or can be installed manually.

eventually yes, let's do this a little slower with the above approach but keep the default weldx behavior (and dependencies) stable for now

marscher:
The approach I suggested would result in a pre-tested release with the conditions we would like to achieve. Otherwise I wonder how the road map for this change would look like? At some point we will do it any ways, right?

marscher (approved):

PR #639, opened 2021-11-09
[pre-commit] added clean notebook output hook

marscher:
## Changes

Added clean notebook output hook. Tested locally, works like a charm.

There is an additional argument to remove kernel metadata (e.g. the constant switching of the Python version, depending on the devs local version). But this also removed the name of the env "weldx". So I'm not sure if this would break something. That's why I just disabled it.

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -34,11 +34,3 @@ jobs:
       - name: run mypy
         run: |
           mypy --install-types --non-interactive .
-
-  check-notebooks:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-      - name: Check notebooks
-        uses: ResearchSoftwareActions/EnsureCleanNotebooksAction@1.1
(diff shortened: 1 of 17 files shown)


CagtayFabry:
nice find, this would replace our [check-notebooks job](https://github.com/BAMWelDX/weldx/blob/8d8601120d74aab1d11b4145c81c47aa0c7eeee3/.github/workflows/static_analysis.yml#L38)

we should certainly try the side effects of the kernel info removal.
For the users this would probably prompt a kernel selection on notebook load (which I think is fine), in our pytest and sphinx runs we might have to set the kernel explicitly (which should be doable, for sphinx we already do/did that)

marscher:
I guess this would simply our CI pipeline a bit, since we need to set the correct environment (or just override the existing kernel metadata). I can open up a follow up PR with removed kernel metadata to see what happens.

CagtayFabry (approved):

CagtayFabry (approved):
nice, thanks for the cleanup

looks like the python version doesn't get cleaned after all but still nice to have a good way to format

PR #626, opened 2021-11-03
New tutorials dl file

marscher:
## Changes

If "single_pass_weld.weldx" does not exist in tutorials_dir, it will be downloaded from IW repo.
The hash is being compared to the one observed from tagged version "weldx_0.5.0" of this repo.

## Checks

- [x] updated doc/

PR #640, opened 2021-11-09
New tutorials dl file within notebooks

marscher:
## Changes

we also need to invoke the dl func within the notebooks (hidden cell on top) to make sure the file is there. The previous PR #626 just downloaded it during rendering the documentation.


vhirtham (approved):
No clue what's going on here with GitHub. It lists all modifications that came with merging master even though I recently merged master into the target branch. So there is a good chance I missed something out, but I guess we will see. Just merge. We get a second review chance when merging into master.

PR #643, opened 2021-11-10
[wxfile] impl inline array threshold

### LINKED ISSUE #601: set default inline threshold
I think it would make sense to set a (sensible) default threshold to inline array values for most of our use cases

I think 10 should be a good value because it will inline a single (static) 3x rotation matrix and short list of vectors e.g. movements between two points
```yaml
coordinates: !weldx!core/variable-0.1.0
    name: coordinates
    dimensions: [time, c]
    dtype: <f8
    data: !core/ndarray-1.0.0
      data:
      - [20.0, 0.0, 3.0]
      - [330.0, 0.0, 3.0]
      datatype: float64
      shape: [2, 3]
```

marscher:
## Changes

ASDF 2.8 introduced the config context and deprecated this as a plain argument. However in the wrapper I think it'd be a sane solution to have it as an (optional) parameter again. It is delegated to the temporary config context.

## Related Issues

Closes #601 

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,12 @@
 added
 =====
 
+- `WeldxFile` handles an ``array_inline_threshold`` parameter to indicate if short arrays
+  will be serialized as strings, or as binary block. Note that this not affect arrays,
+  which are being shared across several objects in the same file.
+  `[#643] <https://github.com/BAMWelDX/weldx/pull/643>`__
+
+
 removed
 =======
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ added
 =====
 
 - `WeldxFile` handles an ``array_inline_threshold`` parameter to indicate if short arrays
-  will be serialized as strings, or as binary block. Note that this not affect arrays,
+  will be serialized as strings, or as binary block. Note that this does not affect arrays,
   which are being shared across several objects in the same file.
   `[#643] <https://github.com/BAMWelDX/weldx/pull/643>`__
 
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -3,7 +3,6 @@ import copy
 import io
 import pathlib
 import warnings
-
 from collections.abc import MutableMapping, ValuesView
 from contextlib import contextmanager
 from io import BytesIO, IOBase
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -9,6 +9,12 @@
 added
 =====
 
+- `WeldxFile` handles an ``array_inline_threshold`` parameter to indicate if short arrays
+  will be serialized as strings, or as binary block. Note that this not affect arrays,
```
CagtayFabry, line 13:
```suggestion
  will be serialized as strings, or as binary block. Note that this does not affect arrays,
```

weldx/asdf/file.py
```diff
@@ -300,6 +312,21 @@ def __init__(
...
+            "array_inline_threshold" not in kwargs
+            or kwargs["array_inline_threshold"] is None
+        ):
+            kwargs["array_inline_threshold"] = self._array_inline_threshold
+
+        with config_context() as config:
+            for k, v in kwargs.items():
+                setattr(config, k, v)
+            yield
+
```
CagtayFabry, line 329: If I understand this correctly all (internal) `WeldxFile` operations should now use this custom context? I think this is a very good solution and more in line with the new asdf API 👍

CagtayFabry (approved):
LGTM 🚀 
should be a nice addition to the `0.5.2` release

weldx/asdf/file.py
```diff
@@ -139,6 +143,10 @@ class WeldxFile(_ProtectedViewDict):
         When `False`, when reading files, attempt to memory map (memmap) underlying data
         arrays when possible. This avoids blowing the memory when working with very
         large datasets.
+    array_inline_threshold :
+        arrays below this threshold will be serialized as string, if larger as binary
+        block. Note that this not affect arrays, which are being shared across several
+        objects in the same file.
```
vhirtham, line 149:
```suggestion
        block. Note that this does not affect arrays, which are being shared across several
        objects in the same file.
```
Same as above (already mentioned by @CagtayFabry )

vhirtham (approved):

weldx/asdf/file.py
marscher: Yes, currently it is only used where really needed. That includes write operations. But I guess later on, we will wrap everything in this context.

PR #648, opened 2021-11-16
csm.__eq__ now generically checks all node and edge attributes

### LINKED ISSUE #395: CSM.__eq__ does not consider attached data
This is a really low priority issue since CSM comparison is mainly used in tests, but as the title says, attached data is not considered during the comparison.

marscher:
## Changes
csm.__eq__ now generically checks all node and edge attributes. This includes transformations and data.

## Related Issues

Closes #395 

## Checks

- [x] updated tests


weldx/transformations/cs_manager.py
```diff
@@ -175,17 +174,19 @@ def __eq__(self, other: Any):
...
                 return False
 
-        # check coordinate systems
-        for edge in graph_0.edges:
-            lcs_0 = self.graph.edges[(edge[0], edge[1])]["transformation"]
-            lcs_1 = other.graph.edges[(edge[0], edge[1])]["transformation"]
-            if lcs_0 != lcs_1:
+            e0 = graph_0.edges[edge]
+            e1 = graph_1.edges[edge]
+            if not util.compare_nested(e0, e1):
```
CagtayFabry, line 189: cool that this works here

CagtayFabry (approved):

weldx/tests/transformations/test_cs_manager.py
```diff
@@ -737,6 +737,18 @@ def test_comparison_wrong_type():
     assert (csm != 4) is True
 
 
+def test_comparison_data():
+    csm1 = tf.CoordinateSystemManager("root", "csm")
+    csm2 = tf.CoordinateSystemManager("root", "csm")
+    data = np.arange(9).reshape((3, 3))
```
vhirtham, line 743:
```suggestion
    data = np.arange(12).reshape((3, 4))
```
Not really relevant here (so no need to apply the suggestion), but symmetrical shapes in tests, tend to be dangerous. They can accidentally pass tests they shouldn't pass.

vhirtham (approved):
LGTM 👍

weldx/tests/transformations/test_cs_manager.py
marscher: Now we receive this error:  ValueError: conflicting sizes for dimension 'c': length 4 on the data but length 3 on coordinate 'c'

Was this intended @vhirtham?

vhirtham: Not really intended. Just forgot that the last dimension needs to be of length 3 since it gets converted to a `SpatialData` IIRC. So changing the shape to `(4, 3)` should fix this

PR #673, opened 2021-12-18
set meshio<5.2 for sphinx builds

### LINKED ISSUE #672: meshio import errors on sphinx builds
This error seems to have popped up very recently, maybe related to meshio 5.2 or 5.1.2 https://pypi.org/project/meshio/#history

https://github.com/BAMWelDX/weldx/runs/4570280776?check_suite_focus=true

CagtayFabry:
## Changes
- set `meshio<5.2` in the sphinx environment to avoid import error
https://github.com/BAMWelDX/weldx/runs/4570280776?check_suite_focus=true

## Related Issues

Closes #672

I am not sure why this only pops up in the sphinx build, do we have a test using meshio? @vhirtham

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -56,7 +56,7 @@ jobs:
         # There is a bug in micromamba, which cd's out of the working-directory during the pip install of weldx.
         # So the initial env creation lacks weldx, so we install it here.
         shell: bash -l {0}
-        run: pwd && pip install -e . && pip install git+https://github.com/CagtayFabry/sphinx-asdf.git@sphinx-weldx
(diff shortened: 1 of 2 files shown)


vhirtham:
> I am not sure why this only pops up in the sphinx build, do we have a test using meshio? @vhirtham

Not that I am aware of

marscher:
I have encountered several of these circular import errors during loading Sphinx conf.py. That's why there is this workaround function to load modules like xarray and pandas. I cannot make anything of that strange behaviour as you noticed it only happens in Sphinx builds. So its probably related to the "changed" import behaviour there.

PR #678, opened 2022-01-14
run PEP585 upgrade

CagtayFabry:
## Changes
- upgrade type hints to build in types (using [PEP585 Upgrade](https://github.com/snok/pep585-upgrade))

I have disabled the pre-commit hook since it breaks some of our type definitions (it might work but I didn't check further)

Using the pep585 type hints should be the new default with `from __future__ import annotations`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,7 +59,7 @@ repos:
         args:
           - --convention=numpy
   - repo: https://github.com/nbQA-dev/nbQA
-    rev: 1.2.2
+    rev: 1.2.3
     hooks:
       - id: nbqa-black
       - id: nbqa-isort
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,11 @@ repos:
       - id: check-yaml
         exclude: devtools/conda.recipe/meta.yaml  # doesn't play nice with jinja
   #    - id: no-commit-to-branch  # only makes sense for local pre-commit hooks
+#  - repo: https://github.com/sondrelg/pep585-upgrade
+#    rev: v1.0.1
+#    hooks:
+#    - id: upgrade-type-hints
+#      args: [ '--futures=true' ]
   - repo: https://github.com/psf/black
     rev: 21.12b0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,11 +19,11 @@ repos:
       - id: check-yaml
         exclude: devtools/conda.recipe/meta.yaml  # doesn't play nice with jinja
   #    - id: no-commit-to-branch  # only makes sense for local pre-commit hooks
-#  - repo: https://github.com/sondrelg/pep585-upgrade
-#    rev: v1.0.1
-#    hooks:
-#    - id: upgrade-type-hints
-#      args: [ '--futures=true' ]
+  - repo: https://github.com/sondrelg/pep585-upgrade
+    rev: v1.0.1
+    hooks:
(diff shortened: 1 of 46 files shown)


marscher:
What is the purpose of using more specific types, e.g. tuple vs. tpying.Tuple, in the annotation? Eventually, we do not want to restrict it to be specific to that particular type, but a broader common interface. Of course this can be changed if desired, I just wonder what this has to do with pep585.

vhirtham:
So most of the time it is just replacing lower-case with upper-case in the type hints and a different import source. Well, to me it is not really important which annotation we use, as long as everything works and it is consistent. 

However:
 
> I have disabled the pre-commit hook since it breaks some of our type definitions (it might work but I didn't check further)

So can we fix this?

marscher:
Now I've learned that the type duplication of collection types in the typing module is now deprecated. Thanks for fixing this.
Why is the pre-commit hook disabled? Are there more spots in the code, where we still use the typing collections?

marscher (approved):

CagtayFabry:
> However:
> 
> > I have disabled the pre-commit hook since it breaks some of our type definitions (it might work but I didn't check further)
> 
> So can we fix this?

I didn't want to spend the time digging into it, feel free to run the hook again and see the changes/errors (they are only a few)


> Why is the pre-commit hook disabled? Are there more spots in the code, where we still use the typing collections?

See above, in addition the hook does 'mess' with the imports (adding new ones, removing some) but doesn't respect `isort` or simliar. So sometimes pre-commit has to run twice (also the reason why I put early in the hook list)

This was mostly intended as a one time upgrade to switch to pep585 annotations, if we simply stick to that from now on we probably don't need to have the hook activated (or we can run it from time to time manually)

marscher:
> See above, in addition the hook does 'mess' with the imports (adding new ones, removing some) but doesn't respect `isort` or simliar. So sometimes pre-commit has to run twice (also the reason why I put early in the hook list)

The readme of the hook does mention that and recommends to run this hook in a "tandem" with isort. So I guess we need to figure how to do that.

quote: 
> Note: even though we remove and add imports reasonably well, I would recommend running this in tandem with hooks like isort to aggregate and sort your imports, and flake8 to discover any unused imports neither were able to remove.

CagtayFabry:
> 

exactly, if you simply run the hook before isort it will 'heal' itself (but the first commit attempt will still fail)
right now the hook doesn't finish cleanly so if you want to enable it someone would quickly have to look into the fixes necessary (or disable the hook for specific files) @vhirtham @marscher

marscher:
RTD has now the following warning: weldx/geometry.py:docstring of weldx.geometry.SpatialData.time:: WARNING: more than one target found for cross-reference 'Time': weldx.Time, weldx.time.Time

This is surprising, since I only changed something in weldx/asdf/util.py

Eventually the new Sphinx version is being used now.

marscher:
Thanks. The default branch triggers the same warning.  So we can either merge now and fix the warning in a separate PR or resolve it here. From my experience you prefer to keep things separated, right?

vhirtham:
> Thanks. The default branch triggers the same warning. So we can either merge now and fix the warning in a separate PR or resolve it here. From my experience you prefer to keep things separated, right?

It's the same on the tutorial PR. I will look into the problem next week.

vhirtham:
> > Thanks. The default branch triggers the same warning. So we can either merge now and fix the warning in a separate PR or resolve it here. From my experience you prefer to keep things separated, right?
> 
> It's the same on the tutorial PR. I will look into the problem next week.

See #684

Issue #682, opened 2022-01-19
CSM/LCS unit support


vhirtham:
The CSM and LCS are still missing unit support (mainly for coordinates). As discussed with @CagtayFabry we are planning to change that in multiple steps/PRs.

First step will be to add `pint` decorators to all functions that should support units. They will be set to `strict=False` so that they still work without units. Functions that support multiple Input types (`xarray`) will provide a similar mechanism without the decorator.

Second step will be to adjust all tests and notebooks of the weldx repository so that we can switch to "strict"-mode without any errors. However, at this point, non-quantity inputs are still allowed to keep support for other projects/repos based on weldx.

The last step is switching to strict mode and assure via tests that you can not provide plain numbers to a CSM or LCS and have to use Quantities.

PR #683, opened 2022-01-24
Check units of coordinates

vhirtham:
## Changes

Adds unit support for the LCS and CSM coordinates without requiring units.

## Related Issues

#682 (do not close)

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/


weldx/transformations/local_cs.py
```diff
@@ -292,8 +296,20 @@ def _build_coordinates(
         if not isinstance(coordinates, xr.DataArray):
             if not isinstance(coordinates, (np.ndarray, pint.Quantity)):
                 coordinates = np.array(coordinates)
+
             coordinates = ut.xr_3d_vector(coordinates, time)
 
+        if isinstance(coordinates.data, pint.Quantity):
+            # The first branch is a workaround until units are mandatory
+            if coordinates.data.u == pint.Unit(""):
+                coordinates.data = coordinates.data.m
```
CagtayFabry, line 305: This looks dangerous to me, I guess the assumption is that if a 'dimensionless' quantity is passed we interpret it as 'default unit' ?
I would prefer to be strict here and force all quantities to be `[length]` so that at least the quantity behavior is already consistent

vhirtham: The problem is that our "unitless" code breaks at multiple places when we remove this line. This is just a workaround to enable units as input without enforcing them. For example, the "TimeSeries" returns unitless quantities in several tests and examples/notebooks. Fixing those problems is part of the transitions second step.

CagtayFabry: ah I see, thank you
Is it only tests/notebooks or also inside the package?

weldx/transformations/local_cs.py
```diff
@@ -310,6 +310,13 @@ def _build_coordinates(
                     extra_msg="\nThe coordinates require units representing a length.",
                 )
 
+        if not isinstance(coordinates.data, pint.Quantity):
+            warnings.warn(
+                "Coordinates without units are deprecated and won't be supported in "
+                "the future",
+                DeprecationWarning,
```
marscher, line 317: please add a stacklevel argument here, to inform the callee of this method. I guess 2 in this case, since it is a private method.

vhirtham:
This is ready now. Remember that it is just an intermediate step with the goal to add unit support to LCS and CSM without breaking code that does not provide units. So there are some workarounds involved that will disappear later.

What doesn't work anymore are mixed cases. So if some of the LCS in a CSM have units while some others don't, you will get an error. This was the case in some of the notebooks so I had to adjust them.

tutorials/welding_example_02_weaving.ipynb
```diff
@@ -113,6 +113,17 @@
     ")"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "test = geometry.spatial_data(\n",
```
marscher, line 122: This `test` variable is not meant to be here any more, right?

weldx/geometry.py
```diff
@@ -1439,6 +1439,9 @@ def local_coordinate_system(
         relative_position = np.clip(relative_position, 0, 1)
 
         coordinates = np.array([1, 0, 0]) * relative_position * self._length
+        if isinstance(coordinates, pint.Quantity):
```
marscher, line 1442: So the basis vectors of the LCS have no unit? This confuses me,  I thought we are now demand units, and trigger a warning if not units are provided.

tutorials/experiment_design_01.ipynb
```diff
@@ -201,7 +202,7 @@
     "t_start = pd.Timedelta(\"0s\")\n",
     "t_end = pd.Timedelta(str(t_weld.to_base_units()))\n",
     "\n",
-    "coords = [tcp_start_point.magnitude, tcp_end_point.magnitude]\n",
+    "coords = np.vstack(tcp_start_point, tcp_end_point)\n",
```
CagtayFabry, line 205: using `np.vstack` or `np.append`, `np.hstack` should be our preferred option since it is safe against conversion errors

CagtayFabry (commented):
There seem to be quite a few `UnitStrippedWarning` in some of the tutorials now, e.g. https://weldx--683.org.readthedocs.build/en/683/tutorials/welding_example_02_weaving.html#add-a-sine-wave-to-the-TCP-movement

I guess it is the same line causing each

weldx/geometry.py
CagtayFabry: This only applies to the LCS generated by a `LinearHorizontalTraceSegment` so more of an edge case that (I guess) is excluded for now because of tests/compatibility

vhirtham: Due to the transition, the whole thing is currently a mess. As @CagtayFabry mentioned, this only applies to this segment class. The LCS currently supports both formats but the geometry classes are expect LCS without coordinates. The point of this PR was just to make LCS coordinate units possible without breaking anything. So as I mentioned earlier, there are some strange and ugly workarounds like this one. These will disappear in the next PR when I fix the whole package to be compatible with strict units.

marscher (approved):

weldx/visualization/k3d_impl.py
```diff
@@ -340,7 +340,7 @@ def __init__(
 
         self._reference_system = reference_system
 
-        self._label_pos = data.coordinates.mean(dim=data.additional_dims).values
+        self._label_pos = data.coordinates.mean(dim=data.additional_dims).data.m
```
vhirtham, line 343:
```suggestion
        self._label_pos = data.coordinates.mean(dim=data.additional_dims).values
```

vhirtham:
Couldn't stick to the original plan (not touching k3d) due to the unit stripped warnings now being errors. However, got everything working now. Additionally, I moved the default units to `weldx.constants`.

tutorials/welding_example_02_weaving.ipynb
```diff
@@ -75,10 +75,10 @@
    "source": [
     "groove = get_groove(\n",
     "    groove_type=\"VGroove\",\n",
-    "    workpiece_thickness=Q_(0.5, \"cm\"),\n",
-    "    groove_angle=Q_(50, \"deg\"),\n",
-    "    root_face=Q_(1, \"mm\"),\n",
-    "    root_gap=Q_(1, \"mm\"),\n",
+    "    workpiece_thickness=\"0.5 cm\",\n",
```
marscher, line 78: Would it also have been possible to pass a tuple[Union[float, int], str], e.g. (1, "mm")? I would prefer such an option, if these quantities are actually the result of a computation.

vhirtham: not sure at this specific point since I didn't touch the internals of the function. Just fixed an issue that popped up due to the function using the LCS. Anyways, all newer functions that support units should usually provide several format options for quantities. I think a tuple is also possible, but I have to check it.

PR #689, opened 2022-01-26
fix time behavior in tests and filter warnings for pandas 1.4

CagtayFabry:
## Changes
- Fix failing tests due to new pandas version ->seems like pandas added own `__add__` functions so that we couldn't use + with a pandas Type on the left-hand side and our Time class on the right-hand side (thanks @vhirtham )
- filter warnings caused by xarray introduced with pandas 1.4

## Checks

- [x] updated tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -113,6 +113,7 @@ asdf_schema_root = weldx/schemas/weldx.bam.de/weldx
 norecursedirs = doc
 filterwarnings =
     ignore::DeprecationWarning:traittypes.*:
+    ignore:Passing method to :FutureWarning:xarray.*:
 
 [isort]
 profile = black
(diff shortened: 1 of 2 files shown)


CagtayFabry:
This should get rid of a few thousand warnings in tests hopefully 🚀😄
(and clean up your notebooks @marscher 😉)

marscher:
> 

Eventually it makes sense to pin to a minor of pandas, which has been tested and only upgrade it, if everything works as expected.

CagtayFabry:
something like `pandas>=1.0,<=1.3` ?

I think that would be ok if we have a good workflow release patch versions if new pandas releases work as expected (while still giving some safety)

marscher:
Yes.

PyPI prohibits uploading patched sources, so this always renders a new micro release just for updating the dependencies. Conda does circumvent this by increasing build numbers. I guess there is no way to avoid this.

Eventually it is too much overhead.

Issue #693, opened 2022-01-31
distutils.version deprecated in favor of packaging.version


marscher:
We currently use the LooseVersion class of distutils to compare tag version in asdf/util.py
This triggers a DeprecationWarning to replace this with packaging.version. However it seems that the Loose or now LegacyVersion is also going to be removed.

PR #694, opened 2022-01-31
Rename wxfile asdf header

### LINKED ISSUE #688: refactor `WeldxFile.show_asdf_header()` into `WeldxFile.header()`
as mentioned in #580 we should deprecate the old `show_asdf_header()` for `header()`

marscher:
## Changes

* Refactored/deprecated WeldxFile.show_asdf_header to header.
* Use packaging.version instead of distutils.version (deprecated).

## Related Issues

Closes #688, #693

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/

### DIFF
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -170,7 +170,7 @@ class WeldxFile(_ProtectedViewDict):
 
     We can have a look at the serialized data by looking at the asdf header
 
-    >>> wx2.show_asdf_header()  # doctest: +ELLIPSIS,+NORMALIZE_WHITESPACE
+    >>> wx2.header()  # doctest: +ELLIPSIS,+NORMALIZE_WHITESPACE
     #ASDF 1.0.0
     #ASDF_STANDARD 1.5.0
     %YAML 1.1
@@ -832,7 +832,7 @@ class WeldxFile(_ProtectedViewDict):
     def data(self):
         return self._data
 
-    def show_asdf_header(
+    def header(
         self,
         use_widgets: bool = None,
         path: tuple = None,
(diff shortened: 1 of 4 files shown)


weldx/asdf/util.py
```diff
@@ -509,7 +509,11 @@ def get_highest_tag_version(
     if not tags:  # no match found
         return None
 
-    tags.sort(key=LooseVersion)
+    def _get_version(tag_name: str):
+        # we assume, that the version of the tag is separated by a right-most '-' char.
+        return Version(tag_name[tag_name.rfind("-") + 1 :])
+
+    tags.sort(key=_get_version)
```
CagtayFabry, line 516:
```suggestion
    tags.sort(key=packaging.version.parse)
```
This should work but seems to use `LegacyVersion` in the back so if we want to avoid that your solution could also be `tags.sort(key=lambda x: Version(x.rpartition("-")[-1]))`

marscher: It does not use LegacyVersion, the parse function does, if the version format is not pep440 compatible. If we do not omit the tag uri, parse would fall back to LegacyVersion, triggering another DeprecationWarning.

weldx/asdf/file.py
```diff
@@ -858,14 +858,18 @@ def show_asdf_header(
             Should not be set.
         """
         return _HeaderVisualizer(self._asdf_handle).show(
-            use_widgets=use_widgets, path=path, _interactive=_interactive
```
CagtayFabry, line 861: Is there a specific reason you removed the `tag` argument here? @marscher 

Did I pout it here by mistake?

marscher: my bad.

marscher: and good catch, it seems this new option lacks a test.

CagtayFabry: maybe .. 👼

CagtayFabry (approved):

PR #696, opened 2022-02-04
use Python-3.10 in docs building, testing

marscher:
## Changes

Since we now have a binary wheel for NumPy for this Python version, we transition to test against it.

UPDATE:
Bottleneck still causes trouble on 3.10, but we can leave this open until this is fixed.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -52,6 +52,7 @@ jobs:
         shell: bash -l {0}
         run: micromamba activate rtd
 
+      # TODO: this should be fixed upstream, try to remove it.
       - name: install dev version again
         # There is a bug in micromamba, which cd's out of the working-directory during the pip install of weldx.
         # So the initial env creation lacks weldx, so we install it here.
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -52,13 +52,6 @@ jobs:
         shell: bash -l {0}
         run: micromamba activate rtd
 
-      # TODO: this should be fixed upstream, try to remove it.
-      - name: install dev version again
-        # There is a bug in micromamba, which cd's out of the working-directory during the pip install of weldx.
-        # So the initial env creation lacks weldx, so we install it here.
-        shell: bash -l {0}
-        run: pwd && pip install -e .
-
       - name: conda info
         shell: bash -l {0}
         run: conda info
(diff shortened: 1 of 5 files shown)


marscher:
https://stackoverflow.com/questions/38762416/how-to-specify-python-requirements-by-allowing-prereleases

marscher:
If we put the prerelease as requirement it will be installed with pip, but conda builds now fail. On Conda-forge we do not have this prerelease version, so `pip check` will fail. I guess it will be safe for us to wait until we have a proper release.

marscher:
Now bottleneck-1.3.3 is on PyPI, and soon on CF.

marscher:
Now it works as expected.
On conda-forge the newest asdf release is blocking, as it depends on asdf-standard and another external schema package, which are not yet available on CF. I just added this here, because this PR is already dealing with dependencies.

CagtayFabry:
Thank you, it looks good !
I think keeping `asdf<2.10` in there is good for now - as you mentioned there is quite some restructuring going on over there right now.

We should probably also "patch" our latest `weldx` conda-forge release to `asdf<2.10`

Should we work with `3.10` as the new default for testing etc. ?

CagtayFabry (approved):

marscher:
I'm fine with testing on 3.10 only, but I think testing on the "oldest" supported minor release would also be reasonable. That means, we omit 3.9 and keep 3.8 and 3.10.

I will create a build on CF adding the asdf requirement.

CagtayFabry:
> I'm fine with testing on 3.10 only, but I think testing on the "oldest" supported minor release would also be reasonable. That means, we omit 3.9 and keep 3.8 and 3.10.
> 
> I will create a build on CF adding the asdf requirement.

In that case let's keep everything on 3.8 by default and just add 3.10 to the testing matrix 👍

marscher:
Notebooks are currently only tested on 3.8 (I guess for the sake of faster CI). So are you saying, we do this now for 3.8 and 3.10?

I also suggest pinning Python <=3.10, so we do not run into failures in the future and unpin it, when we are ready to make the transition?

marscher:
https://github.com/conda-forge/weldx-feedstock/pull/14

CagtayFabry:
> Notebooks are currently only tested on 3.8 (I guess for the sake of faster CI). So are you saying, we do this now for 3.8 and 3.10?
> 
> I also suggest pinning Python <=3.10, so we do not run into failures in the future and unpin it, when we are ready to make the transition?

+1 for pinning `<=3.10` to spare us the trouble next time
The rest is ok from my side as is here 👍

marscher:
The workflows do not like the [ci skip] tag.

PR #697, opened 2022-02-07
split read/write schema validation

### LINKED ISSUE #690: Additional write modes for `WeldxFile`
Currently, we only have the modes `r`  and for reading and `rw` for creating a new file if it doesn't exist or modifying a file if it does exist.
I suggest that we also add `w` to only write if the file doesn't exist and `wf` to overwrite an existing file.

Reason: When developing a file schema, the `rw` mode can have some undesired side effects. For example, you write a file, and afterward, you adjust the schema to be more strict and wouldn't allow the current data to be written. The next write attempt gives you the expected schema validation error. You correct the data and try to write it again, but you get the same validation error, even if the data seems valid. Why? Because `WeldxFile` reads the old file first, that was valid before the schema was updated. If you don't realize this, you have a rough time figuring out why your correct data still causes a validation error (or at least seems to). 

So using `w` mode would raise a meaningful exception if the file already exists while `wf` (force) simply overwrites the old file without opening it.

marscher:
## Changes

WeldxFile now supports two custom_schema sub-arguments. The first argument will be used upon reading, the second upon writing. That way we can use separate schema for both operations. It turned out, that this is useful during schema development.

## Related Issues

Closes #690 

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/tests/asdf_tests/test_weldx_file.py
+++ b/weldx/tests/asdf_tests/test_weldx_file.py
@@ -339,7 +339,7 @@ class TestWeldXFile:
             WeldxFile(custom_schema="no")
 
     @staticmethod
-    def _get_schema_file(arg, mismatch: bool) -> (str, np.ndarray):
+    def _get_schema_file(arg, mismatch: bool, tmp_path) -> (str, np.ndarray):
         data = np.arange(6)
         if not arg or not isinstance(arg, (list, tuple)):
             return None, data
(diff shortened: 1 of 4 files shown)


marscher:
The sphinx warnings are due to the intersphinx mapping to Scipy is currently broken.

weldx/tests/asdf_tests/test_weldx_file.py
```diff
@@ -338,6 +338,79 @@ def test_custom_schema_real_file(tmpdir):
...
+                    fh.write(schema.format(shape=shape_s))
+                result[0] = fname
+            if b is not None:
+                fname = str(d / "b")
+                shape = [3, 2]
+                shape_s = ",".join(str(s) for s in shape)
+
+                with open(fname, "w+") as fh:
+                    fh.write(schema.format(shape=shape_s))
+                result[1] = fname
```
CagtayFabry, line 379: Thank you for adding a testcase for this !

I think there is a way with `asdf` to provide a schema without saving an actual file but I am not sure if it works with our `WeldxFile` wrapper

Here is an example:
https://github.com/asdf-format/asdf/blob/66abc5e4599db381850dddcb293a90c4a9afd999/asdf/tests/test_schema.py#L196-L204

I will look into if we can apply this somehow in our test cases, it should make a lot of things much easier

but this solution is good

CagtayFabry (approved):
I like this solution 👍

weldx/tests/asdf_tests/test_weldx_file.py
marscher: Thanks for providing the file-free version of schema validation! This looks very useful. Since WeldxFile respects outer `Config` contexts, I think it would be possible to use this pattern. I will try it out quickly.

CagtayFabry: I think what might be an issue is that we only look for schemas as files in our current implementation
https://github.com/BAMWelDX/weldx/blob/ae2dff5c2b4c9e90e95f2b4357b60b291496c157/weldx/asdf/file.py#L221-L230

what we might need is another 'fallback' into `asdf.schema.load_schema()` in the place where we 'resolve' the schema

marscher: Although it is a nice pattern, the custom_schema argument routing (e.g. tuple unpacking) is only possible within the wrapper. Also AsdfFile expects user schemas to be a file or URI: https://github.com/asdf-format/asdf/blob/66abc5e4599db381850dddcb293a90c4a9afd999/asdf/tests/test_schema.py#L899

CagtayFabry: So we cant use it here?

PR #701, opened 2022-02-09
add and run mdformat

CagtayFabry:
## Changes

add `mdformat` (and plugins) to pre-commit config
(a markdown linter at last?)


CagtayFabry:
> Is there an equivalent for restructured text as well?

none that I know of that works well for us (`rstfmt` has issues with asdf/sphinx directives)

PR #702, opened 2022-02-09
treat `pint.UnitStrippedWarning` as error in pytest runs

CagtayFabry:
## Changes

`pint.UnitStrippedWarning` will be raised as errors in our test suite

## Checks

- [x] updated tests


vhirtham (approved):
Works like a charm

PR #705, opened 2022-02-15
Download `single_pass_weld.wx` directly in the tutorials

vhirtham:
## Changes

I added a hidden cell at the top in each of the new tutorials. This cell calls the function to download the needed `single_pass_weld.wx` file. The background is that you can't run those tutorials locally without the file but it is only downloaded when in another directory if you actually build the docs. I also edit a flag to the download function to suppress text outputs.

## Checks

- [x] ~~updated CHANGELOG.rst~~
- [x] ~~updated tests~~
- [x] updated doc/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
           - mdformat-config
   # ----- Python formatting -----
   - repo: https://github.com/sondrelg/pep585-upgrade
-    rev: v1.0.1
+    rev: v1
     hooks:
     - id: upgrade-type-hints
       args: [ '--futures=true' ]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@
 added
 =====
 
+-  first draft of the ``multi_pass_weld`` schema for WelDX files [:pull:`667`]
+
 -  add `GenericSeries` as base class supporting arrays and equations [:pull:`618`]
 
 -  add experimental unit support for ``.weldx.interp_like`` accessor [:pull:`518`]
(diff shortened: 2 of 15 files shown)


marscher:
I think we could also remove the download inside the docs (conf.py), right?

CagtayFabry:
> I think we could also remove the download inside the docs (conf.py), right?

wasn't the idea to download the file once for the tests so that it is available?

CagtayFabry:
I think it is cleaner to use the `"nbsphinx": "hidden"` cell metadata so that the download code doesn't show up in the html documentation

CagtayFabry (approved):

marscher:
If the notebooks download the file, we do not need to do it explicitly (even for the tests), right? That should be the case as all notebook cells will be executed by the tests in sequential order. The only thing that could happen then is a race condition, when two notebooks will start to download the file. This should be avoided as it leads to unpredictable test failures. 
The current situation is, that we download during Sphinx building and only the first tutorial will download explicitly. 
The download code has been duplicated to do so. I'd prefer a solution without duplication and having the notebooks download it seems to be the one.

CagtayFabry:
I just wasn't sure about the parallel execution of the notebooks via pytest but that's ok for me too if it works

marscher:
not downloading in Sphinx works. The RTD error is unrelated. 
`
intersphinx inventory 'https://k3d-jupyter.org/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectionError'>: HTTPSConnectionPool(host='k3d-jupyter.org', port=443): Max retries exceeded with url: /objects.inv (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fe6e6f7c760>: Failed to establish a new connection: [Errno 110] Connection timed out'))
`

CagtayFabry (approved):

PR #710, opened 2022-02-28
ignore matplotlib docs link warnings 

CagtayFabry:
## Changes
- ignore `matplotlib` link warnings in documentation

## Checks
- [x] updated doc/


CagtayFabry:
failing weekly docs build for reference
https://github.com/BAMWelDX/weldx/actions/runs/1908741675

CagtayFabry:
oh I was hoping this would fix the broken links but I guess not 😏

CagtayFabry:
I think the problem is in the mpl docs structure, for example the `Axis` class is documented here https://matplotlib.org/stable/api/axes_api.html#the-axes-class but not like https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.html

marscher:
https://github.com/BAMWelDX/weldx/runs/5357164230?check_suite_focus=true seems like we have two notebook jobs now, downloading the file (race condition, we were talking about).

PR #712, opened 2022-03-01
Update `GenericSeries` docstrings

vhirtham:
## Changes

Update docstrings of `GenericSeries` constraint variables to help understand what their purpose is.

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -857,16 +857,16 @@ class GenericSeries:
     """
 
     _required_dimension_units: dict[str, pint.Unit] = {}
-    """A dictionary that maps required unit dimensionality to a dimension.
+    """A dictionary that maps a required unit dimensionality to a dimension.
 
     If a dimension matches one of the keys of this dictionary, its dimensionality
-    is checked against the requirement listed.
+    is checked against the listed requirement.
     """
     _required_dimension_coordinates: dict[str, list] = {}
-    """A dictionary that maps required coordinates to a dimension.
+    """A dictionary that maps required coordinate values to a dimension.
 
     If a dimension matches one of the keys of this dictionary, it is checked if it has
-    the specified coordinates. An example use-case would be a 3d-space where the
+    the specified coordinate values. An example use-case would be a 3d-space where the
     coordinates "x", "y" and "z" are required for a spatial dimension.
     """
 
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -857,10 +857,10 @@ class GenericSeries:
     """
 
     _required_dimension_units: dict[str, pint.Unit] = {}
-    """A dictionary that maps required units to a dimension.
+    """A dictionary that maps required unit dimensionality to a dimension.
 
-    If a dimension matches one of the keys of this dictionary, it is checked that the
-    units are identical to the requirement. If not, an exception is raised.
+    If a dimension matches one of the keys of this dictionary, its dimensionality
+    is checked against the requirement listed.
     """
     _required_dimension_coordinates: dict[str, list] = {}
     """A dictionary that maps required coordinates to a dimension.
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -831,11 +831,10 @@ class GenericSeries:
     as parameters.
     """
     _required_variables: list[str] = []
-    """A list of required variable names.
+    """A list of required variable names. (only expression)
 
-    The names are only relevant if the created `GenericSeries` is based on expressions.
     If one or more variables are missing in the expression, an exceptions is raised.
-    Note, that the required symbols must be variables of the expression. Using one or
+    Note that the required symbols must be variables of the expression. Using one or
     more as a parameter will also trigger an exception.
     """
 
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -840,7 +840,7 @@ class GenericSeries:
     _evaluation_preprocessor: dict[str, Callable] = {}
     """Mapping of variable names to preprocessor functions that are applied prior to evaluation.
 
-    When calling a `GenericSeries.evaluate`, the passed keyword arguments are checked against
+    When calling `GenericSeries.evaluate`, the passed keyword arguments are checked against
     the dictionaries keys. If a match is found, the corresponding preprocessor function
     is called with the variables value and returns the updated value. As an example,
     this can be used to support multiple time formats. The key might be ``t`` and the
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -838,7 +838,7 @@ class GenericSeries:
     """
 
     _evaluation_preprocessor: dict[str, Callable] = {}
-    """A mapping of variable names to preprocessor functions are applied prior to evaluation.
+    """Mapping of variable names to preprocessor functions that are applied prior to evaluation.
 
     When calling a `GenericSeries.evaluate`, the passed keyword arguments are checked against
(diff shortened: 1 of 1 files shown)


weldx/core.py
```diff
@@ -822,22 +822,64 @@ class GenericSeries:
...
     _allowed_variables: list[str] = []
-    """Allowed variable names"""
+    """A list of allowed variable names.
+
+    The names are only relevant if the created `GenericSeries` is based on expressions.
+    If the expression contains any other variable name that is not part of the list an,
+    an exception is raised. It is not required that an expression includes all these
+    variables. Additionally, the expression can contain other symbols if they are used
+    as parameters.
+    """
```
CagtayFabry, line 832:
```suggestion
    """A list of allowed variable names. (only expression)

    If the expression contains any other variable name that is not part of the list an,
    an exception is raised. It is not required that an expression includes all these
    variables. Additionally, the expression can contain other symbols if they are used
    as parameters.
    """
```

weldx/core.py
```diff
@@ -822,22 +822,64 @@ class GenericSeries:
...
     _required_variables: list[str] = []
-    """Required variable names"""
+    """A list of required variable names.
+
+    The names are only relevant if the created `GenericSeries` is based on expressions.
+    If one or more variables are missing in the expression, an exceptions is raised.
+    Note, that the required symbols must be variables of the expression. Using one or
+    more as a parameter will also trigger an exception.
+    """
```
CagtayFabry, line 840:
```suggestion
    """A list of required variable names. (only expression)

    If one or more variables are missing in the expression, an exceptions is raised.
    Note that the required symbols must be variables of the expression. Using one or
    more as a parameter will also trigger an exception.
    """
```

weldx/core.py
```diff
@@ -822,22 +822,64 @@ class GenericSeries:
...
-    """Function that should be used to adjust a var. input - (f.e. convert to Time)"""
+    """A dictionary of preprocessor functions for input variables during evaluation.
+
+    When evaluating a `GenericSeries`, the passed keyword arguments are checked against
+    the dictionaries keys. If a match is found, the corresponding preprocessor function
+    is called with the variables value and returns the updated value. As an example,
+    this can be used to support multiple time formats. The key might be ``t`` and the
+    preprocessor function would turn the original time data into an equivalent
+    `xarray.DataArray`.
+    """
```
CagtayFabry, line 851:
```suggestion
    """A mapping of variable names to preprocessor functions are applied prior to evaluation.

    When calling a `GenericSeries.evaluate`, the passed keyword arguments are checked against
    the dictionaries keys. If a match is found, the corresponding preprocessor function
    is called with the variables value and returns the updated value. As an example,
    this can be used to support multiple time formats. The key might be ``t`` and the
    preprocessor function would turn the original time data into an equivalent
    `xarray.DataArray`.
    """
```

weldx/core.py
```diff
@@ -822,22 +822,64 @@ class GenericSeries:
...
+    represented by a variable or be part of one of the expressions parameters.
+    """
+
     _required_dimension_units: dict[str, pint.Unit] = {}
-    """Required units of a dimension"""
+    """A dictionary that maps required units to a dimension.
+
+    If a dimension matches one of the keys of this dictionary, it is checked that the
+    units are identical to the requirement. If not, an exception is raised.
+    """
```
CagtayFabry, line 866:
```suggestion
    """A dictionary that maps required unit dimensionality to a dimension.

    If a dimension matches one of the keys of this dictionary, its dimensionality
    is checked against the requirement listed.
    """
```
Do we check against exact units or dimensionality here @vhirtham ?
If units then disregard my suggestions.

weldx/core.py
```diff
@@ -822,22 +822,64 @@ class GenericSeries:
...
     _required_dimension_coordinates: dict[str, list] = {}
-    """Required coordinates of a dimension."""
+    """A dictionary that maps required coordinates to a dimension.
+
+    If a dimension matches one of the keys of this dictionary, it is checked if it has
+    the specified coordinates. An example use-case would be a 3d-space where the
+    coordinates "x", "y" and "z" are required for a spatial dimension.
+    """
```
CagtayFabry, line 873:
```suggestion
    """A dictionary that maps required coordinate values to a dimension.

    If a dimension matches one of the keys of this dictionary, it is checked if it has
    the specified coordinate values. An example use-case would be a 3d-space where the
    coordinates "x", "y" and "z" are required for a spatial dimension.
    """
```

CagtayFabry (approved):
Thank you @vhirtham 👍 
This is very helpful :)

I have added some suggestions, please take a look.

weldx/core.py
```diff
@@ -822,22 +822,62 @@ class GenericSeries:
     """Describes a quantity depending on one or more parameters."""
 
     _allowed_variables: list[str] = []
-    """Allowed variable names"""
+    """A list of allowed variable names. (only expression)
+
+    If the expression contains any other variable name that is not part of the list an,
```
vhirtham, line 827:
```suggestion
    If the expression contains any other variable name that is not part of the list,
```

weldx/core.py
```diff
@@ -822,22 +822,62 @@ class GenericSeries:
...
+    """A list of required variable names. (only expression)
+
+    If one or more variables are missing in the expression, an exceptions is raised.
+    Note that the required symbols must be variables of the expression. Using one or
+    more as a parameter will also trigger an exception.
+    """
 
     _evaluation_preprocessor: dict[str, Callable] = {}
-    """Function that should be used to adjust a var. input - (f.e. convert to Time)"""
+    """A mapping of variable names to preprocessor functions are applied prior to evaluation.
```
vhirtham, line 841:
```suggestion
    """Mapping of variable names to preprocessor functions that are applied prior to evaluation.
```

weldx/core.py
```diff
@@ -822,22 +822,62 @@ class GenericSeries:
...
+    If one or more variables are missing in the expression, an exceptions is raised.
+    Note that the required symbols must be variables of the expression. Using one or
+    more as a parameter will also trigger an exception.
+    """
 
     _evaluation_preprocessor: dict[str, Callable] = {}
-    """Function that should be used to adjust a var. input - (f.e. convert to Time)"""
+    """Mapping of variable names to preprocessor functions that are applied prior to evaluation.
+
+    When calling a `GenericSeries.evaluate`, the passed keyword arguments are checked against
```
vhirtham, line 843:
```suggestion
    When calling `GenericSeries.evaluate`, the passed keyword arguments are checked against
```

weldx/core.py
vhirtham: Dimensionality

weldx/core.py
```diff
@@ -822,22 +822,62 @@ class GenericSeries:
...
+    represented by a variable or be part of one of the expressions parameters.
+    """
+
     _required_dimension_units: dict[str, pint.Unit] = {}
-    """Required units of a dimension"""
+    """A dictionary that maps required unit dimensionality to a dimension.
+
+    If a dimension matches one of the keys of this dictionary, its dimensionality
+    is checked against the requirement listed.
+    """
```
vhirtham, line 864:
```suggestion
    """A dictionary that maps a required unit dimensionality to a dimension.

    If a dimension matches one of the keys of this dictionary, its dimensionality
    is checked against the listed requirement.
    """
```

Issue #201, opened 2020-11-19
GenericSeries support for groove parameters


CagtayFabry:
A long term goal of mine for the API implementation would be to define "dynamic" groove parameters via `MathematicalExpression` and in extension `GenericSeries`.

## current implementations
Currently individual groove shapes can be defined with the use of quantities. Here is a simple example:
```python
groove_1 = get_groove(
    groove_type="IGroove",
    workpiece_thickness=Q_(5, "mm"),
    root_gap=Q_(2, "mm"),
)

groove_2 = get_groove(
    groove_type="IGroove",
    workpiece_thickness=Q_(5, "mm"),
    root_gap=Q_(6, "mm"),
)
```
We create a variable groove geometry by coupling the groove profiles with a trace and specifying the correct interpolation.
```python
v_profile = geo.VariableProfile([groove_1.to_profile(), groove_2.to_profile()],
                            [0, 100],
                            [geo.linear_profile_interpolation_sbs]
                            )

geometry = geo.Geometry(profile=v_profile, trace=trace)
```

## design idea
Here is a rough draft of how I would like this to work. (For this to make sense lets assume we have implemented the `GenericSeries` from #106)

1. create `GenericSeries` that takes a *dimensionless scalar input* `s` that runs from `0` to `1` along the trace. This can be a quantity input later but let's keep it simple for now.
   ```python
   expr = MathematicalExpression(expression="a + b*(1-s)", parameters={"a":Q_(2,"mm"),"b":Q_(4,"mm")})
   gap = GenericSeries(expr)
   ```

1. define a single (dynamic) groove
   ```python
   groove = get_groove(
       groove_type="IGroove",
       workpiece_thickness=Q_(5, "mm"),
       root_gap=gap,
   )
   ```
1. combine with a trace
   ```python
   geometry = geo.Geometry(groove, trace=trace)
   ```

To make this work the `trace` should internally also work with the `s` parameter (I think we already do this without naming it)
So every time the trace gets rasterized, the corresponding `s` value gets passend on to the geometry (think `interp_like`)

### outlook
The same principle could later also be applied to other dynamic parameters (e.g. welding power source settings coupled with welding trace)
Ideally we would support `GenericSeries` classes everywhere we currently support `Quantity` objects. (and it would magically work somehow 🧙😉)

PR #713, opened 2022-03-01
Add `DynamicShapeSegment`

### LINKED ISSUE #140: Trace/Profile segment with SymPy support
We have currently only segment for lines and arcs. Adding more segments for all possible shapes would be time-consuming and each one requires its own ASDF schema. A good solution to that would be a segment that supports SymPy functions to describe the shape.

vhirtham:
## Changes

Add `DynamicShapeSegment` as the new base shape segment. Also, introduce a `DynamicBaseSegment`.

## Related Issues

Closes #140 
#201

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/


vhirtham:
Okay, the `ArcSegment` was a tough nut to crack, but I literally hacked my way through the failing tests. Now it's time for the cleanup and some optimizations.

vhirtham:
I still have some minor optimizations in mind, but then it should be ready to go. Think this is a huge improvement of the geometry package

weldx/geometry.py
```diff
@@ -99,255 +98,404 @@ def _to_list(var) -> list:
...
         Returns
         -------
-        pint.Quantity
-            Starting point
+        DynamicShapeSegment:
+            A self-reference to the modified segment
 
         """
-        return self._points[:, 0]
+        vector = Q_([*vector.m, 0], _DEFAULT_LEN_UNIT)
```
CagtayFabry, line 304:
```suggestion
        vector = np.append(vector, 0)
```
Since `@UREG.check` only checks dimensionality we should avoid `.magnitude` (without `.to()`)

weldx/geometry.py
```diff
@@ -99,255 +98,404 @@ def _to_list(var) -> list:
...
-            The desired distance between two raster points
+        points :
+            2x2 matrix of points. The first column is the
+            starting point and the second column the end point.
 
         Returns
         -------
-        pint.Quantity
-            Array of contour points
+        LineSegment
```
CagtayFabry, line 407: A short example in the `__init__` docstring here would be nice to avoid confusion when creating the 2x2 point matrix

weldx/geometry.py
```diff
@@ -99,255 +98,404 @@ def _to_list(var) -> list:
...
+        if not len(points.shape) == 2:
+            raise ValueError("'points' must be a 2d array/matrix.")
+        if not (points.shape[0] == 2 and points.shape[1] == 2):
+            raise ValueError("'points' is not a 2x2 matrix.")
 
-        # normalized effective raster width
-        nerw = 1.0 / num_raster_segments
+        dat = np.zeros((2, 3))
+        dat[:, :2] = points.transpose()
+        super().__init__(Q_(dat, _DEFAULT_LEN_UNIT), coords={"s": [0, 1]})
```
CagtayFabry, line 417:
```suggestion
        super().__init__(
            np.vstack(points.T, Q_([0.0, 0.0], _DEFAULT_LEN_UNIT)),
            coords={"s": [0, 1]},
        )
```
I am not sure if `points.T` is correct here or just `points` with translation after `vstack`

weldx/geometry.py
```diff
@@ -698,98 +810,32 @@ def apply_transformation(self, matrix: np.ndarray):
...
-        ArcSegment
-            Transformed copy
+        DynamicShapeSegment:
+            A self-reference to the modified segment
 
         """
-        new_segment = copy.deepcopy(self)
-        new_segment.apply_translation(vector)
-        return new_segment
+        self._points = (self._points.transpose() + vector.m).transpose()
```
CagtayFabry, line 837: Is `self._points` not a quantity? I'm a little confused by the `.magnitude` here

weldx/geometry.py
```diff
@@ -99,255 +98,404 @@ def _to_list(var) -> list:
...
+            The length at the specified value.
 
         """
-        points = np.transpose(np.array([point_start.m, point_end.m], dtype=float))
-        return cls(Q_(points, _DEFAULT_LEN_UNIT))
+        if self._series.is_expression:
+            length = self._length_expr.evaluate(max_coord=position).data
+        else:
+            length = self._len_section_disc(position=position)
+        if length.m <= 0:
```
CagtayFabry, line 216:
```suggestion
        if length <= 0:
```

CagtayFabry (changes requested):
the sympy stuff seems to work really well 👍 

I have added some changes in places where I think `pint` can be used instead of numpy

The constant `.transpose` with the points format is a little confusing at times

weldx/core.py
```diff
@@ -778,9 +776,6 @@ def __post_init__(self):
         if not isinstance(self.values, (pint.Quantity, xr.DataArray)):
             self.values = Q_(self.values)
 
-        if self.symbol is not None and len(self.symbol) > 1:
-            raise ValueError(f"Cannot use symbol {self.symbol}")
-
```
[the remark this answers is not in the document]
vhirtham: To my knowledge, this is true if you use a string expression. But you can use `sympy.symbols` to create symbol names of arbitrary length and add them to an expression or create a new one. I used this for the translation function of the shape segment. There we do not know, which parameter names will be already in use and there might also be more than one translation. So I just named the translation parameters 'translation_' + number. Of course, this triggered the exception, so I removed it. Also, the safeguard should not work as expected since `"an * expression"` will create single character symbols anyway, right?

vhirtham:
> I have added some changes in places where I think `pint` can be used instead of numpy

I will check it carefully, but some things I saw seamed reasonable.

 
> The constant `.transpose` with the points format is a little confusing at times

Unfortunately, we made a "bad" design decision back then. IIRC the goal was to simply pass `points[0]` and `points[1]` to matplotlibs `plot` function instead of `points[:, 0]` and `points[:, 1]`. So the single point data isn't the innermost dimension as in the new classes. Changing that would require a lot of tests and notebooks to be adjusted :/

weldx/geometry.py
vhirtham: Shouldn't that trigger a dimensionality error? Well, let's try.

weldx/geometry.py
vhirtham: The internal format is without a unit. The unit is removed by `ureg.wraps`. I'll check this line again.

weldx/core.py
```diff
@@ -980,7 +975,7 @@ def __init__(
             if dims is not None and not isinstance(dims, list):
                 raise ValueError(f"Argument 'dims' must be list of strings, not {dims}")
             self._init_discrete(obj, dims, coords)
-        elif isinstance(obj, (MathematicalExpression, str)):
+        elif isinstance(obj, (MathematicalExpression, str, sympy.Expr)):
```
marscher, line 978: This can be a "typedef", right?

weldx/geometry.py
```diff
@@ -99,255 +98,402 @@ def _to_list(var) -> list:
...
-        points :
-            2x2 matrix of points. The first column is the
-            starting point and the second column the end point.
-
-        Returns
-        -------
-        LineSegment
+        series:
+            A `~weldx.core.SpatialSeries` that describes the trajectory of the shape
+            segment. Alternatively, one can pass every other object that is valid as
```
marscher, line 121: That is cryptic/fragmented doc string. I want to know explicitly which types and contents I have to put here. It is fine to cross-ref things of course, but it should not be mandatory to read another class docstring first to understand this one.

weldx/geometry.py
CagtayFabry: I think `@UREG.check` doesn't remove units like `@UREG.wraps` would

weldx/core.py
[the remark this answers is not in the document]
CagtayFabry: Since `SeriesParameter` is just a container to represent parameters/coordiantes/symbols there is no `sympy` involved here.

But apparently it does work with multi character strings anyway? So it's not really needed
```python
from weldx import Q_, GenericSeries

GenericSeries(
    "a*time + b",
    parameters=dict(a=Q_([3, 0, 0], "mm/s"), b=Q_([0, 2, 2], "cm")),
    units=dict(time="s"),
)
#> <GenericSeries>
#> Expression:
#> 	a*time + b
#> Parameters:
#> 	a = [3 0 0] mm / s
#> 	b = [0 2 2] cm
#> Free Dimensions:
#> 	time in s
#> Other Dimensions:
#> 	['dim_0']
#> Units:
#> 	mm
#> 
```

weldx/geometry.py
vhirtham: No it doesn't, but the `__init__` currently uses `wraps`. I can check if changing that to `check` is an easy fix.

vhirtham: I now changed it and it seems to work. Now some internals are quantities and the usage of the magnitude could be further reduced

weldx/geometry.py
```diff
@@ -276,11 +276,10 @@ def rasterize(self, raster_width: pint.Quantity) -> pint.Quantity:
             Array of raster points
 
         """
+        raster_width = Q_(raster_width)
         if raster_width <= 0:
             raise ValueError("'raster_width' must be a number greater than 0.")
-        num_pts = (
-            np.round((self._length / raster_width).to_base_units()).astype(int) + 1
-        )
+        num_pts = np.round((self._length / raster_width).m).astype(int) + 1
```
CagtayFabry, line 282: This doesn't work reliably when units differ, see the following examples:
```python
import numpy as np
from weldx import Q_

_length = Q_("10m")
raster_width = Q_("2mm")

num_pts = np.round((_length / raster_width).magnitude).astype(int) + 1
num_pts = max(num_pts, 2)
num_pts
#> 6

num_pts = np.round((_length / raster_width)).astype(int) + 1
num_pts = max(num_pts, 2)
num_pts
#> <Quantity(5001.0, 'dimensionless')>

num_pts = np.round(_length / raster_width).to_base_units().magnitude.astype(int) + 1
num_pts = max(num_pts, 2)
num_pts
#> 5001

vals = np.linspace(0.0, 1.0, num=num_pts, endpoint=True)
vals
#> array([0.000e+00, 2.000e-04, 4.000e-04, ..., 9.996e-01, 9.998e-01,
#>        1.000e+00])
```

Using `.magnitude` without `.to()` or `.to_base_units()` first is too unreliable

CagtayFabry (changes requested):

weldx/geometry.py
CagtayFabry: yes that is what happens

CagtayFabry (approved):

weldx/core.py
vhirtham: While this is a good idea, the `MathematicalExpression` makes it prone for circular imports. At least when I tried it, those occurred.

marscher: you could also just write it as a str in the typedef, if it causes a circular import. The type annotation system is going to be string only in future Python versions, right?

weldx/geometry.py
```diff
@@ -99,258 +98,405 @@ def _to_list(var) -> list:
...
-        """Get the segments points in form of a 2x2 matrix.
+        if self._series.is_expression:
+            exp = self._series.data.expression
+            params = self._series.data.parameters
+
+            # Segment might have been translated already
+            p_idx = 0
+            p_name = f"translation_{p_idx}"
+            while p_name in params:
+                p_name = f"translation{(p_idx:= p_idx +1)}"
```
marscher, line 314: walrus inside a fstring *yippie*

weldx/geometry.py
```diff
@@ -99,258 +98,405 @@ def _to_list(var) -> list:
...
 
         """
-        return self._points
+        new_segment = copy.deepcopy(self)
+        return new_segment.apply_translation(vector)
 
-    def apply_transformation(self, matrix):
-        """Apply a transformation matrix to the segment.
+    def apply_transformation(self, matrix: np.ndarray) -> DynamicShapeSegment:
+        """Apply an in-place transformation to the segment using a matrix.
```
marscher, line 342: Just an idea, could we also transform a segment using an `Expression` here?

weldx/geometry.py
vhirtham: done

marscher:
Thanks a lot, this is a really cool new feature!

PR #714, opened 2022-03-02
set networkx<2.7

CagtayFabry:
## Changes
- set `networkx<2.7`
- link `asdf` docs to version 2.9.2

CSM graph plotting error with networkx 2.7 see https://github.com/BAMWelDX/weldx/runs/5389199108?check_suite_focus=true#step:8:75

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -92,6 +92,8 @@ dependencies
 
 -  add ``bidict`` dependency [:pull:`618`]
 
+-  set ``networkx <2.7`` for plotting compatibility (for now) [:pull:`714`]
+
 ********************
  0.5.2 (18.11.2021)
 ********************
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -280,7 +280,7 @@ intersphinx_mapping = {
     # "numba": ("https://numba.pydata.org/numba-doc/latest", None),
     "pint": ("https://pint.readthedocs.io/en/latest", None),
     "jsonschema": ("https://python-jsonschema.readthedocs.io/en/stable/", None),
-    "asdf": ("https://asdf.readthedocs.io/en/latest/", None),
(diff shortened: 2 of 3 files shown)


marscher:
There is indeed a bug in the label drawing function of networkx. The function expects the labels to be a dictionary of len(edges) or none, but that is not really being checked. I suggest, that we do not pass empty dicts to networkx, but just none, in case the labels are not set.

CagtayFabry:
> There is indeed a bug in the label drawing function of networkx. The function expects the labels to be a dictionary of len(edges) or none, but that is not really being checked. I suggest, that we do not pass empty dicts to networkx, but just none, in case the labels are not set.

Good to know, thank you. The proposed fix sounds like a good solution for now if possible

I think we should take another look at the graph drawing functions and try again to base them on `graphviz` at some point
I gave it a little try and the the setup with conda seems to work easy now using `graphviz pygraphviz` stack

marscher:
Why merge it without a review then? I just prepared the workaround :D

marscher:
Regarding graphviz, I just saw a comment of the main networkx developer saying that the conda support for graphviz is horrible and upstream refused to fix the situation. Eventually we want to reconsider this.

marscher:
https://github.com/networkx/networkx/issues/4552

CagtayFabry:
I did a quick test with `mamba install graphviz pygraphviz -c conda-forge` and it worked (even on windows!)

It was really inconsistent in the past to get running which was also one of the main reasons why we decided not to depend on it but I think this looks like it has changed

Since we do have quite a few graphs to draw I think it is worth looking into again

CagtayFabry:
> Why merge it without a review then? I just prepared the workaround :D

Sorry I was a bit in a hurry :/
feel free to undo in a PR with the workaround

PR #722, opened 2022-03-14
allow networkx >2.7.0

CagtayFabry:
## Changes
- our code should be good again after 2.7.1 release of networkx

## Related Issues
#714 

## Checks

- [x] updated CHANGELOG.rst


marscher (approved):

CagtayFabry:
> Do we need to patch this for the already released conda-forge package?

The problem only comes up with 2.7.0 which probably never get's installed now that 2.7.1 is released but I will update the recipe just to keep it in sync with setuptools

PR #723, opened 2022-03-14
reshape point data for `meshio` file export

CagtayFabry:
## Changes
- This reshapes all point data to be in the shape `(n,3)` when exporting to other file formats using `meshio`.
Since file formats might not work with our 'profile based' dimensions like `(p,n,3)`this should be a safe way to export the pointcloud.

## Checks

- [x] updated CHANGELOG.rst


CagtayFabry:
The absolute safest way would probably be `points=self.coordinates.transpose(..., "c").data.reshape(-1, 3),` but that should already happen in `SpatialData` so it should be fine.

vhirtham (approved):

PR #727, opened 2022-03-24
Schema media file

### LINKED ISSUE #686: schemas for images/videos
So far we don't have a schema representation of image or video data/files. Both images & videos share some similarities on how we might implement them so I will group them in this issue for now.

I guess the most common use cases would be

1.  adding an image/video as an external file (blob storage or external reference)
2.  adding an image/video as a raw data matrix (without compression)

To implement the first use-case we should probably add a `MIME type` field to the `file` schema to make thing easier to handle.
The second use-case is mostly suited for raw recordings, if available to input into image processing/ML applications.

Images as `file` should be straight forward, for raw data I suggest defining a suitable array format with coordinates such as `row/col` or similar.

Videos should also be easy but we should think about how to handle/read the associated timeline of the video into our data structures. A Video could then be a `file` with an additional `time` attribute.
For raw data, I guess it would make sense to have the video be a series of images with timestamps.

For both raw data we should look into which formats we might want to start with, I suggest at least `grey-scale` and `RGB` should be supported (maybe `HSL`?)

marscher:
## Changes

Adds a schema to describe metadata of a media typed file. The actual file information is handled by the ExternalFile
schema. The underlying media can be accessed by integer based indexing.

The ExternalFile schema now also stores the mime type (which is nothing more than a standardized version of the files extension, indicating the files content).

## Related Issues

Closes #686

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [ ] updated doc/
- [x] update manifest file


marscher:
Actually I wanted to open this as a draft...

CagtayFabry:
thank you for getting this started @marscher 🙌 

Just a few quick notes on the ASDF side of things:
- when changing version numbers of existing schemas (like `id: "asdf://weldx.bam.de/weldx/schemas/core/file-0.1.1"` here) we have to keep the old schema file as is and add a new file for this version (create `weldx/schemas/weldx.bam.de/weldx/core/file-0.1.0.yaml`)
(we don't need to change version numbers for schema files that have not yet been released on pip/conda)
- run `devtools/scripts/update_manifest.py` after adding/removing schema files to update the current manifest
- make sure the `Converter` is associated with and set up to handle all schema versions in the `tags` property:
Is this case change the old version (only supports schema version) https://github.com/BAMWelDX/weldx/blob/a99742a1532ab977d547625cd13e6c9fb096361a/weldx/tags/core/file.py#L142-L144
to
```python
    tags = ["asdf://weldx.bam.de/weldx/schemas/core/file-0.1.*"]
    types = [ExternalFile]
```

devtools/scripts/update_manifest.py
```diff
@@ -12,8 +12,9 @@
 
 
 def update_manifest(
-    search_dir: str = "../../weldx/schemas",
-    out: str = f"../../weldx/manifests/weldx-{WELDX_EXTENSION_VERSION}.yaml",
+    search_dir: str = Path(__file__).parent / "../../weldx/schemas",
+    out: str = Path(__file__).parent
+    / f"../../weldx/manifests/weldx-{WELDX_EXTENSION_VERSION}.yaml",
```
CagtayFabry, line 17: thanks! 👏

weldx/asdf/types.py
```diff
@@ -88,8 +88,8 @@ def __new__(mcs, name, bases, attrs):
 class WeldxConverter(Converter, metaclass=WeldxConverterMeta):
     """Base class to inherit from for custom converter classes."""
 
-    tags: list[str] = []
-    types: list[Union[type, str]] = []
+    tags: tuple[str] = None  # note: this will be updated by WeldxConverterMeta.
+    types: tuple[Union[type, str]] = ()
```
CagtayFabry, line 92: thanks for fixing this oversight on my part ... 👍

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - foo
+    - bar
+
+type: object
+properties:
+  file:
+    description: |
+      The basic file information/content.
+    tag: "asdf://weldx.bam.de/weldx/tags/core/file-0.1.1"
+  recorded_at:
```
CagtayFabry, line 24: I think in the namespace of `weldx` this could be called `reference_time`

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
```diff
@@ -0,0 +1,38 @@
...
+      Time when this media was recorded.
+    tag: "asdf://weldx.bam.de/weldx/tags/time/timestamp-0.1.*"
+  fps:
+    description: |
+      In case of time-dynamic media, the time resolution (frames per second).
+    tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.0"
+
+propertyOrder: [file, recorded_at, fps]
+anyOf:
+  - required: [file, recorded_at]
```
CagtayFabry, line 35: are you planning on adding different variations here?
Otherwise the `anyOf` block is not needed

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
```diff
@@ -0,0 +1,38 @@
...
+  -
+    - foo
+    - bar
+
+type: object
+properties:
+  file:
+    description: |
+      The basic file information/content.
+    tag: "asdf://weldx.bam.de/weldx/tags/core/file-0.1.1"
```
CagtayFabry, line 23: we can (and probably should) add the requirement here that `mimetype` of this file should follow `pattern: video/*`

But we can add that later

weldx/tags/core/file.py
```diff
@@ -187,3 +189,6 @@ def from_yaml_tree(self, node: dict, tag: str, ctx):
                     "The stored hash does not match the stored contents' hash."
                 )
         return ExternalFile(**node)
+
+    def select_tag(self, obj, tags, ctx):
+        return get_highest_tag_version(tags)
```
CagtayFabry, line 194: I had hoped I fixed that sorting oder somewhere already... I guess not, thanks !

cref: https://github.com/asdf-format/asdf/issues/1086

can you check if
```python
    def select_tag(self, obj, tags, ctx):
        return sorted(tags)[-1]
```
also works here? `get_highest_tag_version`is certainly the safest option but also has quite the overhead

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
marscher: Thanks. I will rename it accordingly.

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
marscher: So far, I do not. Will remove it then.

weldx/tags/core/file.py
marscher: we could add a default implementation for select_tag, right? This could be a bit fragile, if some converters are relying on a certain behaviour being changed by enforcing this.
I like the response in the linked issue, saying that asdf users are encouraged to release a new version of the extension, if some schema versions changed.

CagtayFabry: I have implemented this in https://github.com/BAMWelDX/weldx/pull/733, no idea why I missed that when I looked into it

One only has to be careful when there are other non `weldx` tags in the list (e.g. quantity)

marscher:
the pims lib is not yet ready for Python-3.10

weldx/schemas/weldx.bam.de/weldx/core/media_file-0.1.0.yaml
```diff
@@ -29,17 +31,18 @@ properties:
     description: |
       In case of time-dynamic media, the time resolution (frames per second).
     tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.0"
+    wx_unit: "1/s"
   resolution:
     description: |
       Media resolution in pixels.
     type: array
+    wx_shape: [n. m]
```
CagtayFabry, line 39: is this the same as `[n, m]`? (I am not sure how the YAML parsing works in this case)

if this is to store the resolution as a shape-like list of two integers the following is works better:
```suggestion
    type: array
    items:
      type: integer
      number: 2
```

(looking for the asdf/json array syntax I stumbled upon [this example](https://asdf.readthedocs.io/en/2.11.1/asdf/features.html#custom-schemas) in the ASDF docs that is probably pretty close to what we also want 😄 maybe look into that as well for reference 👍)

I *think* a `wx_shape` implementation would be more in line with
```suggestion
    type: array
    wx_shape: [1, 1]
```

marscher:
@CagtayFabry I'd like to merge this soon and post-pone some todo's in follow up issues.

CagtayFabry:
> @CagtayFabry I'd like to merge this soon and post-pone some todo's in follow up issues.

Which ToDos are still open here? The ASDF examples?

marscher:
> Which ToDos are still open here? The ASDF examples?

Basically, yes.

CagtayFabry:
> > Which ToDos are still open here? The ASDF examples?
> 
> Basically, yes.

A simple ASDF example would be nice as those also run during the schema tests
Other than that I think it looks good for now? Is the `mime_type` ASDF pattern working?

marscher:
Right. I'd generate the example simply by instanciating a media file, serialize it and put the contents into the schema file.

marscher:
> Right. I'd generate the example simply by instanciating a media file, serialize it and put the contents into the schema file.

@CagtayFabry in [6a04d5e](https://github.com/BAMWelDX/weldx/pull/727/commits/6a04d5ee5b2a98fde34dea4d39b5775eae300f88) I had to use the "old" style tag syntax. Otherwise the asdf yaml parser would raise some "undefined tag handle" error:

```
E   yaml.parser.ParserError: while parsing a node
E   found undefined tag handle
E     in "<file>", line 5, column 9
```
Do you know what might be going wrong there? I was thinking that the header of the example (used ASDF standard) could be referring to an old version?

PR #730, opened 2022-03-28
add git diff check

CagtayFabry:
## Changes
- order the manifest contents outputs of `devtools/scripts/update_manifest.py` by filename for consistent manifest generation
- add a `git diff` check in `pytest_asdf.yml` workflow against the output of `devtools/scripts/update_manifest.py` (if this fails, try running `devtools/scripts/update_manifest.py` locally and commit changes)

### DIFF
--- a/.github/workflows/pytest_asdf.yml
+++ b/.github/workflows/pytest_asdf.yml
@@ -41,7 +41,7 @@ jobs:
       - name: git diff
         run: |
           git diff-index HEAD
-          git diff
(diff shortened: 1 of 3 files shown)


CagtayFabry:
This should make it easier to find bugs in the asdf manifest and `Converter` setups

If the action fails at the `git diff` step run the `update_manifest.py` script (must be executed in the `devtools/scripts` folder) to fix

PR #731, opened 2022-03-31
Strict quantities

vhirtham:
## Changes

Units are now mandatory and no longer optional in the `CoordinateSystemManager` and the `LocalCoordinateSystem`

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/asdf/cli/welding_schema.py
+++ b/weldx/asdf/cli/welding_schema.py
@@ -98,7 +98,7 @@ def single_pass_weld_example(
 
     rot = WXRotation.from_euler(seq="x", angles=180, degrees=True)
 
-    coords = [tcp_start_point.magnitude, tcp_end_point.magnitude]
+    coords = Q_([tcp_start_point.magnitude, tcp_end_point.magnitude], "mm")
 
     tcp_wire = lcs(coordinates=coords, orientation=rot, time=[t_start, t_end])
 
@@ -109,7 +109,7 @@ def single_pass_weld_example(
         lcs=tcp_wire,
     )
 
-    tcp_contact = lcs(coordinates=[0, 0, -10])
+    tcp_contact = lcs(coordinates=Q_([0, 0, -10], "mm"))
 
     # add the workpiece coordinate system
     csm.add_cs(
--- a/weldx/tests/_helpers.py
+++ b/weldx/tests/_helpers.py
@@ -3,6 +3,7 @@
 from typing import Any
 
 import numpy as np
+import pint
 
 from weldx.constants import Q_
 from weldx.transformations import LocalCoordinateSystem, WXRotation
@@ -110,8 +111,10 @@ def matrix_is_close(mat_a, mat_b, abs_tol=1e-9) -> bool:
         True or False
 
     """
-    mat_a = np.array(mat_a, dtype=float)
-    mat_b = np.array(mat_b, dtype=float)
+    if not isinstance(mat_a, pint.Quantity):
+        mat_a = np.array(mat_a, dtype=float)
+    if not isinstance(mat_b, pint.Quantity):
+        mat_b = np.array(mat_b, dtype=float)
 
     if mat_a.shape != mat_b.shape:
         return False
--- a/weldx/tests/asdf_tests/test_asdf_core.py
+++ b/weldx/tests/asdf_tests/test_asdf_core.py
@@ -284,7 +284,7 @@ def test_local_coordinate_system_shape_violation():
         coords={"u": ["x", "y", "z"], "v": [0, 1, 2]},
     )
     coordinates = xr.DataArray(
-        data=[1, 2],
+        data=Q_([1, 2], "mm"),
         dims=["c"],
         coords={"c": ["x", "y"]},
     )
@@ -302,7 +302,7 @@ def test_local_coordinate_system_shape_violation():
         coords={"c": ["x", "y"], "v": [0, 1]},
     )
     coordinates = xr.DataArray(
-        data=[1, 2, 3],
+        data=Q_([1, 2, 3], "mm"),
         dims=["u"],
         coords={"u": ["x", "y", "z"]},
     )
@@ -320,18 +320,18 @@ def test_local_coordinate_system_shape_violation():
 def get_example_coordinate_system_manager():
     """Get a consistent CoordinateSystemManager instance for test purposes."""
     csm = tf.CoordinateSystemManager("root")
-    csm.create_cs("lcs_01", "root", coordinates=[1, 2, 3])
+    csm.create_cs("lcs_01", "root", coordinates=Q_([1, 2, 3], "mm"))
     csm.create_cs(
         "lcs_02",
         "root",
         orientation=WXRotation.from_euler("z", np.pi / 3).as_matrix(),
-        coordinates=[4, -7, 8],
+        coordinates=Q_([4, -7, 8], "mm"),
     )
     csm.create_cs(
         "lcs_03",
         "lcs_02",
         orientation=WXRotation.from_euler("y", np.pi / 11),
-        coordinates=[4, -7, 8],
+        coordinates=Q_([4, -7, 8], "mm"),
     )
     return csm
 
@@ -349,7 +349,9 @@ def test_coordinate_system_manager(copy_arrays, lazy_load):
 
 
 def get_coordinate_system_manager_with_subsystems(nested: bool):
-    lcs = [tf.LocalCoordinateSystem(coordinates=[i, -i, -i]) for i in range(12)]
+    lcs = [
+        tf.LocalCoordinateSystem(coordinates=Q_([i, -i, -i], "mm")) for i in range(12)
+    ]
 
     # create global system
     csm_global = tf.CoordinateSystemManager("base", "Global System", "2000-06-08")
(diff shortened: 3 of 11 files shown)


setup.cfg
```diff
@@ -43,7 +43,7 @@ install_requires =
     xarray >=0.19
     scipy >=1.4,!=1.6.0,!=1.6.1
     sympy >=1.6
-    pint >=0.18
+    pint ==0.18
```
vhirtham, line 46: Of course, pint had to update to a new incompatible version when I think I finally got everything working. This was hilarious. The problem with the new version is that now uses full names for some reason which causes some doctests to fail. Have to figure out why and unpin the version afterward.

.pre-commit-config.yaml
```diff
@@ -34,7 +34,7 @@ repos:
     - id: upgrade-type-hints
       args: [ '--futures=true' ]
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
```
vhirtham, line 37: Don't know why the friendly pre-commit bot hasn't done it yet, but without this, we get errors during the black run (see some of the commits in the middle)

vhirtham:
@CagtayFabry No hurry to review. Do it if you have the time ;)

.pre-commit-config.yaml
vhirtham: Seems like someone wants to mock me with this PR... today the pre-commit fix arrived...

setup.cfg
CagtayFabry: I cant look into it in detail right now but I guess with 'full names' you mean the unit format?
So `meter` instead of `m`?

The use of 'short notation' is `weldx` specific for our unit registry and done here https://github.com/BAMWelDX/weldx/blob/af9df21e2efe10613847cb1d0c6f4b122bfbc880/weldx/constants.py#L29-L31

Maybe the syntax or notation has changed to set this setting

.pre-commit-config.yaml
CagtayFabry: this is probably caused by the pre-commit update setting 🤖 
https://github.com/BAMWelDX/weldx/blob/af9df21e2efe10613847cb1d0c6f4b122bfbc880/.pre-commit-config.yaml#L8

weldx/tags/core/geometry/spatial_data.py
```diff
@@ -17,11 +17,20 @@ def to_yaml_tree(self, obj: SpatialData, tag: str, ctx) -> dict:
...
         tree = copy(obj.__dict__)  # shallow copy so we dont change original object
         if tree["coordinates"].ndim <= 2:
-            tree["coordinates"] = tree["coordinates"].values
+            tree["coordinates"] = tree["coordinates"].data
         return tree
 
     def from_yaml_tree(self, node: dict, tag: str, ctx) -> SpatialData:
         """Reconstruct from yaml node."""
+        # TODO: This is a workaround until the single_pass_weld.wx file is adjusted,
+        #  that is needed for the tutorials
```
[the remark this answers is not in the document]
CagtayFabry: we have to update the file (and notebook) for `0.6.0` at some point anyway, I haven't done it yet
it doesn't run with the current master
https://github.com/BAMWelDX/IIW2021_AA_CXII/actions/runs/2084693688

You can start the update on a new branch over at https://github.com/BAMWelDX/IIW2021_AA_CXII and link to that file in the tutorials for now

setup.cfg
vhirtham: I will take a look

weldx/geometry.py
```diff
@@ -2627,10 +2627,13 @@ def to_file(
...
-            trace_raster_width=trace_raster_width,
-            stack=False,
+        raster_data = Q_(
+            self._rasterize_constant_profile(
+                profile_raster_width=profile_raster_width,
+                trace_raster_width=trace_raster_width,
+                stack=False,
+            ),
+            _DEFAULT_LEN_UNIT,
         )
```
CagtayFabry, line 2637: in codeblocks like these the added indent from `Q_()` sometimes makes it a little harder to read imo
In that case the `* _DEFAULT_LEN_UNIT` syntax might make it more readable (it is also better to see in the diffs)

weldx/schemas/weldx.bam.de/weldx/core/geometry/spatial_data-0.1.1.yaml
```diff
@@ -70,6 +73,8 @@ properties:
         wx_shape: [(~), 3]
       - tag: "asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"
         wx_shape: [..., 3]
+      - tag: "asdf://weldx.bam.de/weldx/tags/units/quantity-0.1.*"
+        wx_shape: [..., 3]
```
CagtayFabry, line 77: we probably need the `wx_units` validator here as well? (and for the `data_array`)

weldx/tags/core/transformations/local_coordinate_system.py
```diff
@@ -57,6 +60,10 @@ def from_yaml_tree(self, node: dict, tag: str, ctx):
         coordinates = node.get("coordinates")
         if coordinates is not None and not isinstance(coordinates, TimeSeries):
             coordinates = node["coordinates"].data
+            # workaround until the single_pass_weld.wx file needed it the tutorials is
+            # adjusted
+            if not isinstance(coordinates, pint.Quantity):
+                coordinates = Q_(coordinates, "mm")
```
CagtayFabry, line 66: if possible we should use something like `_DEFAULT_LEN_UNIT` everywhere in the codebase

CagtayFabry (changes requested):
great 🚀 !

I think the only things missing are that we need new schema file versions for LCS and CSM with that added unit restriction via `wx_units`

I skipped over it quickly but I didn't see any changes in the `strict` modes @vhirtham ?

weldx/tags/core/transformations/local_coordinate_system.py
vhirtham: True, but since this workaround is hopefully going to disappear soon, it shouldn't matter that much ;)

vhirtham:
> I skipped over it quickly but I didn't see any changes in the `strict` modes @vhirtham ?

This is because all the changed classes also supported other inputs like `DataArray`, so we couldn't use the pint wrapper and checked the units manually. Title of the PR might be a bit misleading

CagtayFabry:
> 

that makes sense, thank you

setup.cfg
vhirtham, line 46:
```suggestion
    pint >=0.18
```

vhirtham:
> I think the only things missing are that we need new schema file versions for LCS and CSM with that added unit restriction via `wx_units`

Only the LCS needed an update, since the CSMs unit restrictions are just a result of the stored LCS instances. I updated the LCS schema version.

By the way, should we have also created a new version for the `SpatialData` schema?

marscher:
Thanks you for this improvement. 
At the very end I suggest adding a test, which checks the raising of the exceptions, if no unit was provided.

weldx/schemas/weldx.bam.de/weldx/core/geometry/spatial_data-0.1.1.yaml
CagtayFabry: If we tell the `wx_unit` validator where to look for the `units` property in the tree (it should be `node["coordinates"]["data"]` here) it should work:
```yaml
      - tag: "asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"
        wx_shape: [..., 3]
        wx_unit:
          coordinates:
            data: "m"
```

tutorials/01_01_introduction.ipynb
```diff
@@ -80,15 +84,26 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
+   "id": "8d197413",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
marscher: I wonder how output still ends up being committed, if we have this precommit hook to clean the notebooks?

CagtayFabry: eventually we should create our own little local pre-commit script that does what we want :)

vhirtham: well, I also ran the clean notebooks script, but it obviously didn't do the trick. Since there are no relevant changes to this file, I can simply copy the original version from master into this branch. Anyways, shouldn't be important for this PR.

weldx/geometry.py
```diff
@@ -2696,7 +2706,7 @@ def from_file(file_name: Union[str, Path]) -> SpatialData:
         mesh = meshio.read(file_name)
         triangles = mesh.cells_dict.get("triangle")
 
-        return SpatialData(mesh.points, triangles)
+        return SpatialData(Q_(mesh.points, "mm"), triangles)
```
CagtayFabry, line 2709:
```suggestion
        return SpatialData(Q_(mesh.points, _DEFAULT_LEN_UNIT), triangles)
```

vhirtham: I think we shouldn't do this here because we are dealing with files that are not generated by WelDX. So we should stick with the CAD standard (mm). Even though they are currently the same, it would probably better to introduce something like `_CAD_LEN_UNIT`

vhirtham: Okay, saw you already handled it in a later commit

CagtayFabry (approved):

CagtayFabry:
the only problem left is probably opening old weldx files now since that would require adding the default quantities in the `from_yaml` conversion

@vhirtham please add a list to the PR comment indicating which classes are affected by this PR

PR #733, opened 2022-04-05
set default tag sorting order

CagtayFabry:
## Changes
- update the default sorting order of ``select_tag`` for ``WeldxConverter``
- use full tag naming schema in all explicit converters

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/weldx/asdf/types.py
+++ b/weldx/asdf/types.py
@@ -88,14 +88,17 @@ class WeldxConverterMeta(type(Converter)):
 class WeldxConverter(Converter, metaclass=WeldxConverterMeta):
     """Base class to inherit from for custom converter classes."""
 
-    tags: list[str] = []
-    types: list[Union[type, str]] = []
+    tags: tuple[str] = None  # note: this will be updated by WeldxConverterMeta.
+    types: tuple[Union[type, str]] = ()
 
     def to_yaml_tree(self, obj, tag: str, ctx: SerializationContext):
         raise NotImplementedError
 
     def from_yaml_tree(self, node: dict, tag: str, ctx: SerializationContext):
         raise NotImplementedError
+
+    def select_tag(self, obj, tags, ctx):
+        return sorted(tags)[-1]
 
     @classmethod
     def default_class_display_name(cls):
--- a/weldx/tags/core/data_array.py
+++ b/weldx/tags/core/data_array.py
@@ -12,8 +12,7 @@ from weldx.asdf.util import _get_instance_shape
 class XarrayDataArrayConverter(WeldxConverter):
     """Serialization class for xarray.DataArray."""
 
-    name = "core/data_array"
-    version = "0.1.0"
+    tags = ["asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"]
     types = [DataArray]
 
     def to_yaml_tree(self, obj: DataArray, tag: str, ctx) -> dict:
--- a/weldx/tags/core/transformations/local_coordinate_system.py
+++ b/weldx/tags/core/transformations/local_coordinate_system.py
@@ -1,3 +1,4 @@
+from weldx.asdf.constants import WELDX_TAG_URI_BASE
 from weldx.asdf.types import WeldxConverter
 from weldx.core import TimeSeries
 from weldx.tags.core.common_types import Variable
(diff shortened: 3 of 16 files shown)


weldx/tags/core/data_array.py
```diff
@@ -12,8 +12,7 @@
 class XarrayDataArrayConverter(WeldxConverter):
     """Serialization class for xarray.DataArray."""
 
-    name = "core/data_array"
-    version = "0.1.0"
+    tags = ["asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"]
```
marscher, line 15: should these now be tuples as well (as the type hint in the base class suggests.). I added this in #727, since I thought it make more sense to indicate that this is immutable.

weldx/tags/core/transformations/local_coordinate_system.py
```diff
@@ -7,8 +8,7 @@
 class LocalCoordinateSystemConverter(WeldxConverter):
     """Serialization class for weldx.transformations.LocalCoordinateSystem"""
 
-    name = "core/transformations/local_coordinate_system"
-    version = "0.1.0"
+    tags = [WELDX_TAG_URI_BASE + "core/transformations/local_coordinate_system-0.1.*"]
```
marscher, line 11: Why now use this pattern?

marscher (approved):

weldx/tags/core/transformations/local_coordinate_system.py
CagtayFabry: the old format was a legacy asdf style that is now superseded for full tag definitions. It is not really useful when mixing different tags/versions which is why I want to get away form it (the `*` syntax makes it easier now anyway)

In case you were referring to the use of `WELDX_TAG_URI_BASE`, that was purely due to line length here (I am not sure we should actually do that)

weldx/tags/core/data_array.py
CagtayFabry: true, probably `tuple` is fitting

I only used `list` here because this is how it is in the asdf codebase.
I **think** the `tags` attribute gets 'resolved' in asdf and ends up as a list anyway. I don't know if that breaks with tuples here.

Frankly I would just stick to what asdf is doing to avoid any troubles if we can

CagtayFabry:
please feel free to make further changes and/or merge @marscher 

As long as the tests pass we should be good here 👍

weldx/tags/core/transformations/local_coordinate_system.py
marscher: I was referring to TAG_URI_BASE pattern. It looks like something special compared to the other converters, but it is not. So I'd suggest just to wrap the long line then.

weldx/asdf/types.py
```diff
@@ -88,15 +88,18 @@ def __new__(mcs, name, bases, attrs):
...
+    types: tuple[Union[type, str]] = ()
 
     def to_yaml_tree(self, obj, tag: str, ctx: SerializationContext):
         raise NotImplementedError
 
     def from_yaml_tree(self, node: dict, tag: str, ctx: SerializationContext):
         raise NotImplementedError
 
+    def select_tag(self, obj, tags, ctx):
+        return sorted(tags)[-1]
```
marscher, line 101: Can we be sure, that the tags/objs involved are part of weldx? Should we make this assertion here, or in converters deriving from this, treating a special case?

CagtayFabry: we cannot really be sure of this here in cases where we also support 'reading' other tags into weldx properties (e.g. asdf/astropy quantities)

Right now I have implemented this manually in places where we support other tags
https://github.com/BAMWelDX/weldx/blob/96d9bbcc4065009c3414230932131deb935c63db/weldx/tags/units/pint_quantity.py#L40-L42

This code is probably a more reasonable default so there is one less thing to worry about in the Converter implementation (and *if* we ever want to write out a non weldx tag we can implement this in the converter)

To be extra save we could also guard the list sorting like so
```python
def select_tag(self, obj, tags, ctx): 
    weldx_tags = [tag for tag in tags if tag.startswith(WELDX_TAG_URI_BASE)]
    if weldx_tags:
        return sorted(weldx_tags)[-1]
    return sorted(tags)[-1] 
```

PR #739, opened 2022-04-19
custom `wx_unit` validation behavior

### LINKED ISSUE #95: custom wx_unit validation behavior
it might be convenient to define `wx_unit` behavior for tags beside `quantity`
Some ideas:
- `number` and `ndarray` should validate like dimensionless `""`
- objects without a `unit` property could validate like dimensionless by default
- validate `timedelta` against `"s"`

CagtayFabry:
## Changes
- add support for custom unit validation via `units_from_tagged` and `_get_instance_units`
- numeric base types (int, float, ndarray) validate against dimensionless
- `data_array` now directly supports validation with `wx_unit`
- `pd.Timedelta`, `pd.TimedeltaIndex` and timedelta-like `weldx.Time` will validate against `wx_unit: "s"`

## Related Issues

Closes #95 

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -75,7 +75,7 @@ ASDF
 
 -  update the default sorting order of ``select_tag`` for ``WeldxConverter`` [:pull:`733`]
 
--  add custom validation behavior to `wx_unit` [:pull:`739`]
+-  add custom validation behavior to ``wx_unit`` [:pull:`739`]
 
 deprecations
 ============
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -75,6 +75,8 @@ ASDF
 
 -  update the default sorting order of ``select_tag`` for ``WeldxConverter`` [:pull:`733`]
 
+-  add custom validation behavior to `wx_unit` [:pull:`739`]
+
 deprecations
 ============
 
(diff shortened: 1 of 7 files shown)


CagtayFabry:
We still have to think about the behavior when not unit information can be found or the property is missing as discussed in https://github.com/BAMWelDX/weldx/issues/184

CagtayFabry:
This should be ready to go now, feel free to merge if you agree @vhirtham 🚀

marscher:
Thanks for this improvement. Would it be possible to also accept "" instead of "<space>" as an empty unit definition?

vhirtham (approved):

CagtayFabry:
> Thanks for this improvement. Would it be possible to also accept `""` instead of`"<space>"`as an empty unit definition?

I'm not sure what happens with `""` actually, maybe it works? (I guess it depends on what `YAML` parses from `""` and what pint unit get's build from that)
Something like "dimensionless" should also work

maybe you can give it a try here
https://github.com/BAMWelDX/weldx/blob/f64658d8eff82da7531679c20cffe00e55b83151/weldx/schemas/weldx.bam.de/weldx/debug/test_unit_validator-0.1.0.yaml#L59-L61

PR #740, opened 2022-04-20
cache conda env in CI runs

CagtayFabry:
## Changes
cache conda env in CI runs

Apparently the micromamba action now has some caching functions built in https://github.com/mamba-org/provision-with-micromamba#example-with-download-caching

### DIFF
--- a/.github/workflows/build_pkg.yml
+++ b/.github/workflows/build_pkg.yml
@@ -38,7 +38,7 @@ jobs:
         uses: mamba-org/provision-with-micromamba@main
         with:
           environment-file: ./devtools/conda.recipe/build_env.yml
-          cache-env: true
+          cache-downloads: true
 
       - name: activate build env
         shell: bash -l {0}
--- a/.github/workflows/build_pkg.yml
+++ b/.github/workflows/build_pkg.yml
@@ -34,20 +34,11 @@ jobs:
         with:
           fetch-depth: 0 # Fetch all history for all tags and branches
 
-      - name: Cache conda
-        uses: actions/cache@v3
-        env:
-          # Increase this value to reset cache if ./devtools/conda.recipe/build_env.yml has not changed
-          CACHE_NUMBER: 2
-        with:
-          path: ~/micromamba/pkgs
-          key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('./devtools/conda.recipe/build_env.yml') }}
-
       - name: Setup Conda Environment
         uses: mamba-org/provision-with-micromamba@v11
         with:
           environment-file: ./devtools/conda.recipe/build_env.yml
-          micromamba-version: 0.16.0
+          cache-env: true
 
       - name: activate build env
         shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


CagtayFabry:
The cache seems to work but it doesn't pick up the pip 'editable' install (which makes sense I guess)

https://github.com/BAMWelDX/weldx/runs/6109402367?check_suite_focus=true#step:3:63

doc/rtd_environment.yml
```diff
@@ -31,4 +31,4 @@ dependencies:
   # pip packages
   - pip
   - pip:
-    - -e ../
+    - ../
```
CagtayFabry, line 34: this works in the [GH workflow](https://readthedocs.org/projects/weldx/builds/16706275/) and on [RTD](https://readthedocs.org/projects/weldx/builds/16706275/) 
I don't think it is necessary to keep the `-e` here, what do you think @marscher ?

CagtayFabry (commented):
the caching seems to work now, it might be a good alternative to using the GH cache actions manually

https://github.com/BAMWelDX/weldx/runs/6109667686?check_suite_focus=true#step:3:63

vhirtham (approved):

.github/workflows/build_pkg.yml
```diff
@@ -34,20 +34,11 @@ jobs:
...
-          path: ~/micromamba/pkgs
-          key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('./devtools/conda.recipe/build_env.yml') }}
-
       - name: Setup Conda Environment
-        uses: mamba-org/provision-with-micromamba@v11
+        uses: mamba-org/provision-with-micromamba@main
         with:
           environment-file: ./devtools/conda.recipe/build_env.yml
-          micromamba-version: 0.16.0
+          cache-downloads: true
```
CagtayFabry, line 41: not using `cache-env` here because the whole build environment and artifacts end up in the environment bloating it up to >1GB

marscher: Caching the archive downloads makes total sense. So the cache size is limited to 1GB currently?

doc/rtd_environment.yml
marscher: If there is anything which needs to refer to to source code directory via the Python module, this would break with this change. Apparently there is no such reference right now.
Actually I would have thought, that at least Sphinx would refer to the source code directory via relative file links. But since the pip installed source tree is equivalent, nothing should break.

marscher (approved):

marscher:
What a great speed up. This is really nice.

.github/workflows/build_pkg.yml
CagtayFabry: no I don't think there is a limit on the cache size but retrieving the huge environment cache was slower than just creating the build environment

marscher:
These caching periods make a lot of sense actually. I like you suggestion with a full work week. There could be also one uncached nightly build for master, to get notified early when things break. This would allow for longer cache periods for usual builds.

PR #742, opened 2022-04-26
Update ci

CagtayFabry:
Since the micromamba caching seems to work pretty well I wanted to try it out with our pytest jobs as well.
I wrote a small GH action [pydeps2env](https://github.com/CagtayFabry/pydeps2env) that creates a conda `environment.yml` from the `setup.cfg` so we don't have to keep the dependencies in two files.

The speedups with `ubuntu` vs pip installs (incl. cache) look really good, the whole env setup now only takes 26s ([was 1 minute before](https://github.com/BAMWelDX/weldx/actions/runs/2218372175))

`macos` is down to 51 s from 150 s for caching + install (but the cache was slightly outdated)

## Changes
- use mamba + [pydeps2env](https://github.com/CagtayFabry/pydeps2env) in pytest jobs
- use default shell command in conda jobs

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,20 +30,24 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # Fetch all history for all tags and branches
-      - uses: actions/setup-python@v3
-        with:
-          python-version: ${{ matrix.py }}
 
-      - uses: actions/cache@v3
+      - uses: CagtayFabry/pydeps2env@main
         with:
-          path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('setup.cfg') }}
-          restore-keys: |
-            ${{ runner.os }}-pip-
+          channels: 'conda-forge defaults'
+          extras: 'test'
+          setup_requires: 'include'
+
+      - uses: mamba-org/provision-with-micromamba@main
+        with:
+          environment-file: ./environment.yml
+          environment-name: demo
+          cache-env: true
+          extra-specs: |
+            python=${{ matrix.py }}
+            wheel
 
       - name: pip installs
         run: |
-          pip install wheel
           pip install -e .[test]
 
       - name: run pytest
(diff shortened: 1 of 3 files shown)


.github/workflows/pytest.yml
```diff
@@ -86,44 +99,43 @@ jobs:
...
         if: startsWith(runner.os, 'Windows')
         run: |
-          if not exist %userprofile%\.matplotlib\ ( mkdir %userprofile%\.matplotlib\ )
-          echo backend: Agg > %userprofile%\.matplotlib\matplotlibrc
+          mkdir $env:USERPROFILE\.matplotlib\
+          echo backend: Agg > $env:USERPROFILE\.matplotlib\matplotlibrc
           python -c "import matplotlib as m; print(m.matplotlib_fname())"
           python -c "import matplotlib as m; print(m.rc_params()['backend'])"
-        shell: cmd
+        shell: pwsh
```
CagtayFabry, line 138: unfortunately I have no idea how to get this running with `shell: bash`

the `cmd` doesn't know about the conda environment so it doesn't work...

CagtayFabry: I have removed the debug code lines so no python/conda is required, it works now

marscher:
This looks very good to me. I wonder why the status of Pytest Python 3.8 (win, osx) is not being reported). In GH Actions interface there are no jobs for this configuration.

marscher (approved):
Eventually its a caching problem with GHA. 

Python 3.8 (win, osx) is in the matrix, but gets overriden by the compat section later which then uses 3.9. So it seems intended, right?

CagtayFabry:
> Eventually its a caching problem with GHA.
> 
> Python 3.8 (win, osx) is in the matrix, but gets overriden by the compat section later which then uses 3.9. So it seems intended, right?

ah I see it now.
I change the mac and win runs to python 3.9 (the default in runner images) but the PR requirements are still looking for a job with 3.8 ('Required')
I will change the branch protection rules to look for 3.9

PR #743, opened 2022-04-27
Update `SpatialData` schema

vhirtham:
## Changes

Update the schema of the `SpatialData`  to support multi-dimensional numpy arrays.

### DIFF
--- a/weldx/schemas/weldx.bam.de/weldx/core/geometry/spatial_data-0.1.1.yaml
+++ b/weldx/schemas/weldx.bam.de/weldx/core/geometry/spatial_data-0.1.1.yaml
@@ -67,7 +67,7 @@ properties:
           datatype:
             type: string
             enum: [float32, float64]
-        wx_shape: [(~), 3]
+        wx_shape: [..., 3]
       - tag: "asdf://weldx.bam.de/weldx/tags/core/data_array-0.1.*"
         wx_shape: [..., 3]
   attributes:


CagtayFabry:
Thinking about this, I am not quite sure if we really want/need this? (or in what situation)
If there are more than 2 dimensions how would those be interpreted? (time? profiles?)

For 2 dimensions it is pretty obvious (simple pointcloud), for more dimensions maybe forcing an explicit description ob the dimensions by wrapping in xarray is safer?

Can you give an example where it is useful to have numpy here instead of xarray? @vhirtham

vhirtham:
This is currently needed for the Scan data. I made this change locally to be able to store the whole experimental data in a WelDX File. I was a bit surprised too, that it is still a NumPy array. I think the best way here is to make a reminder PR and merge this one for now, so that the others can already test the new utility class. Once I am done with that, I check why the `SpatialData` still uses a NumPy array and fix it and remove NumPy from the schema. Since the WelDX Files can be easily recreated from the experimental data, it shouldn't be an issue if some files get invalidated by this.

CagtayFabry:
the current asdf converter stores the values as `ndarray` if (and only if) there are no more than two dimensions which is consistent with the schema.

https://github.com/BAMWelDX/weldx/blob/63d1ce67bc4082a384574d3e1b9349170fbe8e00/weldx/tags/core/geometry/spatial_data.py#L16-L21

I will close this for now since I don't see that we need to change anything here.

PR #753, opened 2022-05-10
Install dependencies for MyPy analysis

### LINKED ISSUE #752: mypy needs an install of weldx in order to perform (deep) checking
@vhirtham noticed, that one needs an install. This makes sense, since the type annotations of dependencies would be missing otherwise. If a dependency has no type stubs on PyPI, then mypy has no possibilty to figure out concrete types.

So we should change the mypy action such, that it installs weldx

marscher:
## Changes

Install a full environment to make sure, MyPy checks all the input types.

## Related Issues

Closes #752

Note that this will introduce like 80 type errors we collected since last year August.


vhirtham:
Okay, if I didn't destroy anything else with my last commit, this should be ready now.

weldx/core.py
```diff
@@ -853,7 +855,7 @@ class GenericSeries:
     represented by a variable or be part of one of the expressions parameters.
     """
 
-    _required_dimension_units: dict[str, pint.Unit] = {}
+    _required_dimension_units: dict[str, Union[str, pint.Unit]] = {}
```
CagtayFabry, line 858: I see a lot of changes like this one from `pint.Unit` to `Union[str, pint.Unit]` for internal attributes or functions. Is there a specific reason for using `str` here (and in the other internal classes)?

Internally everything should run only with `pint.Unit`, only in user facing functions `str` makes sense in my eyes.

CagtayFabry (changes requested):
thank you both for going through all of these

weldx/core.py
vhirtham: I have to double-check this. My first priority was to satisfy mypy. IIRC this is correct since `_required_dimension_units` is just to set some constraints to the `GenericSeries` and we can do this by using `pint.Unit` or a unit string, which is easier to read. This is passed along between multiple functions. At some point when it is necessary, strings get converted to units, but this might happen further down the road.

weldx/core.py
```diff
@@ -1020,7 +1025,9 @@ def __init__(
         elif isinstance(obj, (MathematicalExpression, str, sympy.Expr)):
             if dims is not None and not isinstance(dims, dict):
                 raise ValueError(f"Argument 'dims' must be dict, not {dims}")
-            self._init_expression(obj, dims, parameters, units)
+            self._init_expression(
+                obj, dims, parameters, {k: U_(v) for k, v in units.items()}  # catch str
+            )
```
CagtayFabry, line 1030: I couldn't manage do simply do 
```python
units = {k: U_(v) for k, v in units.items()}
```
outside of the function call, but at least this does the trick @vhirtham

CagtayFabry (approved):
thank you again, this looks good to me 🚀 

please add a brief changelog entry since quite some code is affected

vhirtham:
What do we do about networkx? Pin it in another branch, merge it into master, then update this one and merge it too?

marscher:
I think we should add a Github actions cache for the mypy sqlite cache file, this would speed up the mypy action greatly (like total runtime 10s)

CagtayFabry:
> I think we should add a Github actions cache for the mypy sqlite cache file, this would speed up the mypy action greatly (like total runtime 10s)

works like a charm, nice 🚀

PR #755, opened 2022-05-13
remove old visualization modules

CagtayFabry:
## Changes
- remove the legacy `colors` and `types` modules

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -35,6 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Fetch all history for all tags and branches
 
       - name: Setup Conda Environment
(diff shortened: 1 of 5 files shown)


CagtayFabry:
the code works with the current `weldx-widgets@main`

https://github.com/BAMWelDX/weldx/runs/6420368539?check_suite_focus=true

I guess we can just merge this and make a new release for `weldx-widgets` @marscher ?

marscher:
Yes, if the tests pass for the git version, we should be fine for a released version as well.

.github/workflows/docs.yml
```diff
@@ -35,6 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Fetch all history for all tags and branches
```
CagtayFabry, line 39: we need this if we install other packages that depend on weldx to get the correct version number for the local install

PR #759, opened 2022-05-17
[wxfile] raise KeyError for non-existent keys.

### LINKED ISSUE #757: `WeldxFile` doesn't raise `KeyError` on missing key
When trying to access a non existing key from `WeldxFile`, `None` is returned instead of raising a  `KeyError`.
```python
from weldx import WeldxFile

with WeldxFile(tree={"key":"value"}, mode="rw") as wxf:
    print(wxf["key"])
    print(wxf["anything"])
# value
# None
```

Is this the intended behavior @marscher ?

marscher:
## Changes

WeldxFile now raises a KeyError, if a given key could not be accessed. Previously only a warning was emitted and the change was ignored. As this leads to confusion, e.g. None being returned, we now trigger the exception.

## Related Issues

Closes #757

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests


weldx/tests/asdf_tests/test_asdf_util.py
```diff
@@ -186,3 +188,46 @@ def test_get_highest_tag_version():
...
+class TestProtectedView(unittest.TestCase):
+    def setUp(self):
+        data = dict(foo="blub", bar=42)
+        self.protected_key = "bar"
+        self.view = _ProtectedViewDict(protected_keys=[self.protected_key], data=data)
+
+    def test_protected_keys_hidden(self):
+        assert self.protected_key not in self.view.keys()
+        assert self.protected_key not in self.view
+        assert (self.protected_key, "bar") not in self.view.items()
```
CagtayFabry, line 202:
```suggestion
        assert (self.protected_key, 42) not in self.view.items()
```
it probably doesn't matter but should this be the key/value item pair?

CagtayFabry (approved):
thank you for adding the test cases

weldx/tests/asdf_tests/test_asdf_util.py
marscher: you're absolutely right :+1:

PR #761, opened 2022-05-19
set networkx<=2.8.0

CagtayFabry:
## Changes
set networkx<=2.8.0 until CSM deepcopy is resolved
see: https://github.com/BAMWelDX/weldx/runs/6502339152?check_suite_focus=true#step:9:110

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,8 +37,9 @@ deprecations
 dependencies
 ============
 
--  Weldx now (optionally) requires weldx_widgets to visualize coordinate systems/manager [:pull:`749`].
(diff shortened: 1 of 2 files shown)


marscher (approved):

marscher:
https://github.com/networkx/networkx/issues/2793 might be related, but is from 2017

CagtayFabry:
looks like that could be related, yes

I don't really know what the cause is, if you do `deepcopy(csm)` it fails but `deepcopy(csm._graph)` seems to work

marscher:
that is indeed super weird.

marscher:
`G.edges` access also triggers the error, so it is completely unrelated to deepcopy, or pickle.

PR #763, opened 2022-05-21
Revert "set networkx<=2.8.0"

marscher:
Reverts BAMWelDX/weldx#761

xref: https://github.com/networkx/networkx/issues/5655

networkx-2.8.2 just got released, so we should pin like ..., !=2.8.1


CagtayFabry:
since we already released 0.6.1 with these dependencies I would prefer not to alter the changelog (and move it to the next release section)

PR #770, opened 2022-06-09
Fix length computation sympy

### LINKED ISSUE #725: Optimize length calculation of `DynamicBaseSegment`
The length calculation for expressions can actually fail in its current form. Here is an example:

~~~ python
from weldx import Q_, DynamicShapeSegment


class MySegment(DynamicShapeSegment):
    def __init__(self):
        f = "x * sin(s) + y * s"
        p = dict(x=Q_([1, 0, 0], "mm"), y=Q_([0, 1, 0], "mm"))
        super().__init__(f, parameters=p)


MySegment()
~~~

This crashes with the following error:

~~~
Traceback (most recent call last):
  File ".../sympy_bug.py", line 10, in <module>
    
  File ".../sympy_bug.py", line 8, in __init__
    super().__init__(f, parameters=p)
  File "...\weldx\geometry.py", line 144, in __init__
    self._update_internals()
  File "...\weldx\geometry.py", line 151, in _update_internals
    self._length = self.get_section_length(self._max_coord)
  File "...\weldx\geometry.py", line 213, in get_section_length
    length = self._length_expr.evaluate(max_coord=position).data
  File "...\weldx\core.py", line 277, in evaluate
    return self.function(**variables, **parameters)
  File "<lambdifygenerated-2>", line 4, in _lambdifygenerated
NameError: name 'Integral' is not defined
~~~

Change the expression from `x * sin(s) + y * s` to `x * sin(s) + y * cos(s)` and everything works fine. The problem here seems to be that the integration of the length formula gets really complicated and can't be solved by sympy using symbolic calculations. So the integral is kept as a "wrapper" (I guess) to solve it later in a different way. But this can't be turned into a python function (and we do that). So we might need to provide different methods for the length calculation. IIRC, sympy supports different alternatives, numeric integration amongst them. The last resort would be to catch this error and inform the user, that the length couldn't be determined automatically and he has to provide a custom function for that.

marscher:
## Changes

If we use the `Integrate` function in SymPy lambdify, we need to use "scipy" as method, "numpy" is not sufficient.
xref: https://github.com/sympy/sympy/pull/20134

## Related Issues

Closes #725

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] ~updated doc/~
- [x] ~update example/tutorial notebooks~
- [x] ~update manifest file~


marscher:
note that the deepsource issues are unrelated to my changes :P

marscher:
@vhirtham I've noticed, that using scipy as backend now triggers deprecation warnings like this: 

`DeprecationWarning: scipy.sin is deprecated and will be removed in SciPy 2.0.0, use numpy.sin instead`

However we need to use scipy to make "Integrate" use the right backend. Do you have a suggestion how to proceed?

marscher:
xref: https://github.com/sympy/sympy/issues/20294

vhirtham:
> note that the deepsource issues are unrelated to my changes :P

I silenced it for the future ;)

vhirtham (approved):
LGTM. Curious if this fixes some of the old computations issues I had.

marscher:
Let me know if that helped with your old issues.

CagtayFabry:
this looks great, nice find !

Issue #754, opened 2022-05-12
Setuptools setup.cfg? deprecation


marscher:
It looks like Setuptools discourages the usage of setup.cfg in favour of pyproject.toml.
I need to validate this, but we get the following deprecation warning during the invocation of setup.py 

```
/setuptools/config/__init__.py:28: SetuptoolsDeprecationWarning: As setuptools moves its configuration towards `pyproject.toml`,
`setuptools.config.read_configuration` became deprecated.

For the time being, you can use the `setuptools.config.setupcfg` module
to access a backward compatible API, but this module is provisional
and might be removed in the future.
```
xref: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

CagtayFabry:
good to know, thank you

I guess there is still some time to move over to `pyproject.toml` right?

If possible I would like to wait till Python 3.11 (which afaik should come with a toml parser by default) to move our toolchain over to `pyproject.toml`

marscher:
I think this transition will be carefully treated by the community, as lots of projects use setup.cfg to store their metadata. So we can take our time to make this change.

marscher:
fixed by #774

PR #774, opened 2022-06-15
Migrate from setup.cfg to pyproject.toml

### LINKED ISSUE #769: [setup.cfg] license_file parameter is deprecated, use license_files instead.
setuptools/config/setupcfg.py:459: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.

marscher:
## Changes

- All the metadata is now living in pyproject.toml
- setup.cfg is not needed anymore (just as a workaround to support editable/dev installs ATM).
- setup.py is not needed anymore
- Makes it mandatory to install weldx (source) with pip >= 20
- flake8 maintainer is currently refusing to rely upon pyproject.toml, as there is no std lib to parse (will available in Python 3.11). For this reason we move the cfg of flake8 to .flake8 (in order to leave setup.cfg) empty).

## TODOS
- [x] ~flake8 needs additional package to  https://stackoverflow.com/questions/64482562/specify-per-file-ignores-with-pyproject-toml-and-flake8~ revert settings to setup.cfg (until they decide to support it).
- [x] ~pytest-cov/coverage do not produce coverage data (combination step fails).~
- [x] ~ASDF entry points are not working (yet). Entrypoints are registered successfully. fixed pytest cfg section name.~
- [x] conda-build renders version to "None", seems like dynamically generated versions cannot be used at this stage. There used to be a similar problem for setup.py derived metadata back at the day. Raise issue at conda-build.

## Resources
- https://setuptools.pypa.io/en/latest/userguide/ shows syntax for pyproject.toml

## Related Issues

Closes #769, #754

## Checks

- [x] updated CHANGELOG.rst


marscher:
https://github.com/conda/conda-build/pull/4480 is implemented, but not yet released (it seems).

CagtayFabry:
> [conda/conda-build#4480](https://github.com/conda/conda-build/pull/4480) is implemented, but not yet released (it seems).

that feature would be a godsend.. should we wait for that release to merge?
I like the implementation with `pyproject.toml` here 👍

marscher:
The conda-build containing the desired feature has been released, however it does not render the version correctly and is therefore not yet usable.

CagtayFabry:
I think conda build 3.22 finally has what we need here https://github.com/conda/conda-build/releases/tag/3.22.0 🚀

CagtayFabry (approved):
I have tested the editable install without `setup.cfg` using `pip>=21.3` and `setuptools==65` as suggested here https://stackoverflow.com/a/69711730/11242411 and it works @marscher 

but I think it is good to keep the `setup.cfg` stub for now for compatibility

marscher:
Thanks for reviewing this. What do you mean by "keeping the setup.cfg stub"? I mean by now it would only store redundant information, so I do not understand, what the benefit would be.

PR #775, opened 2022-06-21
Update LCS tutorial

vhirtham:
## Changes

Update the tutorial about the LCS.

## Checks

- [x] updated CHANGELOG.rst
- [x] ~~updated tests~~
- [x] updated doc/


vhirtham:
> @marscher Do you remember how to tell nbconvert that it should only print the output but not the code of a cell?

Okay, should be [this](https://nbsphinx.readthedocs.io/en/0.2.1/hidden-cells.html)

vhirtham:
> > @marscher Do you remember how to tell nbconvert that it should only print the output but not the code of a cell?
> 
> Okay, should be [this](https://nbsphinx.readthedocs.io/en/0.2.1/hidden-cells.html)

Doesn't seem to work cause it also hides the output. I can hide the code cell in jupyter directly, but in the sphinx-doc, the cell is still rendered. Any idea?

PR #777, opened 2022-06-30
Add a test for quality standards and update the quality standard tutorial

vhirtham:
## Changes

Since version `0.5.0` the quality standard tutorial was outdated. The presented standard did not override the schema and the subsequent code didn't produce the exceptions that were mentioned in the text. Due to the fact that it wasn't checked if the exception was actually raised, this wasn't noticed. I added a new test for quality standards and adjusted the tutorial too so that this can't happen again.

## Checks

- [x] updated CHANGELOG.rst
- [x] updated tests
- [x] updated doc/


vhirtham:
Totally forgot about the issue that quality standards affect the whole pytest run. That's why the tests are currently failing. The question is how we want to advance. We can either remove the test and rely on the modified notebook to catch future bugs, or we can exclude the test from the regular pytest run and add a separate run just for this test. What do you think?

CagtayFabry:
> Totally forgot about the issue that quality standards affect the whole pytest run. That's why the tests are currently failing. The question is how we want to advance. We can either remove the test and rely on the modified notebook to catch future bugs, or we can exclude the test from the regular pytest run and add a separate run just for this test. What do you think?

Ah yes, that is tricky when the whole environment is affected...

I would suggest the following:
Exclude the run from regular pytest run (with the setting in `setup.cfg`) and run it only after the last "regular" test here (after running `test with WeldxFile`)
https://github.com/BAMWelDX/weldx/blob/2af3da3a9745343a02ff09dd1c2ff05489f263b0/.github/workflows/pytest.yml#L78-L85

I hope that works

CagtayFabry (approved):

setup.cfg
```diff
@@ -100,7 +100,11 @@ match = (?!__)(?!_version)(?!conftest).*\.py
 match_dir = ^(?!(tests|tags|asdf)).*
 
 [tool:pytest]
-addopts = --tb=short --color=yes -rsw --cov=weldx --cov-report=term-missing:skip-covered --doctest-modules
+addopts =
+    --tb=short --color=yes -rsw
+    --cov=weldx --cov-report=term-missing:skip-covered
+    --doctest-modules
+    --ignore=./weldx/tests/asdf_tests/test_quality_standards.py
```
vhirtham, line 107:
```suggestion
addopts = --tb=short --color=yes -rsw --cov=weldx --cov-report=term-missing:skip-covered --doctest-modules
```

PR #783, opened 2022-07-11
Fix build failure

vhirtham:
## Changes

Fix the current [build failure](https://github.com/BAMWelDX/weldx/runs/7280019473?check_suite_focus=true)

### DIFF
--- a/.github/workflows/build_pkg.yml
+++ b/.github/workflows/build_pkg.yml
@@ -51,9 +51,6 @@ jobs:
 
       - name: conda list
         run: conda list
-        
-      - name: conda config
-        run: conda config --show
 
       - name: Build from local conda recipe
         run: conda mambabuild -c conda-forge .
--- a/.github/workflows/build_pkg.yml
+++ b/.github/workflows/build_pkg.yml
@@ -51,9 +51,12 @@ jobs:
 
       - name: conda list
         run: conda list
+        
+      - name: conda config
+        run: conda config --show
 
       - name: Build from local conda recipe
-        run: conda mambabuild .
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -37,7 +37,7 @@ setup_requires =
     setuptools >=38.3.0
     setuptools_scm
 install_requires =
-    numpy >=1.20
+    numpy ==1.21
```
vhirtham, line 40:
```suggestion
numpy >=1.20
```

setup.cfg
```diff
@@ -37,7 +37,7 @@ setup_requires =
     setuptools >=38.3.0
     setuptools_scm
 install_requires =
-    numpy >=1.20
+numpy >=1.20
```
vhirtham, line 40:
```suggestion
    numpy >=1.20
```

.github/workflows/build_pkg.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         run: conda list
 
       - name: Build from local conda recipe
-        run: conda mambabuild .
+        run: conda mambabuild -c conda-forge .
```
CagtayFabry, line 56: I am not sure why the previously available channels are no longer detected here but this seems to fix the build issues

CagtayFabry (approved):

.github/workflows/build_pkg.yml
vhirtham: Interesting. I thought it was covered by the recipe that contains conda-forge. Good that you found a solution. Didn't know what to try else.

PR #791, opened 2022-07-25
Update `TimeSeries` handling of absolute times

### LINKED ISSUE #471: `TimeSeries` todos
The `TimeSeries` basically supports absolute times since PR #440.  But this support is not really tested and there are probably still some execution paths that might fail (see todo in `_initialize_discrete`). Additionally, the reference time should be stored inside the internal xarray data structure as suggested by @CagtayFabry in this PR #468.

List of todos:
- Check code for missing consideration of absolute times
- Add missing test cases for absolute times
- Store reference time directly inside of internal xarray structure 

I am not sure about this, but we might also consider moving the `TimeSeries` to the time package

vhirtham:
## Changes

Addresses the issues mentioned in #471

## Related Issues

Closes #471

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/weldx/core.py
+++ b/weldx/core.py
@@ -16,7 +16,7 @@ from bidict import bidict
 
 import weldx.util as ut
 from weldx.constants import Q_, U_, UNITS_KEY
-from weldx.time import Time, TimeDependent, types_time_like
+from weldx.time import Time, TimeDependent, types_time_like, types_timestamp_like
 from weldx.util import check_matplotlib_available
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -301,7 +301,7 @@ class TimeSeries(TimeDependent):
         data: Union[pint.Quantity, MathematicalExpression],
         time: types_time_like = None,
         interpolation: str = None,
-        reference_time: pd.Timestamp = None,
+        reference_time: types_timestamp_like = None,
     ):
         """Construct a TimSeries.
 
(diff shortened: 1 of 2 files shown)


weldx/core.py
```diff
@@ -480,9 +484,14 @@ def _init_expression(self, data):
 
     def _interp_time_discrete(self, time: Time) -> xr.DataArray:
         """Interpolate the time series if its data is composed of discrete values."""
+        data = self._data
+        if self.time is None and time.is_absolute:
+            # type: ignore[union-attr]
+            data = data.weldx.reset_reference_time(time.reference_time)
```
vhirtham, line 490: This is something we should probably discuss. It fixes the following problem:

We replaced

~~~ python
        return ut.xr_interp_like(
            self._data,
            {"time": time.as_data_array()},
            method=self.interpolation,
            assume_sorted=False,
            broadcast_missing=False,
        )
~~~

with

~~~ python
return ut.xr_interp_like(
            data,
            time.as_data_array(),
            method=self.interpolation,
            assume_sorted=False,
            broadcast_missing=False,
        )
~~~

to get absolute times working correctly when using `xr_interp_like`. However, this causes the test case `(ts_const, time_mul + pd.Timestamp("2020"), [1, 1, 1, 1, 1, 1, 1, 1], "m")` of `test_interp_time` to fail (currently line 450 in `test_core.py`). The test just checks if the value of a constant `TimeSeries` that gets interpolated with an array of times is broadcasted correctly. Since the value is constant, reference times shouldn't matter, but after the change, they suddenly did. So I caught this special case and just added a reference time to a copy of the data prior to interpolation. The question is if we want to keep it that way, should it be handled differently or is that something that `xr_interp_like` should handle internally

PR #806, opened 2022-09-26
Update rtd_environment.yml

CagtayFabry:
## Changes
use older nbformat due to `MissingIDFieldWarning`

## Checks

- [x] updated doc/


marscher:
I think we would just need to fix our notebooks, e.g. each cell has to have an ID.

CagtayFabry:
> I think we would just need to fix our notebooks, e.g. each cell has to have an ID.

that would probably be the right solution 😁 
I am not sure if the IDs get removed by some kind of script/hook atm though

If this doesn't work as a quick fix I am fine to leave it for now until we fix the notebooks

PR #809, opened 2022-09-30
Conda recipe fix deps

marscher:
## Changes
Since conda does not know the concept of optional dependencies, so we install
all optional features from pyproject.toml to ensure everything is
working.
The test dependencies are also obtained from pyproject.toml


CagtayFabry:
great solution, thank you!

CagtayFabry (approved):

PR #820, opened 2022-10-27
Fix pint intersphinx

marscher:
## Changes

Use latest in intersphinx mapping to resolve issues with documentation building.

For some reason `pint` cannot be resolved in the changelog.rst anymore. So I replaced these with ``pint``.
Additionally the (ugly) hacks for `__module__` of weldx.Q_ and U_ have gone now. Since these hacks were only required for correct Sphinx building, no user code should be affected.


marscher:
now we need to wait for 0.20.1 to arrive on CondaForge.

weldx/tests/asdf_tests/test_asdf_groove.py
```diff
@@ -1,19 +1,29 @@
...
+def _close_plot():
+    try:
+        from matplotlib import pylab
+    except ImportError:
+        yield
+    else:
+        try:
+            yield
+        finally:
+            pylab.close()
```
CagtayFabry, line 25: can you give me a heads up on what this does? @marscher

marscher: the test runs did not close their figures, leading to a warning of matplotlib. This context manager closes the produced figures by every test run.

PR #821, opened 2022-10-28
Split up core in smaller sub modules

marscher:
## Changes

Split up core in smaller sub modules, e.g. generic_series and mathematical_expression and so forth.


marscher:
In my opinion the code is better maintainable, if modules are kept smaller.

The complains of static analysis are not new, but now show up as being newly introduced.
Regarding the RTD issues, I'm not able to reproduce these locally. But I think they are also not related to the changes made here.

CagtayFabry (approved):
I agree, this should be easier to maintain and keep clean 👍 
thanks for doing this

PR #823, opened 2022-11-05
build RTD with sphinx 5.1

CagtayFabry:
## Changes

_Describe changes in this PR_

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/doc/rtd_environment.yml
+++ b/doc/rtd_environment.yml
@@ -18,7 +18,7 @@ dependencies:
   - ipywidgets<8
   - weldx_widgets >=0.2.1
   # documentation
-  - sphinx>=4.1.1
+  - sphinx=5.1
   - jinja2=3.0
   - nbsphinx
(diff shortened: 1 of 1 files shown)


CagtayFabry:
It appears the warnings where introduce by this change: https://github.com/sphinx-doc/sphinx/pull/10781

How ever, I don't see how we could restructure our documentation in this case (since we duplicate the docstring)

marscher:
I do not think that this pinning is mandatory, as the build on RTD now proceeds without any warnings (after merging the 0.6.2 branch).

CagtayFabry:
> 

yes I just wanted to run this as a test to see where the error came from
But I would rather stick with the changes in the 0.6.2 branch

Will close this now

PR #824, opened 2022-11-15
use Ruff linter pre-commit actions

marscher:
## Changes

- Use Ruff linter pre-commit actions
- Fixed linting issues.
- Use Ruff with NbQA.
- Removed (replaced with Ruff) pre-commit hooks:
   - isort
   - pydocstyle
   - flake8
   - pep585-upgrade

Note that I also removed flake8-deprecations, which only checked for a hard coded dictionary of super old deprecations inside the std lib.

## Checks

- [x] updated CHANGELOG.rst


CagtayFabry:
I think the config looks very clean now that nbqa-ruff is also available @marscher 👍 

Just some errors to look into with the latest versions

CagtayFabry:
we could also think about removing isort for the ruff implementation

marscher:
With ruff-0.0.208 the error output does not contain the originating file name. This renders the output useless.

CagtayFabry:
> With ruff-0.0.208 the error output does not contain the originating file name. This renders the output useless.

oh that sounds bad, I hope ruff will introduce an option to toggle this at least..?

marscher:
> oh that sounds bad, I hope ruff will introduce an option to toggle this at least..?

However, this only seems to happen on Github and there some kind of annotations api is used. So the files will be flagged in the "changes" section of the PR :+1:

marscher:
Honestly I do not know how to approach the notebook errors related to implicitly imported attributes. Xref: https://github.com/nbQA-dev/nbQA/issues/785

marscher:
> we could also think about removing isort for the ruff implementation

yes, we should do that. I do not know if Ruff respects the tool settings of isort inside pyproject.toml or refers to its own isort section.

CagtayFabry (approved):
wow great work digging through all of it @marscher 🙏
the linting config also looks super clean 👍

PR #826, opened 2022-11-28
Fix wxextension matching

### LINKED ISSUE #825: asdf-2.14.1 leads to a condition where weldx extension is found twice.
This makes `get_weldx_extension()` fail, as the weldx extension is meant to be found only once. Eventually we register our extension in a legacy manner.
Downgrading to asdf-2.13 circumvents this problem.

marscher:
## Changes

asdf-2.14.0+ returns two `Extension` instances matching the pattern in `get_weldx_extension()`. So we allow this now and add a version mismatch check.
Another minor change is to add caching to the function, as it is being called for some types, each time one of these types is being serialized. The cache hits, if the extension tuple matches (based on id(), I'd guess).

## Related Issues

Closes #825

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,8 +13,16 @@ readme = {file = "README.rst", content-type="text/x-rst"}
 license = {file = "LICENSE", name="BSD License"}
 keywords = [
     "welding",
-    "weldx",
-    "bam", # TODO: add more keywords here! think of them as in a journal article.
+    "measurement",
+    "experiment",
+    "coordinate transformations",
+    "time dependent coordinates",
+    "BAM",
+    "Bundesanstalt für Materialforschung- und Prüfung",
+    "scientific data format",
+    "asdf",
+    "advanced-scientific-data-format",
+    "iso-9692-1",
 ]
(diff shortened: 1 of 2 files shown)


CagtayFabry:
I am not sure if this is a bug on the `asdf` side or if something is wrong with the way we implemented our extension

It would be good to test the caching with the quality standard implementation to see if that has any effect

marscher:
Taking a look in the debugger, other extensions, including asdf standard extensions have this splitted pattern of one extension carrying the tags and another containing the converters. So I doubt that you did something wrong adopting the new pattern (a manifest extension, right?)

marscher:
If you are afraid, that caching breaks something, I can also just remove the implementation. Probably it doesn't make any difference.
How would I test against the quality standard? I thought the default is activated on import time.

marscher:
I expect that asdf changes the entrypoint listing implementation, so we do not have duplicates in this list.

PR #829, opened 2022-12-15
Revert "pin asdf<2.14"

marscher:
Reverts BAMWelDX/weldx#828

Wait until asdf-2.14.3 and re-run failed test jobs

PR #834, opened 2023-01-03
Unpin ASDF

### LINKED ISSUE #825: asdf-2.14.1 leads to a condition where weldx extension is found twice.

WilliamJamieson:
## Changes

This PR unpins asdf. In asdf version 2.14.3 we specifically fixed the bug introduced in asdf 2.14.0 which lead to #825. This means weldx should be able to unpin asdf now.

Note that I specifically excluded the asdf versions: 2.14.0, 2.14.1, and 2.14.2 as they all experience the bug reported in #825.

## Related Issues

Closes #825

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ dependencies
 ============
 
 -  pin ``asdf<2.14`` due to changes in the extension mechanism [:pull:`828`].
+-  Unpin ``asdf`` due to fix in ``asdf 2.14.3`` release [:pull:`834`].
 
 changes
 =======
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
 requires-python = ">=3.8,<3.11"
 dependencies = [
     "numpy >=1.20",
-    "asdf >=2.8.2,<2.14",
+    "asdf >=2.8.2,!=2.14.0,!=2.14.1,!=2.14.2",
     "pandas >=1.0",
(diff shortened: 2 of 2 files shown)


CagtayFabry:
Thank you @WilliamJamieson 😊

We will probably consolidate this with #829 and discard older asdf versions for the next release to clean up some APIs as well (we are just slightly behind on the PRs over the holiday season 😉)

WilliamJamieson:
> Thank you @WilliamJamieson 😊
> 
> We will probably consolidate this with #829 and discard older asdf versions for the next release to clean up some APIs as well (we are just slightly behind on the PRs over the holiday season 😉)

No problem, I noticed in the ASDF downstream tests that weldx was pinning ASDF, which made pip complain. However, everything seemed to be passing.

I didn't even check to see if you had a PR unpinning ASDF, so feel free to close this PR if you want.

marscher:
I prefer this kind of pinning, where the incompatible versions are listed explicitly. Thanks a lot!

CagtayFabry:
> I prefer this kind of pinning, where the incompatible versions are listed explicitly. Thanks a lot!

I agree, however we don't currently run any compatibility tests for older asdf versions
If we introduce those I would feel a bit more at ease keeping the older versions

PR #835, opened 2023-01-05
[GithubActions] Test old ASDF minor versions

marscher:
## Changes

Run asdf tests against all (supported) minor versions of ASDF.

## Related Issues
Related #834, #829


CagtayFabry:
the `pip` version takes priority over the conda install version I assume?

We could also add an `exclude` option to `pydeps2env` to do these kind of things

PR #837, opened 2023-01-06
Unpin maximum Python version

### LINKED ISSUE #836: Python 3.11 incompatibilities
- ```ValueError: mutable default <class 'numpy.ndarray'> for field prop1 is not allowed: use default_factory```

marscher:
## Changes

We allow future versions again. A very detailed discussion why pinning maximum versions can be found here: https://iscinumpy.dev/post/bound-version-constraints/

Also fixed some dataclasses in asdf.tags.debug 

TLDR:
Never cap Python, it is fundamentally broken at the moment. Also, even packing capping has negative consequences that can produce unexpected solves.

## Related Issues

Closes #836 

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/.github/workflows/build_pkg.yml
+++ b/.github/workflows/build_pkg.yml
@@ -81,7 +81,7 @@ jobs:
           CACHE_NUMBER: 0
         with:
           path: ~/.cache/pip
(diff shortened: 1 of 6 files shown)


marscher:
Now we got the first evidence that pinning Python for libraries is a bad idea. Weldx-widgets depends on weldx, but this cannot be installed for Python 3.11. So it is an expected failure.

PR #842, opened 2023-01-24
update ruff 233, nbqa 1.6.1

CagtayFabry:
## Changes

update ruff to 230 which has `--force-exclude` set by default

Not sure if this propagates to NBQA


marscher:
nbqa has its own ruff command line.

CagtayFabry:
> nbqa has its own ruff command line.

Are the NBAa releases tied to a specific ruff release or is that "up to date"?

marscher:
As nbqa is just a wrapper for the underlying tools, I think the latest version of the demanded tool will be installed.

Issue #845, opened 2023-01-27
Install weldx in mypy gha


marscher:
In #753 I pretended that we really have an install of weldx for type checking via Mypy. But it only installs all requirements, not weldx itself.
That is of course useful to determine the involved types, if the packages do not provide type stubs.

This leads to a discrepancy in results comparing the output of Mypy Github actions and a local run (0 vs ~40 errors).

PR #847, opened 2023-01-27
Update mypy action

CagtayFabry:
## Changes

Update mypy GitHub action:
- install mypy via conda
- add local weldx install

see #845

## Checks

- [x] updated tests


CagtayFabry:
Is this what you had in mind with #845 @marscher ?

The local pip install doesn't seem to make a difference here

marscher:
> Is this what you had in mind with #845 @marscher ?

yes, it is. It is kinda strange, that with the same mypy version, configuration, and stub packages I get very different results.

marscher (approved):

PR #848, opened 2023-01-29
[WXFile] determine widget usage by runtime env in _ipython_display_

### LINKED ISSUE #846: [docs] WeldxFile output in notebooks renders to IPython.display.JSON in jupyterhub environment
This can only be shown in a Jupyter lab environment. So the selection of the output should be made aware, that it is contained in a Sphinx/non-interactive session in order to avoid this.

  
image

marscher:
Fixes #846

## Changes

In `_ipython_display_` we assumed, that it will only be called in a Jupyter lab environment. This ain't true - within a Sphinx environment it's also called. So we just let this state be determined automatically.

## Related Issues

Closes #846

## Checks

- [x] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   `TimeSeries` and `SpatialData` converters \[{pull}`831`\].
 - Use Ruff in pre-commit-action \[{pull}`824`\].
 - Use MyST-NB for documentation building \[{pull}`830`\].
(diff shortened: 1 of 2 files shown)


marscher:
Now the extended notebook about WXfile renders correctly (https://weldx--848.org.readthedocs.build/en/848/tutorials/weldxfile.html)

PR #849, opened 2023-02-01
ignore jupyter folders

CagtayFabry:
## Changes

ignore `.ipynb_checkpoints` folders in pytest discovery


marscher:
why would we have these in the root directory?

CagtayFabry:
> why would we have these in the root directory?

jupyter tends to create a lot of temporary files in those folders (e.g. when opening a .py file)
They don't pop up in the root dir, the exclude works in subfolders

PR #830, opened 2022-12-19
Sphinx use myst-nb

marscher:
## Changes
- Sphinx uses myst-nb to directly include Markdown.
- moved tutorials to doc/src/ and removed copying code.
- Converted all remaining documents to MD.
- Notebooks gets executed myst-nb

### TODOs
- add prologue to notebook processing.
- fix: skipping unknown output mime type: application/json [mystnb.unknown_mime_type] https://github.com/executablebooks/MyST-NB/issues/464
- fix: weldx/transformations/__init__.py:docstring of weldx.transformations.CoordinateSystemManager:1: WARNING: unknown document: 'tutorials/transformations_02_coordinate_system_manager'
- fix: weldx/transformations/__init__.py:docstring of weldx.transformations.LocalCoordinateSystem:1: WARNING: unknown document: 'tutorials/transformations_01_coordinate_systems'

## Related Issues

Closes #745 

## Checks

- [x] updated CHANGELOG.rst
- [ ] updated tests
- [x] updated doc/
- [ ] update example/tutorial notebooks
- [ ] update manifest file

PR #850, opened 2023-02-01
update manifest

CagtayFabry:
## Changes
 prepare v0.6.3. release

## Related Issues

Closes # (add issue numbers)

## Checks

- [ ] updated CHANGELOG.rst
- [ ] updated tests
- [ ] updated doc/
- [ ] update manifest file

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.233
+    rev: v0.0.238
     hooks:
       - id: ruff
         args:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Release Notes
 
-## 0.6.3 (unreleased)
+## 0.6.3 (02.02.2023)
+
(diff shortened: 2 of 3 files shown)


.github/workflows/pytest.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       - name: run pytest on notebooks (append coverage)
         if: matrix.py == '3.8'
         run: |
-          pytest -n 2 --dist loadfile --nbval --current-env --cov-append tutorials/
+          pytest -n 2 --dist loadfile --nbval --current-env --cov-append ./doc/src/tutorials/
```
marscher, line 92: good catch! Sorry I've missed that in #830

marscher:
Why did you need to bump the manifest/extension version? I remember I was struggeling with adding the media_file schema to manifest and had some chat with Volker to resolve it.

CagtayFabry:
> Why did you need to bump the manifest/extension version? I remember I was struggeling with adding the media_file schema to manifest and had some chat with Volker to resolve it.

The manifest (and weldx extension) with version `0.1.1` have already been released in `v0.6.2` (or earlier)

In those cases I prefer to create a new version to avoid any confusion and keep the release versions stable

.github/workflows/pytest.yml
CagtayFabry: I just wanted to get rid of the folder link (luckily it only broke pytest and no sphinx links :wink: )

PR #854, opened 2023-02-06
Fix is jlab session

### LINKED ISSUE #846: [docs] WeldxFile output in notebooks renders to IPython.display.JSON in jupyterhub environment

marscher:
## Changes
To detect a Jupyter lab server started by Jupyter Hub, we also inspect the parent processes environment.

## Related Issues

Closes #846 

## Checks

- [x] updated CHANGELOG.rst

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.6.4 (unreleased)
 
+### Changes
+
+- `WeldxFile` also shows its JSON widget header representation in an Jupyter hub based environment \[{pull}`853`\].
+
 ### ASDF
 
 - remove preliminary validators from new style ASDF extension for compatibility with upcoming `asdf` changes \[{pull}`853`\].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


marscher:
Damn, this then breaks the doc building again, because we have api keys set for the ipykernel communication...

CagtayFabry:
> Damn, this then breaks the doc building again, because we have api keys set for the ipykernel communication...

seems tricky..
we could also introduce some kind of flag to disable JSON output altogether for the sphinx build (or simply monkeypatch the behavior we want in the sphinx run)

marscher:
I was wrong (probably looking at an outdated local build of the docs). Here it looks fine: https://weldx--854.org.readthedocs.build/en/854/tutorials/weldxfile.html
I've also tested the patched version on our Jupyterhub server and it shows the JSON output there.

CagtayFabry:
> I was wrong (probably looking at an outdated local build of the docs). Here it looks fine: https://weldx--854.org.readthedocs.build/en/854/tutorials/weldxfile.html I've also tested the patched version on our Jupyterhub server and it shows the JSON output there.

yes it seems to work now, great 🚀 !

CagtayFabry (approved):

PR #861, opened 2023-03-27
Fix XArray LCS join

### LINKED ISSUE #810: xarray-2022.9.0 breaks weldx
- dimensions with just one element cannot be aligned any more with arrays having more dimensions. E.g. dims ("time", ) is not alignable to other dims ("time", "c", "foobar").
`ValueError: cannot align objects with join='exact' where index/labels/sizes are not equal along these coordinates (dimensions): 'time' ('time',)`

marscher:
## Changes

Some new version of XArray changed the merge logic in a way being incompatible with our demands in LCS. Setting the join method of the merge to "override" seems to work (at least the tests are passing).

## Related Issues

Properly fixes #810 

## Checks

- [x] updated CHANGELOG.rst


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "numpy >=1.20",
     "asdf >=2.8.2,!=2.14.0,!=2.14.1,!=2.14.2",
     "pandas >=1.0",
-    "xarray >=0.19,<2022.09.0",
+    "xarray >=2023.3.0",
```
marscher, line 41: Actually we can also think of finding the XArray version which introduced Dataset.merge(join="override"), but I found it easier to just request at least a recent version.

CagtayFabry:
from the description of `xr.merge(join="override")` I don't think this is what we actually want since non equal indexes seem to be overridden.
I am not sure if there ever was any test that would catch this

marscher:
At least we have some test cases which produce non equal indices. Otherwise it would not fail with join="equal", right?

CagtayFabry:
> At least we have some test cases which produce non equal indices. Otherwise it would not fail with join="equal", right?

I assume that's the case, but it would be good to check it properly
Did you test if this runs on the original broken xarray version `2022.9.0`?

marscher:
I just ran the test now and it works as expected.

CagtayFabry:
> I just ran the test now with xarray `2022.09.0` and it works as expected.

great, thank you
I will try to take a look into the PR later and we can just pin `>=2022.09.0`

marscher:
Python 3.11 currently fails with
```
   error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ python 3.11**  is requested and can be installed;
      ├─ weldx_widgets >=0.2  is installable and it requires
      │  └─ weldx [>=0.5 |>=0.6 ] with the potential options
      │     ├─ weldx [0.5.0|0.5.1|0.5.2] would require
      │     │  └─ python >=3.8,<3.10 , which conflicts with any installable versions previously reported;
      │     ├─ weldx [0.6.0|0.6.1|0.6.2] would require
      │     │  └─ python >=3.8,<3.11 , which conflicts with any installable versions previously reported;
      │     └─ weldx [0.6.3|0.6.4] would require
      │        └─ xarray >=0.19,<2022.09.0 , which can be installed;
      └─ xarray >=2022.9.0  is uninstallable because it conflicts with any installable versions previously reported.
```

So we would need to release another version to fix this for Python 3.11, but I do not think this is urgent.

CagtayFabry:
I think I caught the bug though I am not completely sure what caused the change in xarray behavior (I think improved attribute propagation in xarray functions)

I don't mind releasing 0.6.5 at this point, it's nice to be back on the xarray release cycle

marscher:
Brilliant, then let's give a shot!

CagtayFabry (approved):

PR #868, opened 2023-04-16
fix LCS time handling for xarray>=2023.4.0

CagtayFabry:
Some time interpolation code seems to break with the new xarray release (seemingly unrelated of pandas major version 1 or 2)

## Changes

- add explicit handling of reference time for coordinates/orientation in `LocalCoordinateSystem` code
- set `keep_attrs=True` when applying `xarray.apply_ufunc`


marscher:
I'd be good to have a stack trace or failed report inside an issue, since the logs gets deleted after some time.

CagtayFabry:
> Thanks! It'd be good to have a stack trace or failed report inside an issue, since the logs gets deleted after some time.

The thing with these failures it the error itself is not that important, more the environment and installed packages. But we can paste those here

marscher:
It is also strange, that the conda build also installs pandas 1.5.3, but the tests succeed.

CagtayFabry:
> It is also strange, that the conda build also installs pandas 1.5.3, but the tests succeed.

The problem seems to be related to xarray 2023.4.0, which is available on PyPi but not released on conda-forge yet

(maybe the error would also arise with pandas 2 but we cant test that against an older working xarray version)

CagtayFabry:
I have tested this manually with old `xarray=2022.9.0` and `pandas=1.3.5` versions

The new code is more explicit in achieving the desired behavior. However much of the time related code should be reworked soon to reduce the compelxity

weldx/util/xarray.py
```diff
@@ -617,7 +618,7 @@ def xr_3d_vector(
         coords["time"] = time  # type: ignore[assignment]
 
     if "time" in coords:
-        coords["time"] = Time(coords["time"]).index
+        coords["time"] = Time(coords["time"]).as_data_array()
```
CagtayFabry, line 621: This might be slower than sticking to pandas indexes but more stable

weldx/transformations/local_cs.py
```diff
@@ -360,6 +364,7 @@ def _check_and_normalize_orientation(orientation: xr.DataArray) -> xr.DataArray:
             orientation,
             input_core_dims=[["c"]],
             output_core_dims=[["c"]],
+            keep_attrs=True,
```
CagtayFabry, line 367: I am not sure how or if default arguments changed with regards to keeping xarray attributes, however this is what we want

marscher: Eventually, we should make these arguments explicit to avoid breaking changes due to upstream default argument changes?

marscher (approved):

PR #875, opened 2023-05-05
add read_buffer_context and write_read_buffer_context

### LINKED ISSUE #873: `weldx.asdf.util.read_buffer` relies on reading from a 'closed' buffer
ASDF was allowing lazily loaded arrays to be read from an AsdfFile created by reading an `io.BytesIO` instance. This results in unexpected changes to the position of the read/write pointer within the stream.

See https://github.com/asdf-format/asdf/issues/1539 for the relevant ASDF issue.

It appears that weldx relies upon this behavior for any tests using `weldx.asdf.util.read_buffer`. The relevant code is as follows:
https://github.com/BAMWelDX/weldx/blob/c2ee9b1f7cb8df8b2f4723884341e9e20dcd0d50/weldx/asdf/util.py#L171-L189
Note that `data` is returned outside of the `asdf.open` context (which closes the underlying 'file' when the context exits). As `lazy_load=True` was not provided to `asdf.open` the arrays within the AsdfFile are 'lazy loaded' (their data is not read from the file until the array is sliced or in other ways accessed). One example of a test that will fail when the above issue is fixed in ASDF is `test_shape_validator`:
https://github.com/BAMWelDX/weldx/blob/c2ee9b1f7cb8df8b2f4723884341e9e20dcd0d50/weldx/tests/asdf_tests/test_asdf_validators.py#L150-L155
Here `result` contains references to 'lazy loaded' arrays that hold references to the now closed file. The call to `compare_nested` will then attempt to load the arrays which when the issue is fixed in ASDF will result in an error.

I see one usage of `read_buffer` outside of tests in `weldx.asdf.cli.welding_schema`:
https://github.com/BAMWelDX/weldx/blob/c2ee9b1f7cb8df8b2f4723884341e9e20dcd0d50/weldx/asdf/cli/welding_schema.py#L303-L306
However this does not appear to use the result of `write_read_buffer`.

I would imagine that one of two solutions might work:

1. used `lazy_load=True` with `asdf.open` to force loading of all arrays when the file is opened
2. convert `read_buffer` (and `write_read_buffer`) to context managers to allow the file to stay open while the results are used

The first option seems like the easiest but will result in all arrays being loaded into memory (which might not be necessary for every call to `read_buffer`).

braingram:
closes #873

## Changes

Add `read_buffer_context` and `write_read_buffer_context` to `weldx.asdf.util`.

Update previous uses of `read_buffer` and `write_read_buffer` that inspect the result to use the context.

`read_buffer` and `write_read_buffer` were updated to use the new context to avoid code duplication. These were not removed to preserve the API and to leave them available for uses where the result is not inspected. Such as this test:
https://github.com/BAMWelDX/weldx/blob/c2ee9b1f7cb8df8b2f4723884341e9e20dcd0d50/weldx/tests/asdf_tests/test_asdf_base_types.py#L15-L20

## Related Issues

Closes #873

## Checks

- [ ] updated `CHANGELOG.md`
- [x] updated `tests/`
- [ ] ~updated `doc/`~
- [ ] ~update example/tutorial notebooks~
- [ ] ~update manifest file~

### DIFF
--- a/weldx/tests/asdf_tests/test_asdf_util.py
+++ b/weldx/tests/asdf_tests/test_asdf_util.py
@@ -15,7 +15,7 @@ from weldx.asdf.util import (
     get_highest_tag_version,
     get_schema_tree,
     get_yaml_header,
-    read_buffer,
+    read_buffer_context,
     write_buffer,
 )
 
@@ -168,9 +168,10 @@ def test_write_buffer_dummy_inline_arrays():
     buff = write_buffer(tree={name: array}, write_kwargs=dict(dummy_arrays=True))
 
     buff.seek(0)
-    restored = read_buffer(buff)[name]
-    assert restored.dtype == array.dtype
-    assert restored.shape == array.shape
+    with read_buffer_context(buff) as data:
+        restored = data[name]
+        assert restored.dtype == array.dtype
+        assert restored.shape == array.shape
 
 
 def test_get_highest_tag_version():
--- a/weldx/asdf/util.py
+++ b/weldx/asdf/util.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Mapping, Set
+from contextlib import contextmanager
 from io import BytesIO, TextIOBase
 from pathlib import Path
 from typing import Any, Hashable, MutableMapping, Union
@@ -34,8 +35,10 @@ _INVOKE_SHOW_HEADER = False
 __all__ = [
     "get_schema_path",
     "read_buffer",
+    "read_buffer_context",
     "write_buffer",
     "write_read_buffer",
+    "write_read_buffer_context",
     "get_yaml_header",
     "view_tree",
     "notebook_fileprinter",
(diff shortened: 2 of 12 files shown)


weldx/tests/asdf_tests/test_asdf_util.py
```diff
@@ -163,14 +163,19 @@ class _DataClass:
...
-    restored = read_buffer(buff)[name]
-    assert restored.dtype == array.dtype
-    assert restored.shape == array.shape
+    with read_buffer_context(buff) as data:
+        restored = data[name]
+        assert type(restored) == TaggedDict
+        assert restored._tag.startswith("tag:stsci.edu:asdf/core/ndarray-")
+        assert restored["data"] == []
+        assert restored["datatype"] == array.dtype
+        assert restored["shape"][0] == array.shape[0]
```
CagtayFabry, line 178: looking at the old implementation I am surprised that this seemingly converted to `ndarray` previously @braingram 

The current behavior with a warning and reading as raw `TaggedDict` seems more fitting so I adjusted this test for now to reflect that expected output

CagtayFabry (approved):
thank you very much for you contribution here @braingram

weldx/tests/asdf_tests/test_asdf_util.py
braingram: Yeah, it's unclear what should be returned as asdf appears to have issues with empty inline arrays:
https://github.com/asdf-format/asdf/issues/1538
Do you have a link to the run that returned the `TaggedDict`? I don't think I saw that while running locally.

CagtayFabry: It looks like, if we run the tests without the `--weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header` flag, a  `TaggedDict` is returned (cannot convert to `ndarray`)
See an example here: https://github.com/BAMWelDX/weldx/actions/runs/4935769622/jobs/8822498546

However, if `--weldx-file-rw-buffer --weldx-file-rw-buffer-disp-header` is set, it looks like we do get an `ndarray`
example: https://github.com/BAMWelDX/weldx/actions/runs/4947440349/jobs/8846812586?pr=875

Do you have any idea here @marscher ?

marscher: oddly I cannot reproduce this locally. I will have to dig into it a bit.

marscher: I just removed this code branch. It just leads to confusion and is not really used.

PR #880, opened 2023-06-02
update converters for pint 0.22

### LINKED ISSUE #879: PintQuantityConverter error for newest pint
`pint.Quantity` now appears as `pint.registry.Quantity` in pint version 0.22 (as installed from pypi today). This new class path is missing from causing `PintQuantityConverter` to not convert instances.

braingram:
in pint 0.22 Quantity appears at pint.registry.Quantity and requires an update to the corresponding asdf
Converter (the same for pint.registry.Unit).

## Changes

I opted to add the type `pint.Quantity` to the types supported by `PintQuantityConverter`. Types and/or strings can be used here and both have pros/cons. Using the type requires that pint be imported when the `Converter` is created (which can slow down initialization while asdf loads extensions. However this doesn't appear to be an issue here as pint is already imported when `weldx.tags.units.pint_quantity` is imported. By using the type, pint is free to move the internal implementation of Quantity (which can change the class path) but as long as it appears importable via `pint.Quanity` the Converter should stay up-to-date.

In this case the string `"pint.Quantity"` also needs to be included as `weldx.constants.Q_` appears as a `"pint.Quantity"` when asdf inspects instances of `weldx.constants.Q_` (as inspected with [`asdf.util.get_class_name`](https://github.com/asdf-format/asdf/blob/b0c9a50e8d3add337e1271369a3bf834925176a4/asdf/util.py#L321)).

I added some comments with a shorter version of the above description. 

## Related Issues

Fixes #879 

## Checks

- [ ] updated `CHANGELOG.md`
- [ ] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixes
 
 - fix compatibility with `pint=0.21` \[{pull}`876`\].
+- fix typing compatibility with `pint=0.22` \[{pull}`880`\].
 - add `read_buffer_context` and `write_read_buffer_context` to `weldx.asdf.util`
   to fix tests accessing closed files. {issue}`873` \[{pull}`875`\].
 
@@ -16,6 +17,10 @@
 
 - removed keyword `dummy_inline_arrays` from function `asdf.util.write_buffer`
   as it was only used internally \[{pull}`875`\].
+
+### ASDF
+
+- update `PintQuantityConverter` and `PintUnitConverter` class assignments for `pint=0.22` \[{pull}`880`\].
 
 ## 0.6.6 (19.04.2023)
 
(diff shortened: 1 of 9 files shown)


CagtayFabry:
thank you for the thorough explanation and solution with listing `pint.Quantity` as string and class for the converter @braingram 

it is unfortunate that `pint` seems to be slightly unstable with the last releases in terms of typing and class refactoring.

I am waiting for the conda release of `0.22` [here](https://github.com/conda-forge/pint-feedstock/pull/57) to merge this and update our dependencies and CI accordingly

CagtayFabry (approved):

braingram:
Thanks!

PR #886, opened 2023-08-10
import ValidationError from asdf.exceptions

braingram:
## Changes

asdf has recently run into compatibility issues with jsonschema 4.18+. The newer versions of jsonschema dropped support for a feature required in asdf which left us with little choice but to vendorize jsonschema. This presented an opportunity to clean up some leaky bits of the asdf API including the issue that `jsonschema.ValidationError` was transparently passed to user code and libraries (like weldx) had to import `ValidationError` from jsonschema.

asdf 2.15 introduced `ValidationError` as part of the public api available at `asdf.exceptions.ValidationError`. The changes in this PR modify a few tests and docs to import `ValidationError` from `asdf.exceptions` (instead of jsonschema or from the top level asdf module).

asdf 2.15.1 includes the vendorized jsonschema, keeps jsonschema as a dependency and attempts to use the exceptions from the installed jsonschema (to not break some downstream libraries that use asdf). We are planning to drop jsonschema as a dependency and stop using it's exceptions in asdf 3.0 (or possibly in 2.15.2 if further changes to jsonschema make it's exceptions incompatible).

Please see the [What's New](https://asdf.readthedocs.io/en/latest/asdf/whats_new.html#jsonschema) page in the asdf 2.15.1 docs for more information and let me know if you have any questions, comments or concerns.

## Checks

- [ ] updated `CHANGELOG.md`
- [ ] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 ### ASDF
 
 - update `PintQuantityConverter` and `PintUnitConverter` class assignments for `pint=0.22` \[{pull}`880`\].
+- use `ValidationError` from `asdf` instead of `jsonschema` \[{pull}`886`\].
 
 ## 0.6.6 (19.04.2023)
 
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -279,7 +279,6 @@ def _get_intersphinx_mapping():
         # "dask": ("https://docs.dask.org/en/latest", None),
         # "numba": ("https://numba.pydata.org/numba-doc/latest", None),
         "pint": (f"https://pint.readthedocs.io/en/{pint_version}", None),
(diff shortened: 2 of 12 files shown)


CagtayFabry:
thanks a lot @braingram for carrying this over !

Does this require `asdf=2.15.1` for the new imports?

braingram:
Thanks!

Either 2.15.0 or 2.15.1 will work (2.15.0 has `ValidationError` at `asdf.exceptions` but a bug in the docs prevented it from being listed).

2.15.1 is probably safer as it includes the jsonschema vendorization which means 2.15.0 might break at some point if jsonschema decides to change their excepetion classes in a way that is incompatible with the vendorized version (4.17.3).

CagtayFabry (approved):

PR #896, opened 2023-10-03
add Python 3.12 to test

CagtayFabry:
## Changes

- add Python 3.12 and 3.13 to pytest runs
- switch Python 3.9 -> 3.10 in some default runs

## Checks

- [ ] updated `CHANGELOG.md`
- [ ] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file


marscher:
Conda environments cannot be properly created due to the missing mapping of different package names in pip and conda https://github.com/CagtayFabry/pydeps2env/issues/6

This is not specific to Python 3.12 but all Python versions.

marscher:
We first need to unblock weldx-widgets for Python 3.12. 
Is it blocked for the same reasons as weldx @CagtayFabry?

marscher:
The windows test is not passing because of some broken M$ caching. It cannot find the tag in my pydeps2env fork (Linux systems pick it up). So I guess https://github.com/CagtayFabry/pydeps2env/pull/57 is the last requirement for this PR @CagtayFabry

marscher:
Python 3.13 has not been released yet. We should allow failures for this version in our matrix. The current failure is due to Pandas is not yet ready for 3.13.

marscher:
Thanks! I will fix this is in wx-widgets then!

marscher:
fixed the np 2.0 compat, but pint is not py 3.13 ready.

CagtayFabry:
since there hasn't been an update regarding `pint` compatibility with 3.13 in that repository I propose to move this forward with 3.12 and add testing with 3.13 as it becomes available 😃

PR #904, opened 2024-01-03
fix WeldxFile doctest for asdf 3.0

braingram:
## Changes

As mentioned in https://github.com/BAMWelDX/weldx/pull/903

A doctest is failing due to changes in asdf 3.0 (including removal of `BuiltinExtension`) that result in slight and otherwise inconsequential header differences.

This PR expands the use of `...` in the doctest to allow the text matching to ignore these header differences.

## Checks

- [ ] updated `CHANGELOG.md`
- [ ] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file

### DIFF
--- a/weldx/asdf/file.py
+++ b/weldx/asdf/file.py
@@ -193,13 +193,10 @@ class WeldxFile(_ProtectedViewDict):
     %YAML 1.1
     %TAG ! tag:stsci.edu:asdf/
     --- !core/asdf-1.1.0
-    asdf_library: !core/software-1.0.0 {...
-      name: asdf, version: ...}
+    asdf_library: !core/software-1.0.0 ...
     history:
       extensions:
-      - !core/extension_metadata-1.0.0
-        extension_class: asdf.extension.BuiltinExtension
-        software: !core/software-1.0.0 {name: asdf, version: ...}
+      ...
     name: CXCOMP
     value: 42
     <BLANKLINE>


marscher:
LGTM as well. I do not get, why the notebook tests for Python 3.10 fail with an ImportError for WeldxFile.

braingram:
That one is kind of puzzling.

I do see an error related to "quality_standard_demo":
https://github.com/BAMWelDX/weldx/actions/runs/7398732873/job/20179218430?pr=904#step:12:82
```
File ~/work/weldx/weldx/quality_standard_demo/src/quality_standard_demo/__init__.py:1
----> 1 from ._version import __version__

ModuleNotFoundError: No module named 'quality_standard_demo._version'
```
It looks like the cookiecutter template here: https://github.com/BAMWelDX/quality-standard-template
is rendered and installed: https://github.com/BAMWelDX/weldx/blob/bb213a77ee79e8c982adfec0d56b9d5a3efee7b1/.github/workflows/pytest.yml#L78
but is lacking a `_version` file (perhaps something has changed in setuptools-scm that it's no longer generating a `_version.py`?).
The missing file appears to be preventing correct entry point loading for weldx which then leads to the ImportError for weldx.

CagtayFabry:
Thank you for the fix @braingram !

I haven't had time yet to look into the issue regarding the quality standard template I will try to follow up with what you already found

CagtayFabry (approved):

CagtayFabry:
I have updated the outdate build chain causing the issues https://github.com/BAMWelDX/quality-standard-template/pull/20

Thanks again @braingram

PR #918, opened 2024-03-06
use pandas.to_timedelta function instead of invoking deprecated ctor arg

marscher:
Pandas 2.2 deprecated the unit argument of the Timedeltaindex constructor and suggests to use pandas.totimedelta function instaed. This PR changes this as suggested.

## Checks

- [ ] updated `CHANGELOG.md`
- [ ] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file

### DIFF
--- a/weldx/tests/transformations/test_cs_manager.py
+++ b/weldx/tests/transformations/test_cs_manager.py
@@ -7,7 +7,6 @@ import numpy as np
 import pandas as pd
 import pytest
 import xarray as xr
-from pandas import TimedeltaIndex as TDI
 from pandas import Timestamp as TS
 
 import weldx.transformations as tf
@@ -953,7 +952,7 @@ def test_time_union_time_series_coords(
         )
 
     if exp_time is not None:
-        exp_time = TDI(exp_time, unit="s")
+        exp_time = pd.to_timedelta(exp_time, unit="s")
     assert np.all(exp_time == csm.time_union(list_of_edges))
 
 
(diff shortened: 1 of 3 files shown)


CagtayFabry:
Thanks for the fixes @marscher !

what is the minimum pandas version required for this?
Does it work with 1.5.x still?

marscher:
> what is the minimum pandas version required for this? Does it work with 1.5.x still?

I've added some backward compatibility code for reading old unit suffixes. Currently it only replaces the (used by weldx schema examples) suffix for seconds "S". I am not aware of any other units used in the past by weldx. To ensure we can still load old files with the new version, we should probably complete this list.

weldx/tags/time/timedeltaindex.py
```diff
@@ -9,6 +9,14 @@
 __all__ = ["TimedeltaIndexConverter"]
 
 
+def _handle_converted_pd_tdi_units(node: TaggedDict):
```
marscher, line 12: backward compat @CagtayFabry

marscher:
Sorry I misinterpreted your question. I did not check, if it works with old Pandas versions. Is this a requirement or can we just pin a minimum version?

weldx/tags/time/timedeltaindex.py
marscher: changelog of pandas:
> Units ‘H’, ‘T’, ‘S’, ‘L’, ‘U’ and ‘N’ are deprecated and will be removed in a future version. Please use ‘h’, ‘min’, ‘s’, ‘ms’, ‘us’, and ‘ns’ instead of ‘H’, ‘T’, ‘S’, ‘L’, ‘U’ and ‘N’.

marscher:
I've tested pandas-1.5 with Python-3.11 locally and all tests pass.

marscher:
The CI failures are unrelated to my changes. I think this one is ready to go.

CagtayFabry:
> I've tested pandas-1.5 with Python-3.11 locally and all tests pass.

If you tested it with 1.5 I am fine with pinning minimum pandas version >=1.5. 😊
I will update the CI to test for that later as well

weldx/tags/time/timedeltaindex.py
```diff
@@ -8,6 +8,15 @@
...
 __all__ = ["TimedeltaIndexConverter"]
 
+PANDAS_OLD_UNIT_SUFFIXES = dict(H="h", T="min", S="s", L="ms", U="us", N="ns")
+
+
+def _handle_converted_pd_tdi_units(node: TaggedDict):
+    """Convert changed units in Pandas.Datetimeindex to valid values."""
+    unit = node["freq"][-1]
+    if unit in PANDAS_OLD_UNIT_SUFFIXES.keys():
+        node["freq"] = node["freq"][:-1] + PANDAS_OLD_UNIT_SUFFIXES[unit]
```
CagtayFabry, line 18:
```suggestion
    for suf in PANDAS_OLD_UNIT_SUFFIXES:
        node["freq"] = node["freq"].replace(suf, PANDAS_OLD_UNIT_SUFFIXES[suf])
```
I think this should be more stable @marscher ?

marscher: I'm a bit confused about the stability here. Can you help me understand how a frequency string could have multiple time units? Is that even possible? Also, I assume it only covers cases where the frequency string is less than one character long, correct?

CagtayFabry: Multiple frequency "units" can be combined to form a single timedelta:
```python
pd.timedelta_range(start=0, periods=5, freq="2D10L")
```
Not sure if there is a way to end up in the file like that but since the old strings are one-character only I think the replace option should cover everything.

PR #949, opened 2024-11-03
948 add support to create coordinate systems from homogeneous transformation matrices

### LINKED ISSUE #948: Add support to create Coordinate Systems from homogeneous transformation matrices
Add a method to `CoordinateSystemManager` to add a new CoordinateSystem by using homogeneous transformation matrices as predominately used with `open3d.`

**Homogeneous Transformation from Coordinate System A to B**

```math
B = T_{AB} \times A = 
\left[ \begin{matrix}
r_{11} & r_{12} & r_{13} & t_{1}\\
r_{21} & r_{23} & r_{23} & t_{2}\\
r_{31} & r_{32} & r_{33} & t_{3}\\
0 & 0 & 0 & 0\\
\end{matrix}\right] \times 
\left[\begin{matrix}
A_{x} \\
A_{y}\\
A_{z} \\
1\\
\end{matrix}\right]
= \left[ \begin{matrix}
R & t \\
0 & 1\\
\end{matrix}\right] \times 
\left[ \begin{matrix}
A  \\
1\\
\end{matrix}\right] 
```

the transformation matrix is made up of the `Rotation R` and the `Translation t`.

Therefore, to adapt to the current code, the `orientation` and `coordinates` during `LocalCoordinateSystem` construction can be substituted using `homogenous_transformation` and `unit` (unit in which the translation part is given. e.g. "mm")

mbwinkler:
## Changes

Added support for homogeneous transformation matrices.
## Related Issues
Closes #948

## Checks

- [x] updated `CHANGELOG.md`
- [x] updated `tests/`
- [ ] updated `doc/`
- [ ] update manifest file

### DIFF
--- a/weldx/transformations/local_cs.py
+++ b/weldx/transformations/local_cs.py
@@ -757,14 +757,20 @@ class LocalCoordinateSystem(TimeDependent):
             time_dim = 1
 
         rotation = np.resize(self.orientation.data, (time_dim, 3, 3))
-        translation = np.resize(
-            self.coordinates.data.to(translation_unit).m, (time_dim, 3)
-        )
-        homogeneous_matrix = np.resize(np.identity(4), (time_dim, 4, 4))
-        homogeneous_matrix[:, :3, :3] = rotation
-        homogeneous_matrix[:, :3, 3] = translation
+        coordinates = self.coordinates
+        if not isinstance(coordinates, TimeSeries):
+            translation = np.resize(
+                coordinates.data.to(translation_unit).m, (time_dim, 3)
+            )
+            homogeneous_matrix = np.resize(np.identity(4), (time_dim, 4, 4))
+            homogeneous_matrix[:, :3, :3] = rotation
+            homogeneous_matrix[:, :3, 3] = translation
 
-        return np.squeeze(homogeneous_matrix)
+            return np.squeeze(homogeneous_matrix)
+        else:
+            raise NotImplementedError(
+                "Cannot convert LCS with `TimeSeries` coordinates to homogeneous matrix"
+            )
 
     def _interp_time_orientation(self, time: Time) -> xr.DataArray:
         """Interpolate the orientation in time."""
--- a/weldx/transformations/local_cs.py
+++ b/weldx/transformations/local_cs.py
@@ -728,6 +728,32 @@ class LocalCoordinateSystem(TimeDependent):
         """
         return Rot.from_matrix(self.orientation.values)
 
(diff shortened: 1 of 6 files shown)


mbwinkler:
Will update `tests/` and `doc/` and write an example for the notebook sometime this week.

CagtayFabry:
It looks good to me but is still missing tests obviously.
The test suit should be back up running after I merge https://github.com/BAMWelDX/weldx/pull/896 tomorrow

CagtayFabry:
the pytest actions are running again @mbwinkler , you can add some testcases 👍

CagtayFabry:
Don't bother with mypy for now @mbwinkler , it can be quite a hassle

mbwinkler:
@CagtayFabry 
Wouldn't be a hassle if `npt.ArrayLike` would **actually mean** "like an array" and not "can be constructed into an array"...

https://stackoverflow.com/a/76271691

I could change the type to just use `np.array` and then it would work. Right now mypy just tells me that the 3D indexing (obviously) would not work if I passed e.g. a simple `str`.

weldx/transformations/local_cs.py
```diff
@@ -690,6 +734,38 @@ def as_rotation(self) -> Rot:  # pragma: no cover
...
+
+        if self.is_time_dependent:
+            time_dim = self.time.shape[0]
+        else:
+            time_dim = 1
+
+        rotation = np.resize(self.orientation.data, (time_dim, 3, 3))
+        translation = np.resize(
+            self.coordinates.data.to(translation_unit).m, (time_dim, 3)
+        )
```
mbwinkler, line 762: Apparently this expression raises a mypy error, because using `self.coordinates` could return a `TimeSeries`, which in turn would return `pint.Quantity | MathematicalExpression` when `.data` is accessed and the latter one does not support unit transformations. Not sure how to fix this.

CagtayFabry: I think we can leave out supporting `TimeSeries` for now (throwing an error).

We could add
```py
if isinstance(self.coordinates, TimeSeries):
    raise NotImplementedError("Cannot convert LCS with `TimeSeries` coordinates to homogeneous matrix")
```
at the beginning

CagtayFabry (commented):
thanks for adding the test cases!

I think this can be set as "ready for review"? :)

CagtayFabry (approved):

CagtayFabry:
pytest windows failures unrelated

